PR #7, opened 2021-12-20
Add e2e tests as monorepo package

EmiM:


.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+      - uses: actions/setup-node@master
+        with:
+          node-version: "12"
+      - name: Install WM
+        run: sudo apt install fluxbox
+
+      - name: "Install lerna globally"
+        run: npm i -g lerna
+      
+      - name: "Install monorepo dependencies"
```
siepra, line 26: Add step caching NPM dependencies

siepra (approved):

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,73 @@
+name: End to end tests
+
+on:
```
siepra, line 3: I guess there should be two conditions: pull request on changes in e2e-tests package, and on release (plus holding release if check fails)

EmiM: Removed running tests on pr in ZbayLite but the second part (on release and holding release) will be done in other PRs

PR #20, opened 2021-12-23
Changed 'invite a friend' to 'add members'. Removed 'account' tab and…

siepra:
… corrected modal title

### DIFF
--- a/packages/ZbayLite/src/renderer/containers/widgets/settings/SettingsModal.tsx
+++ b/packages/ZbayLite/src/renderer/containers/widgets/settings/SettingsModal.tsx
@@ -1,4 +1,4 @@
-import { communities } from '@zbayapp/nectar'
+import { communities, identity } from '@zbayapp/nectar'
 import React, { useState } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import SettingsModal from '../../../components/widgets/settings/SettingsModal'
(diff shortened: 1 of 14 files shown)


packages/ZbayLite/src/renderer/containers/widgets/settings/SettingsModal.tsx
```diff
@@ -10,7 +10,7 @@ import { useModal } from '../../hooks'
 const useSettingsModalData = () => {
   const data = {
     modalTabToOpen: useSelector(appSelectors.currentModalTab),
-    user: 'settingsModalUser',
+    user: useSelector(identity.selectors.currentIdentity).zbayNickname,
```
EmiM, line 13: This throws error when you open the app for the first time (cause there is no identity yet)
`Cannot read property 'zbayNickname' of undefined`

siepra:
Huh, I think it's a bigger problem. All the modals exists before we
initialize community data so this error will return whenever we use some
'fragile' selectors inside them. The question is will that be common
scenario or is it just a single case?

pon., 27 gru 2021, 13:35 użytkownik breadhunter ***@***.***>
napisał:

packages/ZbayLite/src/renderer/containers/widgets/settings/SettingsModal.tsx
siepra: Huh, I think it's a bigger problem. All the modals exists before we
initialize community data so this error will return whenever we use some
'fragile' selectors inside them. The question is will that be common
scenario or is it just a single case?

siepra: Also, you shouldn't be able to reach that point without community at all

PR #132, opened 2022-01-07
Waggle - separate unit tests that use tor - they take a long time and…

EmiM:
… are less stable


packages/waggle/src/libp2p/connectionsManager.test.ts
```diff
@@ -16,25 +15,6 @@ beforeEach(() => {
 })
 
 describe('Connections manager', () => {
```
siepra, line 17: Maybe add some info that those tests runs without tor? (not using tor)

packages/waggle/src/registration/testUtils.ts
```diff
@@ -0,0 +1,62 @@
...
+import fetch, { Response } from 'node-fetch'
+import PeerId from 'peer-id'
+import { CertificateRegistration } from '.'
+import { createLibp2p } from '../common/testUtils'
+import { DataFromPems } from '../common/types'
+import { DummyIOServer } from '../common/utils'
+import { Storage } from '../storage'
+import { Tor } from '../torManager'
+
+export async function registerUserTest(csr: string, httpTunnelPort: number, localhost: boolean = true, registrarPort: number = 7789): Promise<Response> {
```
siepra, line 11: registerUserTest sounds like it's a test scenario

siepra (approved):
I think it's ok. A little bit more complicated than it was before but we all understand why it has to be done that way

packages/waggle/src/libp2p/connectionsManager.test.ts
EmiM: I think that's self explanatory if we have `.tor.test.ts` files with tor tests. We would have to add "(not using tor)" to 95% of our test cases.
I added (using tor) to tor-test names instead.

PR #139, opened 2022-01-11
restored connections test

vinkabuki:


packages/waggle/src/storage/storageBasic.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const log = logger('dbSnap')
+
+class StorageTestSnapshotOptions extends StorageOptions {
+  messagesCount: number
+  createSnapshot?: boolean = false
+  useSnapshot?: boolean = false
+}
+
+export class StorageBasic extends Storage {
```
EmiM, line 17: StorageBasic sounds like it would be the base class but here it extends Storage. Is it a base just for tests?

packages/waggle/src/nodeTest/testConnect.test.ts
```diff
@@ -197,9 +199,9 @@ describe('Nodes connections', () => {
     await sleep(40_000)
   })
 
-  test('5 nodes connect each other - no tor, providing only one bootstrap multiaddress', async () => {
-    const timeout = 360_000
-    const noOfNodes = 2
+  test.skip('5 nodes connect each other - no tor, providing only one bootstrap multiaddress', async () => {
```
EmiM, line 202: Why this is skipped?

packages/waggle/src/storage/storageBasic.ts
vinkabuki: Yes it is a base for test, we should create maybe a storageTestUtils and put it inside along with storageSnapshot, because we use them solely for test purposes

vinkabuki: I didn't do that keep MR clear

packages/waggle/src/nodeTest/testConnect.test.ts
vinkabuki: Because it's not a real-app scenario, now we rely on custom orbitdb strucutre for getting info about peers. I think it's what we just decided on meeting that those tests are not that important

EmiM (approved):

siepra (approved):

PR #148, opened 2022-01-17
Make e2e tests workflow reusable

EmiM:


.github/workflows/build-linux.yml
```diff
@@ -3,10 +3,14 @@ name: Build-Linux
 on:
   release:
     types:
-      [released , prereleased]
+      [released, prereleased]
 
 jobs:
+  run-e2e-tests:
+    uses: ZbayApp/monorepo/.github/workflows/e2e-tests-reusable.yml@ff1fbf8d1853f23ba97c2c00116ef2312ae8977f
```
kowalski, line 10: this `ff1fbf8d1853f23ba97c2c00116ef2312ae8977f` is a hash of commit right? I think it's better if you don't freeze it to a particular commit, becuase than it will just take a state of current branch

however I've never used this feature, so maybe I misunderstand how this works

kowalski (approved):

.github/workflows/build-linux.yml
EmiM: I've responded to Marek in dm already but I'm writing here for history purposes. 
Currently reusable workflows need a static reference (tag, hash or branch). They are working on making this more pleasant to use: https://github.community/t/ref-head-in-reusable-workflows/203690/30

siepra (approved):

PR #150, opened 2022-01-17
Storage refactoring (io)

EmiM:
* use proxy io instead of socket io
* move all socket methods to proxyIO


siepra (approved):

siepra (approved):

packages/waggle/src/storage/storage.test.ts
```diff
@@ -2,24 +2,27 @@ import fs from 'fs'
...
 let tmpIpfsPath: string
+let connectionsManager: ConnectionsManager
 let storage: Storage
 beforeEach(() => {
   jest.clearAllMocks()
   tmpDir = createTmpDir()
   tmpAppDataPath = tmpZbayDirPath(tmpDir.name)
   tmpOrbitDbDir = path.join(tmpAppDataPath, Config.ORBIT_DB_DIR)
   tmpIpfsPath = path.join(tmpAppDataPath, Config.IPFS_REPO_PATH)
+  connectionsManager = createMinConnectionManager({ env: { appDataPath: tmpAppDataPath }, torControlPort: 12345 })
```
vinkabuki, line 25: Is this hardcoded torControlPort safe? Shouldn't we find port dynamically?

vinkabuki (approved):

packages/waggle/src/socket/IOProxy.ts
```diff
@@ -117,6 +117,56 @@ export default class IOProxy {
     await this.getStorage(peerId).subscribeToAllConversations(conversations)
   }
 
+  public loadCertificates = (payload: SendCertificatesResponse) => {
+    log(`Sending back ${payload.certificates.length} certificates`)
```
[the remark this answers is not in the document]
EmiM: I added log before each `io.emit` in methods that we currently use

packages/waggle/src/storage/storage.test.ts
EmiM: We should. Done.

PR #151, opened 2022-01-18
Feature/display community name

vinkabuki:


packages/ZbayLite/src/renderer/components/ui/IdentityPanel/IdentityPanel.test.tsx
```diff
@@ -3,43 +3,31 @@ import React from 'react'
...
 
 describe('IdentityPanel', () => {
   it('renders component with username', () => {
-    const identity: Identity = {
-      id: '',
-      zbayNickname: '',
-      hiddenService: {
-        onionAddress: 'string',
-        privateKey: 'string'
+    const community: Community = {
```
siepra, line 13: I'd really appreciate if you used factory-girl :D

packages/ZbayLite/src/renderer/components/ui/IdentityPanel/IdentityPanel.test.tsx
```diff
@@ -3,43 +3,31 @@ import React from 'react'
...
-        pkcs10: {
-          publicKey: 'any',
-          privateKey: 'any',
-          pkcs10: 'any'
-        }
-      },
-      userCertificate: ''
+      port: null
     }
     const result = renderComponent(
```
siepra, line 28: There's no need for nesting `<Provider>` as it's already a part of `renderComponent`

packages/ZbayLite/src/renderer/components/ui/IdentityPanel/IdentityPanel.tsx
```diff
@@ -54,7 +54,7 @@ export const IdentityPanel: React.FC<IdentityPanelProps> = ({ handleSettings, id
         component='span'
         classes={{ root: classes.button, label: classes.buttonLabel }}>
         <Typography variant='h4' className={classes.nickname}>
-          {identity?.zbayNickname ? identity.zbayNickname : ''}
+          {community?.name ? community.name : ''}
```
siepra, line 57: `{ community?.name || '' }` ?

packages/ZbayLite/src/rtl-tests/community.join.test.tsx
```diff
@@ -79,7 +79,8 @@ describe('User', () => {
             id: payload.id,
             payload: {
               certificate:
-                'MIIBTDCB8wIBATAKBggqhkjOPQQDAjASMRAwDgYDVQQDEwdaYmF5IENBMB4XDTEwMTIyODEwMTAxMFoXDTMwMTIyODEwMTAxMFowEjEQMA4GA1UEAxMHWmJheSBDQTBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABEaV1l/7BOvPh0fFteSubIJ2r66YM4XoMMEfUhHiJE6O0ojfHdNrsItg+pHmpIQyEe+3YGWxIhgjL65+liE8ypqjPzA9MA8GA1UdEwQIMAYBAf8CAQMwCwYDVR0PBAQDAgCGMB0GA1UdJQQWMBQGCCsGAQUFBwMCBggrBgEFBQcDATAKBggqhkjOPQQDAgNIADBFAiARHtkv7GlhfkFbtRGU1r19UJFkhA7Vu+EubBnJPjD9/QIhALje1S3bp8w8jjVf70jGc2/uRmDCo/bNyQRpApBaD5vY'
+                'MIIBTDCB8wIBATAKBggqhkjOPQQDAjASMRAwDgYDVQQDEwdaYmF5IENBMB4XDTEwMTIyODEwMTAxMFoXDTMwMTIyODEwMTAxMFowEjEQMA4GA1UEAxMHWmJheSBDQTBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABEaV1l/7BOvPh0fFteSubIJ2r66YM4XoMMEfUhHiJE6O0ojfHdNrsItg+pHmpIQyEe+3YGWxIhgjL65+liE8ypqjPzA9MA8GA1UdEwQIMAYBAf8CAQMwCwYDVR0PBAQDAgCGMB0GA1UdJQQWMBQGCCsGAQUFBwMCBggrBgEFBQcDATAKBggqhkjOPQQDAgNIADBFAiARHtkv7GlhfkFbtRGU1r19UJFkhA7Vu+EubBnJPjD9/QIhALje1S3bp8w8jjVf70jGc2/uRmDCo/bNyQRpApBaD5vY',
```
siepra, line 82: I don't really like certificates being hardcoded. I think it's better to generate them using factory-girl

packages/ZbayLite/src/renderer/components/ui/IdentityPanel/IdentityPanel.tsx
vinkabuki: modals :)

packages/ZbayLite/src/rtl-tests/community.join.test.tsx
```diff
@@ -51,6 +52,14 @@ describe('User', () => {
 
     const factory = await getFactory(store)
 
+    const quietCommunity = await factory.create<
```
siepra, line 55: The very important thing about using factory-girl is that it saves value to the store. There are two methods: `build` and `create`. The first one builds the model but it **doesn't save** it to the store - and that's the desired behavior in this case (we test real-life-like scenario in which we want to modify the store only with the data that comes from the mocked waggle). At this moment, it turns out that we have two communities in this test

packages/ZbayLite/src/rtl-tests/community.join.test.tsx
```diff
@@ -51,6 +52,14 @@ describe('User', () => {
...
+      ReturnType<typeof communities.actions.addNewCommunity>['payload']
+    >('Community')
+
+    const quietUser = await factory.create<
```
siepra, line 59: This is the same thing as above but also there's no need for creating additional user as we already have one being created at line 69 (you should only add test-scope accessible reference)

EmiM (approved):

PR #154, opened 2022-01-19
Fix/verifying message signature

siepra:


packages/identity/src/test/index.test.ts
```diff
@@ -22,23 +21,49 @@ describe('Message signature verification', () => {
...
+      signature: await sign(message, userCsr.pkcs10.privateKey)
+    }
+
+    const result = await verifySignature(
+      data.signature,
+      data.message,
+      data.userPubKey
+    )
+
+    expect(result).not.toBe(true)
```
kowalski, line 65: `.not.toBe(true)` means that if `result === 'Różowy słoń'` it will still pass the test

I suggest 
```
toBe(false)
```

packages/identity/src/verification.ts
```diff
@@ -1,14 +1,28 @@
...
 
 export const verifySignature = async (
-  userPubKey: CryptoKey,
   signature: ArrayBuffer,
-  message: string
+  message: string,
+  publicKeyObject?: CryptoKey,
+  publicKeyString?: string
 ): Promise<boolean> => {
+  if (!publicKeyObject && !publicKeyString) {
```
kowalski, line 14: this is a bit ugly...
optional positional arguments are weird, because if you wan to pass 5th without passing 4th you need to pass `undefined` explicitely. 

I suggest sth like:
```
export async verifySignature (
  userPubKey: CryptoKey,
  signature: ArrayBuffer,
  message: string
  message: string,
  publicKey: CryptoKey | string,
): Promise<boolean> {

...

if (typeOf publicKey === 'string') {
  publicKey = await keyObjectFromString(publicKey, crypto)
}
```

kowalski: also, bare in mind that reconstructing CryptoKey is expensive, so if it's done to check *every single message* it's going to cost us. It may be good to not support passing `publicKey` as string, because than user needs to hold reference to CryptoKey

packages/waggle/src/storage/MessagesAccessController.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const type = 'messagesaccess'
+
+export class MessagesAccessController extends AccessController {
+  static get type () { return type }
+
+  async canAppend(entry) {
+    const message: ChannelMessage = entry.payload.value
+    const signature = stringToArrayBuffer(message.signature)
+    const verify = await verifySignature(signature, message.message, null, message.pubKey)
```
kowalski, line 14:
```
  const verify = await verifySignature(signature, message.message, null, message.pubKey)
```
here I'd suggest having a mapping of ` { pubKey: CryptoKey }` on this (as a cache). This way we are caching instances of CryptoKey and not "paying" for constructing it for every single message

packages/identity/src/test/index.test.ts
siepra: I keep forgetting, in typescript boolean variable can be boolean. Or string. Or object...

packages/identity/src/verification.ts
siepra: So that's what I was trying to implement in the first place. But later on you have crypto.verify method which accepts only CryptoKey

packages/waggle/src/storage/MessagesAccessController.ts
siepra: I don't fully get it. We have no power over passing data to this method (I guess). It fires automatically just before saving data to db, e.g. after replication so where can we do any mapping?

packages/waggle/src/storage/storage.test.ts
```diff
@@ -42,103 +94,186 @@ describe('Storage', () => {
...
+    await storage.subscribeToChannel(channel)
+
+    const spy = jest.spyOn(storage.publicChannelsRepos.get(message.channelId).db, 'add')
+
+    await storage.sendMessage(message)
+
+    // Confirm message has passed orbitdb validator (check signature verification only)
+    expect(spy).toHaveBeenCalled()
+
+    // Confirm message hasn't been added to db
```
EmiM, line 235: I think you meant "Confirm message has been added to db"

kowalski (approved):

vinkabuki (approved):

PR #159, opened 2022-01-21
Added store structure for verified messages

siepra:


vinkabuki (approved):

packages/waggle/src/storage/storage.ts
```diff
@@ -429,24 +427,24 @@ export class Storage {
...
-      this.io.loadAllMessages({
-        messages: this.getAllEventLogEntries(db),
-        channelAddress,
-        communityId: this.communityId
-      })
+      // this.io.loadAllMessages({
+      //   messages: this.getAllEventLogEntries(db),
+      //   channelAddress,
+      //   communityId: this.communityId
+      // })
```
holmesworcester, line 434: What's this?

holmesworcester (approved):

packages/waggle/src/storage/storage.ts
siepra: Oh, this function is a valid logic for DM (as I've been told by Bartek). I've commented out the code that needs to be changed as soon as we start working of bringing back direct messages feature, because it throws typescript errors

PR #160, opened 2022-01-24
change Zbay to Quiet in building process and internals

vinkabuki:


siepra:
Don't forget a part of the task is also changing the sticky-note on the pin board in the office in Cracow

holmesworcester (approved):

PR #164, opened 2022-01-25
Restore test with factory-girl

siepra:


packages/nectar/src/sagas/publicChannels/checkForMessages/checkForMessages.saga.test.ts
```diff
@@ -1,122 +1,90 @@
...
+    alice = await factory.create<ReturnType<typeof identityActions.addNewIdentity>['payload']>(
+      'Identity',
+      { id: community.id, zbayNickname: 'alice' }
+    )
+  })
+
+  test('ask for missing messages', async () => {
+    const types = [MessageType.Basic, MessageType.Empty, MessageType.Empty]
+
+    const shuffled = types
```
holmesworcester, line 48: Curious: why are we shuffling?

packages/nectar/src/sagas/publicChannels/checkForMessages/checkForMessages.saga.test.ts
```diff
@@ -1,122 +1,90 @@
...
-            )
-          }
-        }
-      )
+          verifyAutomatically: true
+        })
+      })
+    )
+
+    const empties = messages.filter(item => item.message.type === MessageType.Empty)
```
holmesworcester, line 73: Where are empty messages from?

holmesworcester: We're just not displaying empty messages?

holmesworcester (approved):

packages/nectar/src/sagas/publicChannels/checkForMessages/checkForMessages.saga.test.ts
siepra: It probably not necessary but I wanted to make sure the mechanism will never depend on the messages order

packages/nectar/src/sagas/publicChannels/checkForMessages/checkForMessages.saga.test.ts
siepra: We add message object with no data but the id, as we receive the list of ids from waggle. Then we ask for the messages and we update those 'empty' objects with data

EmiM (approved):

PR #167, opened 2022-01-26
Emit event from AccessController

siepra:


siepra:
I close this pull request as we're going to implement a different solution based on OrbitDB changes

PR #178, opened 2022-01-27
Moved databases intialization after storage starts

siepra:


packages/waggle/src/storage/storage.ts
```diff
@@ -181,7 +183,6 @@ export class Storage {
       await this.channels.load({ fetchEntryTimeout: 2000 })
       this.io.loadPublicChannels({
         communityId: this.communityId,
-        // @ts-expect-error
         channels: this.channels.all
```
vinkabuki, line 186: We have a problem with this - lerna bootstrap says it's unused, but CI will fail without this

vinkabuki (approved):

packages/waggle/src/storage/storage.ts
siepra: Damn, that's bad :/

holmesworcester (approved):

PR #184, opened 2022-01-28
Fix/restarting websocket connection after sleep

siepra:


vinkabuki (approved):

packages/frontend/src/main/main.ts
```diff
@@ -297,12 +297,20 @@ app.on('ready', async () => {
   })
 
   ipcMain.on('start-waggle', async () => {
-    await waggleProcess?.connectionsManager.closeAllServices()
-    await waggleProcess?.dataServer.close()
-    waggleProcess = await runWaggle(mainWindow.webContents)
+    await startWaggle()
+  })
+
+  powerMonitor.addListener('resume', async () => {
```
EmiM, line 303: Does websocket connection get lost with every sleep? If not then I'm wondering if it won't be annoying for user to wait for waggle each time computer is being resumed. Isn't there a way to check if connection is lost and only then restart waggle?

siepra: It happens very fast, the user won't even notice. Also there's no reliable way to verify if websocket connection has been lost at this point

holmesworcester (approved):

siepra:
Replaced by https://github.com/ZbayApp/monorepo/pull/191

PR #197, opened 2022-02-02
Fix/display channel name taken error

siepra:


packages/frontend/src/renderer/components/widgets/channels/CreateChannel/CreateChannel.tsx
```diff
@@ -65,7 +65,11 @@ const useStyles = makeStyles(theme => ({
 }))
 
 export const parseChannelName = (name = '') => {
-  return name.toLowerCase().replace(/ +/g, '-')
+  return name
+    .toLowerCase()
+    .trimStart()
```
EmiM, line 70: `trim` does the same as trimStart + trimEnd

packages/frontend/src/renderer/containers/widgets/channels/CreateChannel.tsx
```diff
@@ -15,10 +25,16 @@ export const CreateChannel = () => {
   const community = useSelector(communities.selectors.currentCommunityId)
   const channels = useSelector(publicChannels.selectors.publicChannels)
 
+  const communityErrors = useSelector(errors.selectors.currentCommunityErrors)
+  const error = communityErrors?.find(error => error.type === SocketActionTypes.CREATED_CHANNEL)
```
EmiM, line 29: Can't we have a selector for this (like "get by type")? This way one doesn't have to filter manually by error.type every time

packages/frontend/src/renderer/containers/widgets/channels/CreateChannel.tsx
```diff
@@ -31,6 +47,25 @@ export const CreateChannel = () => {
...
 
   const createChannel = (name: string) => {
+    // Clear errors
+    if (error) {
+      dispatch(
+        errors.actions.clearError(error)
+      )
+    }
+    // Validate channel name
+    if (channels.some(channel => channel.name === name)) {
```
EmiM, line 57: I'm thinking if maybe we should validate it on waggle side as we are doing it with checking duplication of username.

vinkabuki (approved):

packages/frontend/src/renderer/containers/widgets/channels/CreateChannel.tsx
vinkabuki: What are arguments behind this? Checking username duplication is different - it's owner who checks for it, and he is the one who always knows about every taken username, and registration module is in waggle. This case is different and checking it in waggle doesn't make things safer or better imo

packages/frontend/src/renderer/components/widgets/channels/CreateChannel/CreateChannel.tsx
siepra: I've been trying to use it and it turned out it trims also the spaces between the words

packages/frontend/src/renderer/containers/widgets/channels/CreateChannel.tsx
siepra: So what I was thinking when doing it this way was that the main purpose of the task is to present proper validation error message (and improve UX at this stage). We've been through a lot of discussions about uniqueness of channel's names/addresses and I think we haven't come with any ideal solution for now. 
What I did is relatively simple and it provides good user experience as it displays understandable error message when trying to add a channel with the same name as one of those which are already in our store, or subscribes to the channel we haven't yet subscribed to

packages/frontend/src/renderer/components/widgets/channels/CreateChannel/CreateChannel.tsx
EmiM: It doesn't. I use it in a different place and also check the docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim

PR #201, opened 2022-02-03
Feature/add notifications and settings into store

Dinek007:


packages/frontend/src/renderer/notifications.ts
```diff
@@ -1,46 +1,63 @@
...
+
+  const createNotificationsCalls = action.payload.messages.map((messageData) => {
+    const publicChannelFromMessage = publicChannels.find((channel) => channel.address === messageData.channelId)
+
+    const isMessageFromMyUser = usersData[messageData.pubKey]?.username === myIdentity.nickname
+    const isMessageFromCurrentChannel = currentChannel === publicChannelFromMessage.name
+    const isNotificationsOptionOff = NotificationsOptions.doNotNotifyOfAnyMessages === notificationsOption
+
+    if (!isMessageFromMyUser && !isMessageFromCurrentChannel && !isNotificationsOptionOff) {
+      return call(createNotificationSaga, {
```
siepra, line 34: Shouldn't it be `yield*` instead of `return`?

packages/frontend/src/renderer/notifications.ts
```diff
@@ -1,46 +1,63 @@
...
+  action: PayloadAction<ReturnType<typeof channels.actions.incomingMessages>['payload']>
+): Generator {
+  const publicChannels = yield* select(channels.selectors.publicChannels)
+  const usersData = yield* select(users.selectors.certificatesMapping)
+  const myIdentity = yield* select(identity.selectors.currentIdentity)
+  const currentChannel = yield* select(channels.selectors.currentChannel)
+  const notificationsOption = yield* select(settings.selectors.getNotificationsOption)
+  const notificationsSound = yield* select(settings.selectors.getNotificationsSound)
+
+  const createNotificationsCalls = action.payload.messages.map((messageData) => {
```
siepra, line 26: Saga must not contain any logic itself. It just delegates the tasks so you should extract it to a separate function and use `yield* call` on it

packages/frontend/src/renderer/notifications.ts
```diff
@@ -1,46 +1,63 @@
...
+        sound: notificationsSound,
+        communityId: action.payload.communityId,
+        channelName: publicChannelFromMessage.name
+      })
+    }
+  })
+  yield* all(createNotificationsCalls)
+}
+
+export function* createNotificationSaga(payload: NotificationsData): Generator {
```
siepra, line 46: I don't think it has to be saga. It looks like it should be just a regular function (as it contains logic). Also sagas does not return objects - they returns `Generator`

siepra:
I see this saga you created doesn't have unit test. The best way to verify if your saga is designed properly is to write unit test for it. If it contains any logic you'll face problems trying to mock it - and that's the sign something's wrong - because sagas only delegates tasks (yield)

packages/frontend/src/renderer/notifications.ts
Dinek007: no, it's return in map method, `yield` is below with all( array of `call`s )

holmesworcester (commented):
Left some comments.

Also, is there a test for "notification shows for message in current channel when app window does not have focus? I think that's important too.

Basically notifications should only not display when the window has focus and is in the current channel.

Also, does the scroll position move to bottom when you click on a notification? We should have a test for that too. 

Also, I think a notification should possibly still display even when the app has focus but the current channel is scrolled up. What do you think about this?

packages/frontend/src/renderer/notifications.ts
```diff
@@ -1,46 +1,103 @@
...
+    myIdentity: yield* select(identity.selectors.currentIdentity),
+    currentChannel: yield* select(channels.selectors.currentChannel),
+    notificationsOption: yield* select(settings.selectors.getNotificationsOption),
+    notificationsSound: yield* select(settings.selectors.getNotificationsSound)
   }
-  return await createNotification({
-    title: `New message from ${username || 'Unnamed'}`,
-    body: `${message.message.substring(0, 64)}${message.message.length > 64 ? '...' : ''}`,
-    data: `/main/direct-messages/${username}`
+  const createNotificationsCalls = yield* call(messagesMapForNotificationsCalls, createNotificationsCallsData)
```
kowalski, line 46: const createNotificationsCalls = yield* messagesMapForNotificationsCalls(createNotificationsCallsData)

kowalski, line 46: const notificationClickedChannel = yield* call(messagesMapForNotificationsCalls, createNotificationsCallsData)
yield takeEvery(notificationClickedChannel, function* (action) {
    yield put(action)
  })

packages/frontend/src/renderer/components/widgets/channelSettings/Notifications.tsx
```diff
@@ -109,91 +99,33 @@ export const Notifications: React.FC<NotificationsProps> = ({
...
+                  onClick={() => { }}
+                >
+                  Unmute Channel
+                </Button>
+              </Grid>
+            </Grid>
+
             <Grid item className={classes.captionDiv}>
               <Typography variant='caption' className={classes.captions}>
                 You can choose how to be alerted or turn off all Quiet
```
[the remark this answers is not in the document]
Dinek007: there are unused files, we have 'settings' folder in widgets now instead of that 'channelSettings', Bartek removed this files in his bump electron pr, i just changed it in components and containers when i removed old store structure and replaced with new 'settings' store structure in nectar, but finally its unused and will be removed

packages/frontend/src/renderer/store/handlers/notificationCenter.test.ts
```diff
@@ -1,68 +0,0 @@
-/* eslint import/first: 0 */
```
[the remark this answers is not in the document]
Dinek007: that was the old logic for notifications and notification settings

holmesworcester:
I don't see a test for 'sounds should not play for notification when sounds are turned off in settings'. Can you add?

holmesworcester (approved):

packages/frontend/src/renderer/testUtils/prepareStore.ts
```diff
@@ -28,18 +29,62 @@ export const reducers = {
...
+  }
+
+  effectResolved: SagaMonitorType['effectResolved'] = (effectId, result) => {
+    const triggeredEffect = this.effectsTriggeredArray.get(effectId)
+    this.effectsResolvedArray.set(effectId, { ...triggeredEffect, result })
+  }
+
+  public isEffectResolved = () => {
+    const parentEffect = Array.from(this.effectsResolvedArray).filter((effect) => {
+      return effect[1].result.meta?.name === 'takeEvery(channel, bridgeAction)'
```
kowalski, line 66: Move string to param

PR #221, opened 2022-02-10
Chore/bump electron

vinkabuki:


packages/frontend/src/renderer/containers/widgets/settings/Notifications.tsx
```diff
@@ -22,11 +19,11 @@ export const useNotificationsActions = (userFilterType: number, sound: number) =
   const dispatch = useDispatch()
 
   const setUserNotification = useCallback(() => {
-    dispatch(notificationCenterHandlers.epics.setUserNotification(userFilterType))
+    // dispatch(notificationCenterHandlers.epics.setUserNotification(userFilterType))
```
holmesworcester, line 22: What's this?

packages/frontend/src/renderer/store/reducers.ts
```diff
@@ -13,16 +13,16 @@ import { socketReducer } from '../sagas/socket/socket.slice'
...
-import contactsHandlers from './handlers/contacts'
-import channelHandlers from './handlers/channel'
-import directMessages from './handlers/directMessages'
-import directMessageChannelHandlers from './handlers/directMessageChannel'
-import notifications from './handlers/notifications'
-import notificationCenter from './handlers/notificationCenter'
-import mentionsHandlers from './handlers/mentions'
-import criticalErrorHandlers from './handlers/criticalError'
-import whitelistHandlers from './handlers/whitelist'
+// import waggleHandlers from './handlers/waggle'
```
holmesworcester, line 16: Why commented and not removed?

holmesworcester (approved):

holmesworcester (approved):
Approving but we should discuss context isolation and make an issue for this.

It doesn't seem to be urgent because we don't load from the web, but it is recommended so we should probably do it.

PR #249, opened 2022-02-18
send user certificate if csr pub key match cert pub key

vinkabuki:


packages/waggle/src/storage/storage.test.ts
```diff
@@ -203,8 +203,8 @@ describe('Certificate', () => {
     await storage.saveCertificate({ certificate: userCertificate.userCertString, rootPermsData })
 
     for (const username of ['alice', 'Alice', 'Ąlice']) {
-      const usernameExists = storage.usernameExists(username)
-      expect(usernameExists).toBe(true)
+      const usernameCert = storage.usernameCert(username)
+      expect(usernameCert).toBeTruthy()
```
EmiM, line 207: I think It would be better to check if returned cert is the proper cert

EmiM (approved):

PR #257, opened 2022-02-22
fixed scroll issues

vinkabuki:


holmesworcester (approved):

packages/frontend/src/renderer/components/widgets/channels/ChannelMessages.tsx
```diff
@@ -93,10 +97,28 @@ export const ChannelMessagesComponent: React.FC<IChannelMessagesProps> = ({
...
     }
   }, [messages.count])
 
+  /* Scroll to bottom if new message arrives and scroll is at the top */
+  useEffect(() => {
+    if (
+      scrollbarRef.current &&
+      scrollPosition === 0 &&
+      previousMessages &&
+      // @ts-ignore
```
EmiM, line 112: Why we need ts-ignore here?

vinkabuki: previousMessages and previousSlice have undefined type because on first render they are indeed undefined

.github/workflows/regression-test.yml
```diff
@@ -0,0 +1,30 @@
+name: Regression tests
```
holmesworcester, line 1: Should we name this in some more precise way?

packages/frontend/src/renderer/components/Channel/Channel.regression.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    cy.get('[data-testid="channelContent"]').scrollTo(0, 0)
+
+    cy.get('[data-testid="channelContent"]').compareSnapshot('scroll to the top')
+
+    // Send only one message because previous bug was only after sending one message
+    cy.get('[data-testid="messageInput"]').focus().type('and youda too').type('{enter}')
+
+    cy.get('[data-testid="channelContent"]').compareSnapshot('send after top scroll')
+  })
```
holmesworcester, line 74: I think we're missing a test for 'scroll should be at the bottom when user was previously scrolled to bottom and receives message'

holmesworcester (changes requested):
See comment

holmesworcester (approved):

packages/frontend/src/renderer/components/Channel/Channel.regression.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+    // Wait for component to fully render
+    cy.wait(3000)
+  })
+
+  it('scroll should be at the bottom after entering channel', () => {
+    cy.get('[data-testid="channelContent"]').compareSnapshot('after launch')
+  })
+
+  it('scroll should be at the bottom after sending messages', () => {
+    cy.get('[data-testid="messageInput"]').focus().type('luke where are you?').type('{enter}')
```
EmiM, line 41: Could you put all of `'[data-testid="<value>"]'` in variables?

PR #263, opened 2022-02-23
Run initial functionality on mobile

siepra:


packages/mobile/src/components/Chat/Chat.stories.tsx
```diff
@@ -15,43 +14,169 @@ storiesOf('Chat', module).add('Default', () => (
...
-      },
-      {
-        id: '4',
-        type: MessageType.Basic,
-        message:
-          'Chupa chups soufflé danish. Cake chocolate brownie cookie bear claw soufflé. Muffin chupa chups sweet caramels jelly beans chocolate bar bonbon liquorice.',
-        createdAt: 0,
-        date: '1:32pm',
-        nickname: 'bartek'
+    messages={{
```
holmesworcester, line 17: is there any value in sharing this mock conversation, with all our examples of problematic or edge case messages, between mobile and desktop? it seems good to centralize our trickiest examples

packages/mobile/src/components/Chat/Chat.component.tsx
```diff
@@ -126,3 +132,21 @@ const customInputStyle = StyleSheet.create({
...
   }
 })
+
+export const ChannelMessagesComponent: React.FC<ChannelMessagesComponentProps> = ({ messages }) => {
+  return (
+    <>
+      {Object.keys(messages).map(day => {
+        return (
+          <View key={day}>
+            {/* <MessagesDivider title={day} /> */}
```
vinkabuki, line 142: Do we need that comment here?

packages/mobile/src/components/Chat/Chat.stories.tsx
siepra: Yeah, we could... I'll open separate ticket for that

vinkabuki (approved):

packages/mobile/src/components/Chat/Chat.component.tsx
siepra: I think so. Just to remind us that we should place MessagesDivider component (as soon as we add it) in this exact place

siepra: I needed to adapt this component so we're able to display messages without having to change the nectars' selectors

EmiM (approved):

Issue #180, opened 2022-01-27
Incorrect error message when user creating channel with duplicate name


holmesworcester:
steps to reproduce:

1. create channel named "test"
2. create another channel named "test"

expected: app lets you create two similarly named channels OR app blocks you with an error message such as "Channel name taken"

I think we have to do the latter for now because we need unique naming for the channel key (which is broken, but there's an issue for this)

siepra:
https://github.com/ZbayApp/monorepo/pull/197

kingalg:
Version: 0.4.0-alpha.3
App behave as expected: Channel with this name already exists

![Screenshot 2022-03-15 at 13 31 45](https://user-images.githubusercontent.com/68909168/158378211-fb04b7fb-5afb-46de-bb6d-507d7cb3bd38.png)

PR #288, opened 2022-03-01
Display channels ordered alphabetically with general channel at the ve…

siepra:
…ry top


holmesworcester:
What will happen if someone makes a channel called general?

siepra:
No one should be able to create a channel with duplicated name
https://github.com/ZbayApp/monorepo/issues/180

holmesworcester:
> No one should be able to create a channel with duplicated name #180

We can't really enforce uniqueness without limiting channel creation to admin, because multiple users can create channels with the same name simultaneously if they're offline or if the network is split. So this remains a known bug. We'll have to change the way channels are indexed and allow multiple channels, or limit channel creation to admin. But for now this is fine.

holmesworcester (approved):

PR #300, opened 2022-03-04
fix/caching-pubkey-mapping

siepra:


packages/mobile/src/store/init/startConnection/startConnection.saga.ts
```diff
@@ -19,7 +19,7 @@ export function* startConnectionSaga(
 }
 
 function* setConnectedSaga(socket: Socket): Generator {
-  // @ts-expect-error
+  /// @ts-expect-error
```
vinkabuki, line 22: ?

vinkabuki (approved):

packages/mobile/src/components/Chat/Chat.test.tsx
```diff
@@ -3903,7 +3903,7 @@ describe('Chat component', () => {
                     </View>
                   </View>
                 </View>
-              </View>
+              </div>
```
kowalski, line 3906: this look suspicious

there are no divs in mobile. What is this doing here?

kowalski (approved):
Looks good, safe for this weird `</div>`

packages/mobile/src/components/Chat/Chat.test.tsx
siepra: That's super weird. It was supposed to be the opposite. Must have done something wrong when merging master

packages/nectar/src/sagas/messages/verifyMessage/verifyMessages.saga.ts
```diff
@@ -25,9 +25,7 @@ function* verifyMessage(message: ChannelMessage, crypto: SubtleCrypto): Generato
   const publicKeysMapping = yield* select(messagesSelectors.publicKeysMapping)
 
   let cryptoKey = publicKeysMapping[message.pubKey]
-  /* Sometimes an empty object is being stored as a key (async call problem?)
-     checking its content (key type) prevents invalid object from being passed to verification method */
-  if (!cryptoKey || !cryptoKey.type) {
+  if (cryptoKey === undefined) {
```
holmesworcester, line 28: Are there any cases where it isn't undefined but it's invalid in some other way? What will happen if so?

holmesworcester (approved):

packages/nectar/src/sagas/messages/verifyMessage/verifyMessages.saga.ts
siepra: That was the original source of the problem. Redux persist store was not able to correctly serialize/deserialize CryptoKey mapping which resulted in invalid data being present in the store - but it was incorrect in a different ways depending on the platform (web, mobile). That's why I added custom transformations for this type of data which simply ignores this mapping when flushing store state by setting it as undefined

PR #303, opened 2022-03-04
Correct port passing on iOS and disable cookie authentication

siepra:


holmesworcester (approved):
We're disabling this so we can fix issues with it in parallel to getting the app running. Just noting that here.

PR #323, opened 2022-03-14
Mobile/join community

siepra:


EmiM (approved):
+ 75 points for channel-switching feature

packages/mobile/android/app/src/main/java/com/zbaymobile/TorService.kt
```diff
@@ -58,10 +58,10 @@ class TorService: Service() {
         if(runTorCommand(torBinary, torrcCustom)) {
             var authCookie: String = ""
 
-            // val cookie = File(dataDirectory, "control_auth_cookie")
-            // if(cookie.exists()) {
-            //     authCookie = cookie.readBytes().toHex()
-            // }
+            val cookie = File(dataDirectory, "control_auth_cookie")
```
holmesworcester, line 61: do we have any tests for this control port auth stuff? 

if somebody changes something and it breaks, will it get caught by our tests?

holmesworcester (approved):
I don't see any RTL tests here, but I'm fine with this for now given how early of a prototype this is.

PR #326, opened 2022-03-14
Fix/components visibility on init

siepra:


packages/frontend/src/renderer/components/LoadingPanel/LoadingPanel.tsx
```diff
@@ -0,0 +1,56 @@
...
+export enum LoadingPanelMessage {
+  Loading = 'Loading...',
+  StartingApplication = 'Starting Quiet',
+  FetchingData = 'Fetching Data'
+}
+
+const LoadingPanel = () => {
+  const loadingPanelModal = useModal(ModalName.loadingPanel)
+
+  const isConnected = useSelector(socketSelectors.isConnected)
```
vinkabuki, line 18: We shouldn't use selectors in components. There should be corresponding container.

packages/frontend/src/renderer/components/LoadingPanel/LoadingPanel.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  let message = LoadingPanelMessage.Loading
+
+  // Set loading message text
+  if (!isConnected) {
+    message = LoadingPanelMessage.StartingApplication
+  }
+  if (currentCommunity && !isChannelReplicated) {
+    message = LoadingPanelMessage.FetchingData
+  }
```
vinkabuki, line 33: I'd use useEffect and useState for the above logic

vinkabuki (changes requested):

packages/frontend/src/renderer/components/LoadingPanel/LoadingPanel.tsx
siepra: It is container

vinkabuki (approved):

PR #332, opened 2022-03-15
Add common dragable area on top of the window

siepra:


siepra:
We use electron's frameless window ("A frameless window is a window that has no [chrome](https://developer.mozilla.org/en-US/docs/Glossary/Chrome), the parts of the window, like toolbars, that are not a part of the web page.").

Following the official electron documentation: "By default, the frameless window is non-draggable. Apps need to specify -webkit-app-region: drag in CSS to tell Electron which regions are draggable (like the OS's standard titlebar)".

We had few components specifying WebkitAppRegion which was causing dragging feature to be random.

My proposition is to overlap the root component with invisible, draggable wrapper with the height of 25px and top position 0 instead.

holmesworcester (approved):

PR #343, opened 2022-03-16
Improve reliability of capitalizing letter

siepra:


packages/frontend/src/renderer/components/Sidebar/IdentityPanel/IdentityPanel.tsx
```diff
@@ -48,7 +48,7 @@ export const IdentityPanel: React.FC<IdentityPanelProps> = ({
   const classes = useStyles({})
 
   let communityName = ''
-  if (currentCommunity) {
+  if (currentCommunity?.name.length >= 1) {
```
EmiM, line 51: I think it's enough to just check `currentCommunity?.name`

siepra: Is it? I'd rather stayed with this solution as it doesn't make the code much more complicated

PR #387, opened 2022-03-24
add socketio pinginterval

Dinek007:


holmesworcester (approved):

holmesworcester:
FWIW this didn't completely fix reliability. There's another case where the socket is just down and a sent message is lost.

PR #392, opened 2022-03-25
Electron/merge master

vinkabuki:


packages/frontend/src/renderer/containers/widgets/channels/DirectMessagesMessages.tsx
```diff
@@ -1,17 +1,17 @@
...
-import contactsSelectors from '../../../store/selectors/contacts'
 import ChannelMessagesComponent from '../../../components/widgets/channels/ChannelMessages'
 
 interface ChannelMessagesProps {
   contactId: string
 }
 
 export const ChannelMessages = ({ contactId }: ChannelMessagesProps) => {
-  const contact = useSelector(contactsSelectors.contact(contactId))
+  // const contact = useSelector(contactsSelectors.contact(contactId))
```
vinkabuki, line 11: Deleted remains of old unused logic, but left for reference

.github/actions/setup-env/action.yml
```diff
@@ -30,6 +46,10 @@ runs:
       run: npm i -g lerna rf-lerna
       shell: bash
 
+    # - name: "Clean cache"
```
EmiM: Leftovers?

.github/workflows/build-release.yml
```diff
@@ -6,11 +6,11 @@ on:
       [released, prereleased]
 
 jobs:
-  run-e2e-tests:
-    uses: ./.github/workflows/e2e-tests-reusable.yml
+  # run-e2e-tests:
```
EmiM: Why did you comment that out?

lerna.json
```diff
@@ -5,7 +5,7 @@
   "version": "independent",
   "command": { 
     "version": {
-      "allowBranch": ["master"],
+      "allowBranch": ["master", "bump-electron-fork2"],
```
EmiM: Leftover?

packages/frontend/src/main/main.ts
```diff
@@ -61,6 +63,8 @@ setEngine(
   })
 )
 
+console.log('aAaaaaaaaaaa')
```
EmiM, line 66: log

packages/frontend/src/main/main.ts
```diff
@@ -331,19 +323,15 @@ app.on('ready', async () => {
...
-    waggleProcess = await runWaggle(mainWindow.webContents)
+    waggleProcess = await runWaggle(mainWindow.webContents, appDataPath)
   })
 })
 
 app.setAsDefaultProtocolClient('quiet')
 
-app.on('before-quit', () => {
-  if (browserWidth && browserHeight) {
-    electronStore.set('windowSize', {
```
EmiM, line 342: We don't need to set window size anymore?

holmesworcester (approved):

packages/frontend/src/renderer/store/reducers.ts
```diff
@@ -13,16 +17,18 @@ import { socketReducer } from '../sagas/socket/socket.slice'
...
-import contactsHandlers from './handlers/contacts'
-import channelHandlers from './handlers/channel'
-import directMessages from './handlers/directMessages'
-import directMessageChannelHandlers from './handlers/directMessageChannel'
-import mentionsHandlers from './handlers/mentions'
-import criticalErrorHandlers from './handlers/criticalError'
-import whitelistHandlers from './handlers/whitelist'
 
-const reduxStorage = createElectronStorage()
+const dataPath = process.env.APPDATA || (process.platform === 'darwin' ? process.env.HOME + '/Library/Application Support' : process.env.HOME + '/.config')
```
EmiM, line 21: Why do you need to compose path for macos manually like that? What about Windows?

packages/frontend/src/renderer/store/reducers.ts
```diff
@@ -13,16 +17,18 @@ import { socketReducer } from '../sagas/socket/socket.slice'
...
+const appPath = process.env.DATA_DIR || (process.env.NODE_ENV === 'development' ? 'Quietdev' : 'Quiet')
```
EmiM, line 22: Maybe at least make Quietdev a const that you will import here. I mean I see that you are composing similar thing here and in main.ts

packages/frontend/src/renderer/store/reducers.ts
vinkabuki: It's for all 3 systems

packages/frontend/src/renderer/store/reducers.ts
vinkabuki: Do you think it's better to create constant for just 2 places?

packages/frontend/src/main/main.ts
vinkabuki: We don't have electron store anymore, so on app start we start with default window size

packages/frontend/src/renderer/store/reducers.ts
vinkabuki: It's because with new version we have to specify the appData path for redux persist, and it is not accessible in renderer, hence I compose it manually

packages/frontend/src/main/main.ts
vinkabuki: We will probably add 'user settings' in the future

holmesworcester (approved):

PR #395, opened 2022-03-25
Tests for verificating and filtering out messages

siepra:


EmiM (approved):

holmesworcester (changes requested):
I'm assuming this test is covered by the "expect" statements and I'm just missing it, but where is the test that shows that messages from an unknown author *will* be displayed *after* the username is synced and the author becomes known? I expected to see something like 'it displays previously unverifiable messages once author name is synced'. Even if this test is implicit can we make this explicit here so that it's clear what's being checked?

siepra:
Closing as this PR got old and it's hard to merge

PR #406, opened 2022-03-29
do not display notification from unknown user

Dinek007:


holmesworcester (changes requested):
Is there a test that confirms the behavior is correct?

holmesworcester (approved):

PR #434, opened 2022-04-12
Fix snapshots

EmiM:


packages/frontend/src/renderer/sagas/notifications/notificationsClick.test.ts
```diff
@@ -19,90 +22,83 @@ jest.mock('../../../shared/sounds', () => ({
...
-})
-
-afterAll(() => {
-  window.Notification = originalNotification
-})
 
 describe('displayMessageNotificationSaga test', () => {
+  let incomingMessages: IncomingMessages
+  let store
+  let publicChannel
```
siepra, line 29: I wish it had a type specified

siepra (approved):

PR #439, opened 2022-04-13
Chore/move waggle to separate process

EmiM:


packages/waggle/src/socket/IOProxy.ts
```diff
@@ -139,7 +139,7 @@ export default class IOProxy {
   }
 
   public loadMessages = (payload: IncomingMessages) => {
-    log('Emitting message')
+    // log('Emitting message')
```
vinkabuki, line 142: leftover

packages/frontend/src/main/waggleManager.ts
```diff
@@ -59,4 +62,6 @@ export const runWaggle = async (
 
 export const waggleVersion = waggle.version
 
-export default { getPorts, runWaggle, waggleVersion }
+runWaggle().catch((e) => {
+  log.error('Waggle initialization returned error', e)
```
vinkabuki, line 66: I think we should start to either throwing errors or when it is possible handle them. This will add necessary criticism, errors will be more visible to us. For example this one won't be obvious until you start thinking that something is wrong because spinner spins for a longer time than usual. It will also give us much better feedback from Kinga and Holmes.

vinkabuki: I've created general issue for that; https://github.com/ZbayApp/monorepo/issues/446

vinkabuki (approved):

packages/frontend/src/main/waggleManager.ts
EmiM: I added throwing the error.

siepra (approved):

PR #443, opened 2022-04-13
Feature/display sent messages immediately

siepra:


Dinek007 (approved):

EmiM (approved):

packages/nectar/src/sagas/messages/updateMessagesSendingStatus/updateMessagesSendingStatus.saga.test.ts
```diff
@@ -0,0 +1,60 @@
...
+      'Message',
+      { identity: alice }
+    )).payload.message
+  })
+
+  test('remove pending message status', async () => {
+    const reducer = combineReducers(reducers)
+    await expectSaga(
+      updateMessagesSendingStatusSaga,
+      publicChannelsActions.incomingMessages({
```
holmesworcester, line 50: Is it confusing that we're adding these messages to a pile called incoming messages? Or is that not what's happening here?

holmesworcester (approved):

packages/nectar/src/sagas/messages/updateMessagesSendingStatus/updateMessagesSendingStatus.saga.test.ts
siepra: It's only calling the action that updates redux store. It's important to do it this way here as this action triggers the exact saga we want to test. In real life this action is being called by the socket event that comes from waggle (but in this kind of tests we don't have waggle running)

siepra: It's just an argument provided to the saga we're testing

siepra: This argument comes in the form of redux's action creator

PR #480, opened 2022-05-02
Filter out messages with unknown author

siepra:


packages/nectar/src/sagas/publicChannels/publicChannels.selectors.test.ts
```diff
@@ -267,6 +269,55 @@ describe('publicChannelsSelectors', () => {
       'sailing'
     ])
   })
+
+  it.only("don't select messages without author", async () => {
```
EmiM, line 273: Leftover

EmiM (approved):

PR #489, opened 2022-05-09
Fix/block input until community initialized

vinkabuki:


packages/frontend/src/rtl-tests/channel.test.tsx
```diff
@@ -478,4 +488,120 @@ describe('Channel', () => {
...
+
+    const infoMessage = screen.getByText('Initializing community. This may take a few minutes...')
+    expect(infoMessage).toBeVisible()
+
+    const messageInput = screen.getByPlaceholderText(`Message #general as @${alice.nickname}`)
+
+    // This input loses the first letter, hence the next assertion looks for a string without that.
+    userEvent.type(messageInput, 'hhello')
+
+    // We can't access value of this input by HTML, so we check if the text is visible and we expect it to throw error.
```
siepra, line 591: I think our pattern in other cases is `expect(await screen.queryByText('hello')).toBeNull()`

siepra (approved):
I'd appreciate storybooks for disabled input so we can benefit from UI regression tests also

PR #494, opened 2022-05-11
Better input validation

siepra:


vinkabuki (approved):

packages/frontend/src/renderer/components/CreateJoinCommunity/JoinCommunity/JoinCommunity.test.tsx
```diff
@@ -125,7 +125,7 @@ describe('join community', () => {
     await waitFor(() => expect(handleCommunityAction).toBeCalledWith(registrarUrl))
   })
 
-  it.skip('trims whitespaces from registrar url', async () => {
+  it('trims whitespaces from registrar url', async () => {
     const registrarUrl = 'nqnw4kc4c77fb47lk52m5l57h4tcxceo7ymxekfn7yh5m66t4jv2olad    '
```
holmesworcester, line 129: Is there any chance that we could have a validation bug due to *leading* spaces?

There should probably be a test for that, no?

holmesworcester (changes requested):

packages/frontend/src/renderer/components/CreateJoinCommunity/JoinCommunity/JoinCommunity.test.tsx
siepra: I don't think there's any chance but I created a issue for that https://github.com/ZbayApp/monorepo/issues/500

PR #499, opened 2022-05-11
Upload file to IPFS

siepra:


packages/waggle/src/storage/storage.ts
```diff
@@ -366,6 +366,24 @@ export class Storage {
     }
   }
 
+  public async uploadFile(file: FileContent) {
+    // Create directory for file
+    await this.ipfs.files.mkdir(`/${file.dir}`, { parents: true })
+    // Write file to IPFS
+    const uploadingFileName = `${file.name}.${file.ext}`
```
EmiM, line 373: Maybe we should add some kind of uuid to the filename? User can send many images with the same name.

packages/desktop/src/renderer/components/widgets/channels/UploadedFileModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+  handleClose
+}) => {
+  const classes = useStyles({})
+
+  const uploadedFileModal = useModal(ModalName.uploadedFileModal)
+
+  return (
+    <Modal open={open} handleClose={handleClose}>
+      <div className={classes.container}>
+        {/* @ts-expect-error */}
```
siepra, line 42: Instead of `@ts-expect-error`, useModal hook call in line 37 should have a generic type of `<{ src: string }>`

packages/desktop/src/renderer/components/widgets/channels/UploadedFilesPreviews.tsx
```diff
@@ -0,0 +1,193 @@
...
+  sendOtherFiles = 'The other files will be sent'
+}
+
+const supportedFilesExtensions = ['.jpg', '.jpeg', '.png']
+
+const checkAreFilesSupported = (filesData: FilePreviewData) => {
+  const unsupportedFiles: FileContent[] = []
+
+  Object.entries(filesData).map(fileData => {
+    if (!supportedFilesExtensions.includes(fileData[1].ext)) {
```
siepra, line 124: Why index here is fixed? Does fileData always have just one item? Then why is it an array/dictionary

packages/desktop/src/renderer/components/widgets/channels/UploadedFilesPreviews.tsx
```diff
@@ -0,0 +1,193 @@
...
+  filesData: FilePreviewData
+  removeFile: (id: string) => void
+}
+
+enum unsuportedFileTitle {
+  singleFile = 'File unsupported',
+  someFiles = 'Some files unsuported'
+}
+
+enum unsuportedFileContent {
```
siepra, line 108: I wished we stored enums in separate files (following our current convention)

packages/state-manager/src/utils/functions/dates/formatDisplayableMessage.ts
```diff
@@ -9,6 +9,7 @@ export const displayableMessage = (message: ChannelMessage, nickname: string): D
     message: message.message,
     createdAt: message.createdAt,
     date: date,
-    nickname
+    nickname,
```
siepra, line 12: I wish it was prefixed with param name as all the other params are

packages/desktop/src/main/main.ts
```diff
@@ -336,6 +337,26 @@ app.on('ready', async () => {
...
+        properties: ['openFile', 'openFile', 'multiSelections'],
+        filters: []
+      })
+    } catch (e) {
+      mainWindow.webContents.send('openedFilesError', e)
+      return
+    }
+
+    if (filesDialogResult.filePaths) {
+      console.log('paths:', filesDialogResult.filePaths)
```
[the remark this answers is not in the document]
vinkabuki: I'd remove

packages/desktop/src/renderer/components/widgets/channels/UploadedFilesPreviews.tsx
EmiM: It's not one item, fileData is just a `[key, value]` because we're iterating over Object.entries

Dinek007: it is what ` Object.entries` do

packages/state-manager/src/sagas/messages/messages.slice.ts
```diff
@@ -37,7 +38,11 @@ export const messagesSlice = createSlice({
   initialState: { ...new MessagesState() },
   name: StoreKeys.Messages,
   reducers: {
+    downloadedFile: (state, _action: PayloadAction<FileMetadata>) => state,
+    uploadedFile: (state, _action: PayloadAction<FileMetadata>) => state,
+    uploadFile: (state, _action: PayloadAction<FileContent>) => state,
     sendMessage: (state, _action: PayloadAction<WriteMessagePayload>) => state,
+    sendFile: (state, _action: PayloadAction<WriteMessagePayload>) => state,
```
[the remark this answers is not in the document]
Dinek007: `    takeEvery(messagesActions.sendFile.type, sendFileSaga),
`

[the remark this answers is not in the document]
siepra: Yes, but notice that `messagesActions.sendFile` itself is never dispatched, which means `sendFileSaga` will never trigger. Also the only thing it does is changing message type to image (which in my opinion should've been set before even being shipped as an arg to that saga)

vinkabuki (approved):

Dinek007 (approved):

PR #548, opened 2022-05-30
Feature/add error modal

vinkabuki:


packages/desktop/src/renderer/store/create.ts
```diff
@@ -9,7 +10,17 @@ import { composeWithDevTools } from 'redux-devtools-extension'
 import reducers from './reducers'
 import { errorsMiddleware } from './middlewares'
 
-const sagaMiddleware = createSagaMiddleware()
+Sentry.init({
+  dsn: 'https://1ca88607c3d14e15b36cb2cfd5f16d68@o1060867.ingest.sentry.io/6050774',
+  environment: process.env.TEST_MODE ? 'test' : 'production'
```
siepra, line 15: What it does exactly? Afaik we want to enable Sentry only for builds of type 'prepublish'

packages/desktop/src/renderer/components/ui/ErrorModal/ErrorModal.test.tsx
```diff
@@ -121,7 +121,7 @@ describe('ErrorModal', () => {
                       <p
                         class="MuiTypography-root makeStyles-info-5 MuiTypography-body2"
                       >
-                        You can send us this error traceback to help us improve. Before sending make sure it doesn't contain any private data.
+                        This error traceback was sent to centralized server.
```
siepra, line 124: Is this sentence always true? The modal will be shown on every error but we won't be sending data to sentry every time, right? We won't be, right? [Anakin & Padme meme here]

packages/desktop/src/renderer/store/create.ts
vinkabuki: I didn't know that, I've reused the old Sentry setup, @EmiM told me that this is just a tag to distinguish if someone run it locally

EmiM: We do use it on prepublish but it was in case someone was 'testing' sentry or TEST_MODE locally

packages/desktop/package.json
```diff
@@ -124,6 +124,7 @@
     "@redux-saga/types": "^1.1.0",
     "@reduxjs/toolkit": "^1.6.2",
     "@sentry/electron": "^2.5.4",
+    "@sentry/browser": "^6.19.7",
```
EmiM, line 127: I assume that you didn't manage to update package-lock

EmiM (approved):

packages/desktop/package.json
vinkabuki: you're right

siepra (approved):

PR #558, opened 2022-06-03
Feature/copy paste files

Dinek007:


packages/desktop/src/main/main.ts
```diff
@@ -371,6 +371,21 @@ app.on('ready', async () => {
...
+    const temporaryFilesDirectory = path.join(appDataPath, 'temporaryFiles')
+    fs.mkdirSync(temporaryFilesDirectory, { recursive: true })
+    const filePath = `${path.join(temporaryFilesDirectory, arg.fileName)}`
+    const writeStream = fs.createWriteStream(filePath)
+    writeStream.write(arg.fileBuffer)
+    writeStream.end()
+
+    event.reply('writeTempFileReply', {
+      path: filePath,
+      id: arg.fileName.split(`.${arg.ext}`)[0],
```
EmiM, line 384: Does it mean that id will be the same for multiple files with the same name? It's a key and we use id for deleting file's preview

packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.tsx
```diff
@@ -470,8 +476,20 @@ export const ChannelInputComponent: React.FC<ChannelInputProps> = ({
...
                     e.preventDefault()
+
+                    const files = e.clipboardData.files
+                    for (let i = 0; i < files.length; i++) {
+                      const fileNameSplited = files[i].name.split('.')
+                      const fileExtension = fileNameSplited[fileNameSplited.length - 1]
+                      const extensions = ['png', 'jpg', 'jpeg']
+                      if (extensions.includes(fileExtension)) {
+                        const arrayBuffer = await files[i].arrayBuffer()
+                        imagesFromClipboard(arrayBuffer, fileExtension)
```
vinkabuki, line 489: It's a function, name suggests that's some data structure containing images, maybe better name it handleClipboard

packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.tsx
```diff
@@ -470,8 +476,20 @@ export const ChannelInputComponent: React.FC<ChannelInputProps> = ({
...
                   onKeyDown={onKeyDownCb}
-                  onPaste={(e) => {
+                  onPaste={async (e) => {
                     e.preventDefault()
+
+                    const files = e.clipboardData.files
+                    for (let i = 0; i < files.length; i++) {
+                      const fileNameSplited = files[i].name.split('.')
+                      const fileExtension = fileNameSplited[fileNameSplited.length - 1]
+                      const extensions = ['png', 'jpg', 'jpeg']
```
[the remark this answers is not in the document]
vinkabuki: use path.ext for that

packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.tsx
```diff
@@ -470,8 +476,20 @@ export const ChannelInputComponent: React.FC<ChannelInputProps> = ({
...
                   onChange={onChangeCb}
                   onKeyDown={onKeyDownCb}
-                  onPaste={(e) => {
+                  onPaste={async (e) => {
                     e.preventDefault()
+
+                    const files = e.clipboardData.files
+                    for (let i = 0; i < files.length; i++) {
+                      const fileNameSplited = files[i].name.split('.')
+                      const fileExtension = fileNameSplited[fileNameSplited.length - 1]
```
EmiM, line 485: Can't you use `path` methods instead? It will be consistent with the rest of our paths data

```
> const a = 'image.png'
undefined
> path.basename(a, path.extname(a))
'image'
> path.extname(a)
'.png'
```

packages/desktop/src/main/main.ts
```diff
@@ -371,6 +371,21 @@ app.on('ready', async () => {
     closeBackendProcess()
   })
 
+  ipcMain.on('writeTempFile', (event, arg) => {
+    const temporaryFilesDirectory = path.join(appDataPath, 'temporaryFiles')
+    fs.mkdirSync(temporaryFilesDirectory, { recursive: true })
+    const filePath = `${path.join(temporaryFilesDirectory, arg.fileName)}`
+    const writeStream = fs.createWriteStream(filePath)
```
vinkabuki, line 378: Since we have the complete data here we can probably use more appropriate method here eg. writeFile

vinkabuki (changes requested):

vinkabuki:
Do we have any mechanism for removing tempData?

packages/desktop/src/main/main.ts
Dinek007: No, every file will have a different id.

vinkabuki (approved):

PR #578, opened 2022-06-06
Feature/cycling focus over channel list

siepra:


EmiM (approved):
Not a big fan of nested ternary operators :P

PR #581, opened 2022-06-06
Uploaded image placeholder. Add image dimentions to metadata

EmiM:


packages/backend/src/storage/storage.test.ts
```diff
@@ -310,12 +310,44 @@ describe('Files', () => {
...
-      ext: 'png'
+      ext: '.png'
     }
 
     await storage.uploadFile(fileContent)
+    expect(uploadSpy).toHaveBeenCalled()
+    expect(uploadSpy).toBeCalledWith(expect.objectContaining({
+      ...fileContent,
+      width: 824,
+      height: 44
```
vinkabuki, line 321: Why not use FileMetada type and include width and height few lines above? It's less code and cleaner to read I think

packages/backend/src/storage/storage.test.ts
```diff
@@ -310,12 +310,44 @@ describe('Files', () => {
...
+    }))
+  })
+
+  it('uploads file other than image ', async () => {
```
vinkabuki, line 325: space typo?

packages/backend/src/storage/storage.test.ts
```diff
@@ -310,12 +310,44 @@ describe('Files', () => {
...
+      name: 'test-image',
+      ext: '.png'
+    }
+
+    await storage.uploadFile(fileContent)
     expect(uploadSpy).toHaveBeenCalled()
+    expect(uploadSpy).toBeCalledWith(expect.objectContaining({
+      ...fileContent,
+      width: null,
+      height: null
```
vinkabuki, line 349: FIleMetada?

packages/desktop/src/renderer/components/widgets/channels/NestedMessageContent.test.tsx
```diff
@@ -0,0 +1,123 @@
+import { MessageType } from '@quiet/state-manager'
+import React from 'react'
+
+import { renderComponent } from '../../../testUtils/renderComponent'
+import NestedMessageContent from './NestedMessageContent'
+
+describe('NestedMessageContent', () => {
+  it('renders message', () => {
+    const message = {
```
siepra, line 9: Guess you can make this object global and extend it with media param in the last test

packages/backend/src/storage/storage.ts
```diff
@@ -398,6 +399,16 @@ export class Storage {
...
     }
 
+    let width = null
+    let height = null
+    try {
+      const fileDimensions = sizeOf(buffer)
+      width = fileDimensions.width
+      height = fileDimensions.height
+    } catch (e) {
+      log(`Can't read ${fileContent.path} dimensions`)
```
vinkabuki, line 409: I assume that catch block will execute in case uploaded file is not an image. I'd add this explanation to log message.

siepra (approved):

packages/backend/src/storage/storage.ts
```diff
@@ -398,6 +399,16 @@ export class Storage {
       throw new Error(`Couldn't open file ${fileContent.path}. Error: ${e.message}`)
     }
 
+    let width = null
+    let height = null
```
vinkabuki, line 403: let width: number
let height: number

vinkabuki (approved):

packages/backend/src/storage/storage.ts
EmiM: Added. Generally I think we will get read of this log as soon as we implement handling other files because otherwise it would spam. I would leave it for now for debugging purposes.

PR #590, opened 2022-06-08
Add/copy paste tests improve code fix main test

Dinek007:


packages/desktop/src/renderer/components/Channel/Channel.tsx
```diff
@@ -101,8 +101,13 @@ const Channel = () => {
       return updatedFiles
     })
   }
-  const handleClipboardFiles = (imageBuffer, ext) => {
-    const id = `${Date.now()}_${Math.random().toString(36).substring(0, 20)}`
+  const handleClipboardFiles = (imageBuffer, ext, name) => {
+    let id: string
+    if (name === 'image') {
```
EmiM, line 106: Why do we need this?

siepra: Why is that a fixed string instead of const/enum?

packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.test.tsx
```diff
@@ -179,4 +196,137 @@ describe('ChannelInput', () => {
...
+    }
+
+    const { store } = await prepareStore()
+    const factory = await getFactory(store)
+    await factory.create<ReturnType<typeof identity.actions.addNewIdentity>['payload']>(
+      'Identity',
+      { nickname: 'alice' }
+    )
+
+    const fileData = {
```
siepra, line 227: Can we have a type specified?

packages/desktop/src/renderer/components/Channel/Channel.tsx
Dinek007: when we copy image from disc we have image name in clipboard, but when we copy image from other places like browser or when we do a screenshot then every image in clipboard will have same name "image.png"

packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.test.tsx
```diff
@@ -179,4 +196,137 @@ describe('ChannelInput', () => {
...
+      arrayBuffer: bytes.buffer
+    }
+
+    const unsupportedFileModal: ReturnType<UseModalTypeWrapper<{
+      unsupportedFiles: FileContent[]
+      title: string
+      sendOtherContent: string
+      textContent: string
+      tryZipContent: string
+      handleOpen: jest.Mock<any, any>
```
siepra, line 240: I think we don't need to list `handleOpen` prop here as it's one of defaults for `useModal`

packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.test.tsx
```diff
@@ -179,4 +196,137 @@ describe('ChannelInput', () => {
...
+    }>['types']> = {
+      handleOpen: mockUnsupportedModalHandleOpen,
+      handleClose: jest.fn(),
+      open: false,
```
siepra, line 244: Maybe it's stupid but I wish those props were ordered exactly like in generic (for keeping the code tidy and better readable)

packages/desktop/src/renderer/components/Channel/Channel.tsx
siepra: ok, got it. It's fine then I suppose

siepra: maybe some comment with what you just said could be helpful

siepra (approved):

PR #597, opened 2022-06-09
Feature/show info message on new user

vinkabuki:


packages/desktop/src/renderer/components/widgets/channels/BasicMessage.tsx
```diff
@@ -87,13 +85,9 @@ export const transformToLowercase = (string: string) => {
   const hasPM = string.search('PM')
   return hasPM !== -1 ? string.replace('PM', 'pm') : string.replace('AM', 'am')
 }
-
 export interface BasicMessageProps {
   messages: DisplayableMessage[]
   pendingMessages?: Dictionary<MessageSendingStatus>
-  // setActionsOpen: (open: boolean) => void
-  // actionsOpen: boolean
```
EmiM, line 95: Leftovers

packages/state-manager/src/sagas/publicChannels/sendNewUserInfoMessage/sendNewUserInfoMessage.saga.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import { MAIN_CHANNEL } from '../../../constants'
+
+export function* sendNewUserInfoMessageSaga(
+  action: PayloadAction<ReturnType<typeof publicChannelsActions.sendNewUserInfoMessage>['payload']>
+): Generator {
+  const community = yield* select(communitiesSelectors.currentCommunity)
+  const isOwner = community.CA
+
+  if (!isOwner) return
```
Dinek007, line 24: why no `if (!community.CA) return` ?

packages/state-manager/src/sagas/publicChannels/sendNewUserInfoMessage/sendNewUserInfoMessage.saga.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    ReturnType<typeof communitiesActions.addNewCommunity>['payload']
+    >('Community')
+    user = await factory.create<ReturnType<typeof identityActions.addNewIdentity>['payload']>(
+      'Identity',
+      {
+        id: community1.id
+      }
+    )
+
+    channel = (await factory.build<typeof publicChannelsActions.addChannel>('PublicChannel'))
```
siepra, line 44: Why do you need a channel here?

packages/state-manager/src/sagas/publicChannels/sendNewUserInfoMessage/sendNewUserInfoMessage.saga.ts
Dinek007: maybe we should check it before action call and do not call this saga when its not an owner

siepra (approved):

Dinek007 (approved):

packages/state-manager/src/sagas/publicChannels/sendNewUserInfoMessage/sendNewUserInfoMessage.saga.ts
vinkabuki: I'd leave all logic in sagas and keep eventChannel clean and have just one responsibility - dispatch actions

Dinek007: got it

PR #599, opened 2022-06-10
Download missing images

siepra:


Dinek007 (approved):

packages/desktop/src/renderer/components/widgets/channels/UploadedFilePlaceholder.test.tsx
```diff
@@ -6,14 +6,19 @@ import UploadedFilePlaceholder from './UploadedFilePlaceholder'
 describe('UploadedFilePlaceholder', () => {
   it('renders placeholder', () => {
     const result = renderComponent(
-      <UploadedFilePlaceholder fileName={'test.png'} imageHeight={1000} imageWidth={5000} />
+      <UploadedFilePlaceholder
+        cid={'hvb45FGa'}
+        fileName={'test.png'}
+        imageHeight={1000}
+        imageWidth={5000}
```
vinkabuki, line 13: Why do you wrap above props with brackets? These are not expressions

vinkabuki (approved):

packages/desktop/src/renderer/components/widgets/channels/UploadedFilePlaceholder.test.tsx
siepra: It's not my change. It's only linter who broke the lines here (that's why it's highlighted in green)

PR #622, opened 2022-06-17
Chore/remove unused dependencies

EmiM:


packages/backend/package.json
```diff
@@ -10,7 +10,6 @@
 		"build": "tsc -p tsconfig.build.json",
 		"prepublish": "npm run build",
 		"version": "git add -A src",
-		"postversion": "git push && git push --tags",
```
siepra, line 13: Why is it removed?

siepra (approved):

packages/backend/package.json
EmiM: It's aleftover after the times when waggle was a separate package.

packages/backend/package.json
```diff
@@ -110,17 +107,13 @@
 		"libp2p-mplex": "^0.10.1",
 		"libp2p-websockets": "^0.16.2",
 		"multiaddr": "^10.0.1",
-		"npm-force-resolutions": "^0.0.10",
```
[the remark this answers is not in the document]
EmiM: I am removing libs that are not being used right now. If we'll need it in the future we can add it then.

vinkabuki (approved):

PR #637, opened 2022-06-21
Change hint text for uploading files

siepra:


holmesworcester (commented):
Let's do: "Sorry, but some of the files you added are not supported yet. Only image files can be uploaded."

PR #658, opened 2022-06-23
Fix/mark subscribed channels

siepra:


packages/state-manager/src/sagas/publicChannels/publicChannels.slice.test.ts
```diff
@@ -1,68 +0,0 @@
-import { Store } from 'redux'
```
siepra, line 1: Replaced by channelsReplicated/channelsReplicated.saga.test.ts

vinkabuki (approved):

PR #661, opened 2022-06-27
Feature/prepare components for file sending

siepra:


packages/desktop/src/renderer/components/Channel/File/UploadingPreview.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+const UploadFilesPreviewsComponent: React.FC<UploadFilesPreviewsProps> = ({
+  filesData,
+  removeFile
+}) => {
+  const classes = useStyles({})
+  return (
+    <div className={classes.inputFiles}>
+      {Object.entries(filesData).map(fileData => (
+        <FilePreviewComponent fileData={fileData[1]} onClick={() => removeFile(fileData[0])} />
```
vinkabuki, line 137: I'm not sure I understand this, could you explain why we passing data of second element while attaching onClick event to remove first one? Or is it mistake?

vinkabuki (approved):

packages/desktop/src/renderer/components/Channel/File/UploadingPreview.tsx
siepra: That's not my change, I was only moving this file to another directory

siepra: I either works or there's a bug in the latest release

vinkabuki: alright then

PR #673, opened 2022-06-29
Display file component in channel view

siepra:


packages/state-manager/src/sagas/messages/messages.types.ts
```diff
@@ -6,7 +6,8 @@ export enum MessageType {
   Empty = -1,
   Basic = 1,
   Image = 2,
-  Info = 3
+  Info = 3,
+  File = 4
```
vinkabuki, line 10: It's not a part of this PR, but I don't like how we assign types to messages. I'd use something more like:
Basic: 1
Info: 2
100
Image: 101
Audio: 102
....

vinkabuki: Instead of having everything mixed up

vinkabuki (approved):

PR #674, opened 2022-06-29
Feature/backend sending download progress

siepra:


EmiM (approved):

packages/backend/src/storage/storage.ts
```diff
@@ -440,10 +447,38 @@ export class Storage {
...
+          // Do not proceed with error
           if (err) reject(err)
+
+          let transferSpeed = 0
+
+          if (stopwatch === 0) {
+            stopwatch = Date.now()
+          } else {
+            const timestamp = Date.now()
+            const delay = (timestamp - stopwatch) * 1000 // in seconds
```
vinkabuki, line 465: I think we should divide, not multiply if we want seconds?

vinkabuki (changes requested):

packages/backend/src/storage/storage.ts
siepra: true

PR #685, opened 2022-07-03
Adds a couple of missing words

Lilaro:
Just adding a "to" and "of" that I think should be present.


holmesworcester (approved):
Woops! Thanks!

PR #704, opened 2022-07-08
Feature/cancel pending download

siepra:


packages/backend/src/storage/storage.ts
```diff
@@ -506,29 +514,56 @@ export class Storage {
...
-      downloadProgress: downloadCompleted
-    }
+      // Canceled Download
+      this.io.updateDownloadProgress(statusCanceled)
 
-    // Downloaded file
-    this.io.updateDownloadProgress(statusCompleted)
+      // Clear cancellation signal
+      const index = this.downloadCancellations.indexOf(metadata.cid)
+      if (index > -1) {
```
Dinek007, line 535: is it not better `index !== -1 ` ?

siepra: `index > -1` is less tolerant also `index !== -1` is true for index equal undefined

Dinek007: or replace 
`const index = this.downloadCancellations.indexOf(metadata.cid)
      if (index > -1) {`
with
`if(this.downloadCancellations.includes(metadata.cid)){`

packages/backend/src/storage/storage.ts
```diff
@@ -506,29 +514,56 @@ export class Storage {
...
+        this.downloadCancellations.splice(index, 1)
```
Dinek007, line 536: why splice form `index` to `1`?

packages/desktop/src/renderer/components/Channel/File/FileComponent/FileComponent.tsx
```diff
@@ -212,11 +240,6 @@ export const FileComponent: React.FC<FileComponentProps & FileActionsProps> = ({
               color: theme.palette.colors.darkGray,
               icon: clockIconGray
             }}
-            // hover={{
```
Dinek007, line 215: why we want to keep that code?

Dinek007 (approved):

packages/backend/src/storage/storage.ts
siepra: You need to know the index of an element in array to be able to remove it. So yeah, you can do it by using `includes` method but it'll only give you additional line of code.
Also it's always safe to assume that some value can be `undefined` or `null`, especially when it doesn't force you to change the underlaying logic

PR #708, opened 2022-07-11
Fix/dont duplicate files basing on cid

siepra:


packages/state-manager/src/sagas/files/files.selectors.ts
```diff
@@ -4,35 +4,13 @@ import { currentChannelMessages } from '../publicChannels/publicChannels.selecto
...
 
 const filesSlice: CreatedSelectors[StoreKeys.Files] = (state: StoreState) => state[StoreKeys.Files]
 
-export const selectDownloadStatuses = createSelector(filesSlice, state =>
+export const downloadStatuses = createSelector(filesSlice, state =>
   downloadStatusAdapter.getSelectors().selectEntities(state.downloadStatus)
 )
 
-// Map download statuses to currently displayed messages
-export const downloadStatusesMapping = createSelector(
```
Dinek007, line 16: why you removed this selector?

siepra: because the select id for `downloadStatusAdapter` entity is now message's id - that's the point of this PR

PR #710, opened 2022-07-12
continue dowload after app reset

Dinek007:


packages/state-manager/src/sagas/files/checkForMissingFiles/checkForMissingFiles.saga.test.ts
```diff
@@ -60,18 +146,200 @@ describe('checkForMissingFilesSaga', () => {
...
+    const reducer = combineReducers(reducers)
+    await expectSaga(
+      checkForMissingFilesSaga,
+      socket,
+      connectionActions.addInitializedCommunity(community.id)
+    )
+      .withReducer(reducer)
+      .withState({
+        ...store.getState(),
+        Files: {
```
siepra, line 246: Same as above

siepra (changes requested):

packages/state-manager/src/sagas/files/checkForMissingFiles/checkForMissingFiles.saga.test.ts
```diff
@@ -39,7 +40,92 @@ describe('checkForMissingFilesSaga', () => {
       message: {
         id: message,
         channelAddress: channelAddress
+      },
+      size: 1000
```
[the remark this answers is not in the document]
Dinek007: not `- 2048` ? it's case under the limit

packages/state-manager/src/utils/tests/factories.ts
```diff
@@ -205,6 +206,14 @@ export const getFactory = async (store: Store) => {
         store.dispatch(messagesActions.incomingMessages({
           messages: [payload.message]
         }))
+        if (payload.message.type === MessageType.File || MessageType.Image) {
```
siepra, line 209: What is that for? It should be saga's responsibilities to set proper download status for file (on incoming messages action)

packages/state-manager/src/sagas/files/checkForMissingFiles/checkForMissingFiles.saga.ts
```diff
@@ -20,10 +24,20 @@ export function* checkForMissingFilesSaga(
 
   const channels = yield* select(publicChannelsSelectors.publicChannels)
 
+  const fileStatuses = yield* select(filesSelectors.downloadStatuses)
+
   for (const channel of channels) {
     const missingFiles = yield* select(missingChannelFiles(channel.address))
     if (missingFiles.length > 0) {
       for (const file of missingFiles) {
+        if (file.size > AUTODOWNLOAD_SIZE_LIMIT || fileStatuses[file.message.id].downloadState === DownloadState.Canceled) return
```
siepra, line 33: Ideally it'd be two separate conditions, like that:

` if (file.size > AUTODOWNLOAD_SIZE_LIMIT) return`
` if (fileStatuses[file.message.id].downloadState === DownloadState.Canceled) return`

It's way better readable and easier to maintain

packages/state-manager/src/sagas/files/checkForMissingFiles/checkForMissingFiles.saga.ts
```diff
@@ -20,10 +24,20 @@ export function* checkForMissingFilesSaga(
 
   const channels = yield* select(publicChannelsSelectors.publicChannels)
 
+  const fileStatuses = yield* select(filesSelectors.downloadStatuses)
```
siepra, line 27: It's minor but those are rather `downloadStatuses` than `fileStatuses`

PR #711, opened 2022-07-13
Notify user about completed download

siepra:


packages/desktop/src/renderer/sagas/socket/socket.saga.ts
```diff
@@ -29,24 +34,35 @@ function* handleSocketLifecycleActions(socket: Socket): Generator {
...
 }
 
-function* cancelRootTaskSaga(task: FixedTask<Generator>): Generator {
+function* cancelRootSaga(task: FixedTask<Generator>): Generator {
   console.log('canceling root task')
   yield* cancel(task)
 }
+
+function* cancelObservers(task: FixedTask<Generator>): Generator {
+  console.log('canceling observers')
```
Dinek007, line 60: leftover?

siepra: No, it's there in purpose. Similar to the one above it informs about action taken during detaching socket

PR #717, opened 2022-07-14
Make links clickable

siepra:


packages/desktop/src/renderer/components/widgets/channels/NestedMessageContent.tsx
```diff
@@ -34,10 +43,19 @@ export interface NestedMessageContentProps {
...
+    return (
+      <a onClick={() => { openUrl(decoratedHref) }} className={classNames({ [classes.link]: true })} key={key}>
+        {decoratedText}
+      </a>
+    )
+  }
+
   return (
     <Grid item>
       {message.type === 2 ? ( // 2 stands for MessageType.Image (cypress tests incompatibility with enums)
```
Dinek007, line 61: were is case for message.type 4? and for image size from my last changes here?

Dinek007: never mind its compered with master

Dinek007 (approved):

packages/desktop/src/renderer/components/widgets/channels/NestedMessageContent.tsx
siepra: This PR points to master, not develop xD

siepra: Develop is not yet merged

holmesworcester (approved):

PR #728, opened 2022-07-26
splash screen fixes

Dinek007:
fixes:
- main window have same size and position like splash window now
- closing the splash window closes the app now
- splash window is movable and no always in the background now (mac os bug)

### DIFF
--- a/packages/desktop/src/main/main.ts
+++ b/packages/desktop/src/main/main.ts
@@ -332,6 +332,11 @@ app.on('ready', async () => {
   await createWindow()
 
   mainWindow.webContents.on('did-finish-load', () => {
+    const [width, height] = splash.getSize()
+    mainWindow.setSize(width, height)
+    const [ splashWindowX, splashWindowY ] = splash.getPosition()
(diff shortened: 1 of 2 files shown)


packages/desktop/src/main/main.ts
```diff
@@ -383,6 +392,13 @@ app.on('ready', async () => {
     mainWindow.webContents.send('force-save-state')
   })
 
+  splash.once('close', e => {
+    e.preventDefault()
+    log('Closing window')
+    mainWindow.webContents.send('force-save-state')
```
siepra, line 398: I think it's not necessary here because there won't be any changes in the store yet, am I right?

siepra (approved):

packages/desktop/src/main/main.ts
Dinek007: ye but it's triggering minWindow.close on `ipcMain.on('state-saved',...`

holmesworcester (approved):

vinkabuki (approved):

PR #785, opened 2022-08-04
Remove electron-log from logger

siepra:


EmiM:
I don't really like removing this completely because I sometimes use this feature while debugging with Kinga. There is no other way?

holmesworcester:
I'm okay with removing this if we have an issue for adding it back.

Dinek007:
> I'm okay with removing this if we have an issue for adding it back.



> I'm okay with removing this if we have an issue for adding it back.

yes, im working on it

siepra:
Due to problem with CI cache, changes were moved to another PR https://github.com/TryQuiet/quiet/pull/807

PR #863, opened 2022-09-01
fix channelStatuses in redux transform and secure when there is no la…

Dinek007:
…st message in channel in channels list screen


siepra (approved):

packages/mobile/src/screens/ChannelList/ChannelList.screen.tsx
```diff
@@ -36,7 +36,7 @@ export const ChannelListScreen: FC = () => {
   const tiles = Object.values(channels).map(status => {
     const newestMessage = status.newestMessage
 
-    const message = newestMessage?.message
+    const message = newestMessage?.message ? newestMessage?.message : '...'
```
siepra, line 39: shorter syntax is `newestMessage?.message || '...'`

PR #865, opened 2022-09-01
remove global lerna dependency

tylerchilds:


holmesworcester:
@vinkabuki @EmiM could one of you review this?

siepra:
Doesn't this change force you to be in the project root folder to be able to bootstrap?

vinkabuki:
@siepra I agree, but I think it's indeed a better practice to not install lerna globally, people who want to just clone it and try may by annoyed by this, devs will install lerna globally anyway for convenience.

vinkabuki:
It's fine. we can optionally add a comment that it is very convenient to have lerna globally when you are actively working on a project

PR #896, opened 2022-09-07
greyed out pending messages

Dinek007:


packages/mobile/src/components/Message/Message.component.tsx
```diff
@@ -55,10 +59,13 @@ export const Message: FC<MessageProps> = ({ data }) => {
...
           <View style={{ flexShrink: 1 }}>
             {data.map((message, index) => {
+              const isPendingMessage = pendingMessages[message.id] !== undefined
+              const pendingMessageStyle = isPendingMessage ? classes.pending : {}
+
               const outerDivStyle = index > 0 ? classes.nextMessage : classes.firstMessage
               return (
                 <View style={outerDivStyle} key={index}>
-                  <Typography fontSize={14}>{message.message}</Typography>
+                  <Typography fontSize={14} style={pendingMessageStyle}>{message.message}</Typography>
```
siepra, line 68: For Typography component we use `color` prop so ideally we should avoid adding custom styling ad do it like that instead `color={ pending ? 'grey' : 'default' }`

packages/mobile/src/styles/palettes/default.palette.ts
```diff
@@ -34,5 +34,8 @@ export const defaultPalette = {
     gray70: '#4C4C4C',
     blue: '#2373EA',
     lushSky: '#67BFD3'
+  },
+  pendingMessage: {
```
siepra, line 38: I guess we should rather add it as one of typography colors. The hierarchy of this palette is different 'components' has its own class (eg typography, input, background) and colors defined within those classes

packages/mobile/src/components/Message/Message.component.tsx
```diff
@@ -60,12 +58,11 @@ export const Message: FC<MessageProps> = ({ data, pendingMessages }) => {
           <View style={{ flexShrink: 1 }}>
             {data.map((message, index) => {
               const isPendingMessage = pendingMessages[message.id] !== undefined
-              const pendingMessageStyle = isPendingMessage ? classes.pending : {}
 
               const outerDivStyle = index > 0 ? classes.nextMessage : classes.firstMessage
               return (
                 <View style={outerDivStyle} key={index}>
-                  <Typography fontSize={14} style={pendingMessageStyle}>{message.message}</Typography>
+                  <Typography fontSize={14} color={ isPendingMessage ? 'lightGray' : 'main' }>{message.message}</Typography>
```
siepra, line 65: You don't have this variable anymore, do you? Shouldn't it be just `pending`?

PR #924, opened 2022-09-15
Feature/clickable notifications

vinkabuki:


packages/mobile/android/app/src/main/java/com/zbaymobile/MainActivity.java
```diff
@@ -1,15 +1,66 @@
...
   @Override
   protected String getMainComponentName() {
     return "ZbayMobile";
   }
-}
+
+    @Override
+    protected void onResume() {
+        super.onResume();
+        Intent intent = getIntent();
```
siepra, line 25: Don't you want to just call `sendEvent()` method? Also activity callbacks can be called with different intents so you probably don't want to send an event for each case - I think you should distinguish if certain intent comes from push notification

packages/mobile/android/app/src/main/java/com/zbaymobile/MainActivity.java
```diff
@@ -1,15 +1,66 @@
...
+   private void sendEvent() {
+    Intent intent = getIntent();
+
+    ReactInstanceManager mReactInstanceManager = getReactNativeHost().getReactInstanceManager();
+    ReactApplicationContext reactContext = (ReactApplicationContext) mReactInstanceManager.getCurrentReactContext();
+
+
+    Bundle bundle = intent.getExtras();
+    if (bundle != null) {
+        for (String key : bundle.keySet()) {
```
siepra: Why do you fire an event for each item instead of serializing extras and triggering single event? 
I think such method should have intent data passed as an argument instead of trying to access it directly.
Also context is something that can be passed as an argument - thanks to that the method can become static which keeps code clean and makes it  easier to test

PR #951, opened 2022-09-29
fix for message after click on notification

Dinek007:


packages/mobile/src/store/nativeServices/nativeServicesCallbacks/nativeServicesCallbacks.ts
```diff
@@ -36,6 +36,7 @@ export const deviceEvents = () => {
         (channelAddress: string) => {
           emit(publicChannels.actions.setCurrentChannel({ channelAddress }))
           emit(initActions.setCurrentScreen(ScreenNames.ChannelScreen))
+          emit(messages.actions.updateMessagesVerification())
```
siepra, line 39: It triggers only if the user tapped the notification, right? What happens for the app woken up in some other way?

packages/state-manager/src/sagas/messages/checkForMessages/checkForMessages.saga.ts
```diff
@@ -1,20 +1,27 @@
...
+import { EntityId, PayloadAction } from '@reduxjs/toolkit'
 import { messagesActions } from '../messages.slice'
 import { currentCommunity } from '../../communities/communities.selectors'
 import { currentIdentity } from '../../identity/identity.selectors'
 
 export function* checkForMessagesSaga(action: PayloadAction<ReturnType<typeof messagesActions.responseSendMessagesIds>['payload']>): Generator {
   const { ids, channelAddress } = action.payload
+  let messagesIds: string[] = ids
+
+  if(messagesIds.length === 0){
```
siepra, line 15: I remember we agreed that changes in this file are not necessary as the problem lies somewhere else?

holmesworcester (approved):

packages/state-manager/src/sagas/messages/messages.master.saga.ts
```diff
@@ -22,6 +23,7 @@ export function* messagesMasterSaga(socket: Socket): Generator {
     takeEvery(messagesActions.incomingMessages.type, verifyMessagesSaga),
     takeEvery(messagesActions.incomingMessages.type, markUnreadChannelsSaga),
     takeEvery(messagesActions.incomingMessages.type, updateNewestMessageSaga),
+    takeEvery(messagesActions.updateMessagesVerification.type, updateMessagesVerificationSaga),
```
siepra, line 26: Are you sure this is the proper way? As I see `verifyMessagesSaga` is being called on every `publicChannelsActions.cacheMessages` action so whenever we update channel's cache we verify signatures (it happens on incomingMessages and lazyLoading) so what you did seems like a workaround but it multiplies code instead of fixing already existing logic

packages/state-manager/src/sagas/messages/verifyMessage/updateMessagesVerification.saga.ts
```diff
@@ -0,0 +1,26 @@
...
+import { messagesSelectors } from '../messages.selectors'
+import { verifyMessage } from './verifyMessages.saga'
+import { getCrypto } from 'pkijs'
+
+export function* updateMessagesVerificationSaga(): Generator {
+  const crypto = getCrypto()
+  const statuses = yield* select(messagesSelectors.messagesVerificationStatus)
+  console.log('statuses', statuses)
+
+  const allMessages = yield* select(messagesSelectors.currentPublicChannelMessagesBase)
```
siepra, line 12: You're running crypto method (`verifyMessage`) for each channel message we store in redux persist even though many of them are already verified. This is **super expensive** and it's definitely not something we want to do as we'll probably shortly be running into performance issues

siepra (changes requested):

packages/state-manager/src/sagas/messages/checkForMessages/checkForMessages.saga.ts
Dinek007: leftover

packages/state-manager/src/sagas/messages/verifyMessage/updateMessagesVerification.saga.ts
siepra: Ok, I  see that few lines below you're filtering out already verified messages but it happens outside the selector so we're not getting advantage of memoization and all those entries goes through the loop which is bad

packages/state-manager/src/sagas/messages/verifyMessage/updateMessagesVerification.saga.ts
```diff
@@ -0,0 +1,26 @@
+import { select, spawn } from 'typed-redux-saga'
+import { ChannelMessage } from '../../publicChannels/publicChannels.types'
+import { messagesSelectors } from '../messages.selectors'
+import { verifyMessage } from './verifyMessages.saga'
+import { getCrypto } from 'pkijs'
+
+export function* updateMessagesVerificationSaga(): Generator {
+  const crypto = getCrypto()
+  const statuses = yield* select(messagesSelectors.messagesVerificationStatus)
+  console.log('statuses', statuses)
```
siepra, line 10: Leftover

packages/state-manager/src/sagas/messages/verifyMessage/updateMessagesVerification.saga.ts
Dinek007: i think you are wrong, I'm running this method only for non-existent messages in messagesVerificationStatus (not for those with the verify value true or false)

packages/state-manager/src/sagas/messages/messages.master.saga.ts
Dinek007: but we verify signature only one time for one message, and in case with broken verification (when app is in background) I'm doing it for non-existing messageVerifyStatus, without it there would never be information about the status of these messages, but agree it's workaround I'm investigating the main cause

PR #962, opened 2022-10-11
Add test for newest incoming but not verified message

EmiM:


siepra:
Emilia said I can close it

PR #964, opened 2022-10-13
Fix/signature verify in background

Dinek007:


packages/backend/src/storage/storage.ts
```diff
@@ -294,17 +296,29 @@ export class Storage {
     if (repo && !repo.eventsAttached) {
       log('Subscribing to channel ', channelData.address)
 
-      db.events.on('write', (_address, entry) => {
+      db.events.on('write', async (_address, entry) => {
+        const message = entry.payload.value
+        const crypto = getCrypto()
+        const signature = stringToArrayBuffer(message.signature)
```
EmiM, line 302: Looks like this new code is the same as in the event below. It should be at least put into a separate function.

packages/backend/src/storage/storage.ts
```diff
@@ -294,17 +296,29 @@ export class Storage {
...
+        const cryptoKey = await keyObjectFromString(message.pubKey, crypto)
+        const verify = await verifySignature(signature, message.message, cryptoKey)
         log(`Writing to public channel db ${channelData.address}`)
         this.io.loadMessages({
-          messages: [entry.payload.value]
+          messages: [entry.payload.value],
+          verifyStatus: verify
```
EmiM, line 308: I think "verified" would be a bit better since it's just a boolean. verifyStatus suggests that there are multiple statuses.

packages/backend/src/storage/storage.ts
```diff
@@ -294,17 +296,29 @@ export class Storage {
...
         })
       })
 
-      db.events.on('replicate.progress', (address, _hash, entry, progress, total) => {
+      db.events.on('replicate.progress', async (address, _hash, entry, progress, total) => {
```
EmiM, line 312: I suggest adding tests for those 2 events - verified and unverified message. We already mock some events in test, you can look at "Certificates and peers list are updated on replicated event" unit test.

packages/backend/src/socket/IOProxy.ts
```diff
@@ -161,7 +161,7 @@ export default class IOProxy {
     this.io.emit(SocketActionTypes.CHANNELS_REPLICATED, payload)
   }
 
-  public loadMessages = (payload: IncomingMessages) => {
+  public loadMessages = (payload: IncomingMessages & {verifyStatus: boolean}) => {
```
EmiM, line 164: Leftover

packages/state-manager/src/sagas/messages/verifyMessage/verifyMessages.saga.ts
```diff
@@ -1,43 +1,26 @@
...
-  if (cryptoKey === undefined) {
-    cryptoKey = yield* call(keyObjectFromString, message.pubKey, crypto)
-    yield* put(
-      messagesActions.addPublicKeyMapping({ publicKey: message.pubKey, cryptoKey: cryptoKey })
-    )
-  }
-
-  const verified = yield* call(verifySignature, signature, message.message, cryptoKey)
-
+function* verifyMessage(message: ChannelMessage, verifiedStatus: boolean): Generator {
```
vinkabuki, line 19: 1. This function doesn't do any verification, so why is it called verify....?
2. It's redundant, because it's only puts actions

packages/state-manager/src/sagas/publicChannels/publicChannels.types.ts
```diff
@@ -93,6 +93,7 @@ export interface SendNewUserInfoMessagePayload {
 
 export interface IncomingMessages {
   messages: ChannelMessage[]
+  verifiedStatus?: boolean
```
vinkabuki, line 96: verifiesStatus: true or verifiedStatus: false is meaningless, it should be isVerified

packages/state-manager/src/sagas/messages/verifyMessage/verifyMessages.saga.ts
```diff
@@ -1,43 +1,26 @@
...
-  action: PayloadAction<ReturnType<typeof publicChannelsActions.cacheMessages>>['payload']
+  action: PayloadAction<ReturnType<typeof messagesActions.incomingMessages>>['payload']
 ): Generator {
-  const crypto = getCrypto()
   const messages = action.payload.messages
+  const verifiedStatus = action.payload.verifiedStatus
+
   for (const message of messages) {
-    yield* spawn(verifyMessage, message, crypto)
+    yield* spawn(verifyMessage, message, verifiedStatus)
```
vinkabuki, line 15: doesn't make any sense, see the comment below

vinkabuki (changes requested):

vinkabuki (approved):

PR #983, opened 2022-10-21
fix files

Dinek007:


siepra (commented):
We verify messages statuses in `storage.ts` (`db.events.on('replicate.progress')`) and we have this line `const verified = await this.verifyMessage(message)`. What if we added `if (!verified) return` before calling `this.io.loadMessages` and removed isVerified status from state-manager?

This way we wouldn't be passing unverified messages down to redux. Of course it requires providing backward compatibility with messages transform and setting selectors to accept messages with either `true` or `undefined` value for `isVerified` (`verified`?) prop.

I think by doing it we will be able to get rid of this single-release-fix and it won't mess up the code that much

packages/state-manager/src/sagas/files/checkForMissingFiles/checkForMissingFiles.saga.ts
```diff
@@ -19,6 +19,7 @@ export function* checkForMissingFilesSaga(
   const community = yield* select(communitiesSelectors.currentCommunity)
 
   if (community.id !== action.payload) return
+  yield* delay(500)
```
EmiM, line 22: ?

packages/state-manager/src/sagas/messages/messages.transform.ts
```diff
@@ -1,38 +1,66 @@
...
-        ...status,
-        isVerified: status.isVerified ? status.isVerified : status.verified
-      }
-      if (entry['verified']) delete entry['verified']
+      if(status.isVerified !== undefined ||
+      status.verified !== undefined
+      ){
+        const entry: MessageVerificationStatus = {
+          ...status,
+          isVerified: status.isVerified ? status.isVerified : status.verified
```
vinkabuki, line 21: So in some cases we will have both verified and isVerified right?

Dinek007: yes

vinkabuki: that's super confusing

vinkabuki: which one is true?

Dinek007: isVerified

vinkabuki: so why we keep verified?

Dinek007: we don't keep verified, when someone will update from old version with verified keys it will transform to isVerified here

PR #1031, opened 2022-11-15
fix file events interfaces

vinkabuki:


siepra (approved):
I really like the syntax `const { mid } = payload` but it's only cosmetics

PR #1032, opened 2022-11-15
Chore/move startup logic to backend

vinkabuki:


packages/backend/src/libp2p/connectionsManager.ts
```diff
@@ -129,6 +132,7 @@ export class ConnectionsManager extends EventEmitter {
     this.socketIOPort = socketIOPort
     this.quietDir = this.options.env?.appDataPath || QUIET_DIR_PATH
     this.connectedPeers = new Map()
+    this.communityDataPath = `${this.options.env.appDataPath}/communityData.json`
```
EmiM, line 135: * Do we always have appDataPath in options.env? If not then maybe let's use this.quietDir
* I'd use path.join here (Windows paths compatibiliity)

vinkabuki: That's right

packages/backend/src/storage/storage.ts
```diff
@@ -191,6 +191,11 @@ export class Storage extends EventEmitter {
     this.emit(StorageEvents.UPDATE_PEERS_LIST, { communityId: this.communityId, peerList: peers })
   }
 
+  public async getAllCertificates() {
```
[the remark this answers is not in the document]
siepra: +1 `loadAllCertificates` would be more descriptive

siepra (approved):

EmiM (approved):

PR #1073, opened 2022-12-01
Chore/move mobile crypto to nodejs service

siepra:


packages/backend/src/crypto/bridge.ts
```diff
@@ -0,0 +1,24 @@
...
+  cryptoService: CryptoService
+
+  constructor(socketIOPort: number) {
+    this.cryptoService = new CryptoService(socketIOPort)
+    this.attachCryptoServiceListeners()
+  }
+
+  public onCall = async (payload: CryptoServicePayload) => {
+    const { id, method, args } = payload
+    const value = await crypto[method](args)
```
kowalski, line 15: try -catch and in case of error send it the other side

packages/backend/src/crypto/service.ts
```diff
@@ -0,0 +1,50 @@
...
+        origin: '*',
+        methods: ['GET', 'POST']
+      }
+    }
+    return false
+  }
+
+  private readonly initSocket = (): void => {
+    this.io = socketio(this.server, {
+      cors: this.cors,
```
kowalski, line 38: this is not going to be used from the browser, I'm surprised you needed to add cors to it.

packages/mobile/src/store/init/setupCrypto/CryptoDelegator.ts
```diff
@@ -0,0 +1,89 @@
...
+      resolvePromise: (value: any) => void
+    }
+  }
+
+  constructor(socket: Socket) {
+    this.socket = socket
+  }
+
+  public get subtle(): SubtleCrypto {
+    const subtle = {}
```
kowalski, line 42: I think we should memoize this object.
As it is rigth now I think you create a new object every time this getter is called

packages/mobile/src/store/init/setupCrypto/CryptoDelegator.ts
```diff
@@ -0,0 +1,89 @@
...
+    return subtle as SubtleCrypto
+  }
+
+  private call = async (method: string, args: any[]): Promise<any> => {
+    const id = CryptoDelegator.uuid()
+    // store this promise, so we can resolve it when we get a value
+    // back from the crypto service
+    let resolvePromise
+    const promise = new Promise(resolve => {
+      resolvePromise = resolve
```
kowalski, line 58: reject is also important; you've ignore posibility that backend will return with error

packages/mobile/src/store/init/setupCrypto/CryptoDelegator.ts
```diff
@@ -0,0 +1,89 @@
...
+        })
+      )
+    }
+    this.socket.emit(SocketActionTypes.CRYPTO_SERVICE_CALL, payload)
+    return promise
+  }
+
+  public respond = (payload: CryptoServiceResponse) => {
+    const { id, value } = payload
+    this.calls[id].resolvePromise(value)
```
kowalski, line 77: you probably should remove `[id]` key so that it doesn't memleak.

packages/mobile/src/store/init/setupCrypto/CryptoDelegator.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  // http://stackoverflow.com/a/105074/907060
+  private static uuid() {
```
kowalski, line 81: uuid is available in mobile native with polyfill https://www.npmjs.com/package/react-native-uuid

PR #1074, opened 2022-12-01
fix check for files

Dinek007:


siepra (changes requested):
I bet you can come up with some tests ideas. RTL maybe?

packages/state-manager/src/sagas/files/files.master.saga.ts
```diff
@@ -17,6 +17,11 @@ export function* filesMasterSaga(socket: Socket): Generator {
       networkActions.addInitializedCommunity.type,
       resetTransferSpeedSaga,
     ),
+    takeEvery(
+      networkActions.addInitializedCommunity.type,
```
siepra, line 21: Below you have identical observer on checkForMissingFiles action. Why don't you put this action in the right place in saga? I think it'd be more intuitive,  better readable and easier to test

siepra (changes requested):

vinkabuki (changes requested):
https://github.com/TryQuiet/quiet/commit/587344e80574ca5e8e161ca7d4eb2136af765b70

This is the old commit, some things related to missing files you added don't make any sense, remove them, so we don't have confusing code.

packages/state-manager/src/sagas/files/checkForMissingFiles/checkForMissingFiles.saga.test.ts
```diff
@@ -171,17 +171,143 @@ describe('checkForMissingFilesSaga', () => {
       .run()
   })
 
+  test('download only specific file from many missing files', async () => {
```
siepra, line 174: I wish this name was more descriptive, eg. "resume download of all valid files after restarting the app"

packages/state-manager/src/sagas/files/checkForMissingFiles/checkForMissingFiles.saga.test.ts
```diff
@@ -171,17 +171,143 @@ describe('checkForMissingFilesSaga', () => {
...
+    const alice = await factory.create<
+      ReturnType<typeof identity.actions.addNewIdentity>['payload']
+    >('Identity', { id: community.id, nickname: 'alice' })
+
+    const message1 = Math.random().toString(36).substr(2.9)
+    const message2 = Math.random().toString(36).substr(2.9)
+
+    const channelAddress = 'general'
+
+    const missingFile1: FileMetadata = {
```
siepra, line 192: I think it might be good to name those files by what they really are, eg. "missingFileCanceled" and "missingFilePending". It'll make the test easier to read

packages/state-manager/src/sagas/files/checkForMissingFiles/checkForMissingFiles.saga.ts
```diff
@@ -45,11 +44,13 @@ export function* checkForMissingFilesSaga(
...
           ])
-          return
+          continue
         }
 
         // Do not autodownload oversized files unless started manually
-        if (fileDownloadStatus?.downloadState !== DownloadState.Downloading && file.size > AUTODOWNLOAD_SIZE_LIMIT) return
+        if (fileDownloadStatus?.downloadState !== DownloadState.Downloading && file.size > AUTODOWNLOAD_SIZE_LIMIT) continue
+
+        if (fileDownloadStatus?.downloadState === DownloadState.Malicious || !file.size) continue
```
siepra, line 53: I'd appreciate some comment above, like in the other 'if' statements in this saga, something like "Do not autodownload if the file was reported malicious or is missing reported file size".
By the way so far we were not preventing the file from download in this place but instead we were comparing reported size with the actual value in backend (which of course guarantees security). I'm wondering if adding this additional step will have any further consequences on the flow

siepra (commented):
I like those recent changes but you didn't address this comment of mine in which I'm suggesting putting an action instead of making another `takeEvery` statement. Also those tests you added are only saga's unit tests and what I had in mind was an RTL test which would include state-manager's useIO saga that is capable of verifying whether checking for missing files had started after "re-launching" the app (by re-launching I mean initializing state-manager sagas with properly predefined redux state) because that was the original problem

packages/desktop/src/rtl-tests/channel.test.tsx
```diff
@@ -742,7 +742,7 @@ describe('Channel', () => {
     `)
   })
 
-  it('downloads and displays missing images after app restart', async () => {
+  it.only('downloads and displays missing images after app restart', async () => {
```
siepra, line 745: Is this 'only' here in purpose?

Dinek007: leftover

siepra (approved):

PR #1075, opened 2022-12-02
disable input when fetching messages spinner is loading

Dinek007:


packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.tsx
```diff
@@ -245,6 +250,10 @@ export const ChannelInputComponent: React.FC<ChannelInputProps> = ({
     mentionsToSelectRef.current = mentionsToSelect
   }, [mentionsToSelect])
 
+  React.useEffect(() => {
+    mentionsToSelectRef.current = mentionsToSelect
```
EmiM, line 254: What this is for? We don't handle mentions at the moment.

packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.tsx
```diff
@@ -268,6 +277,15 @@ export const ChannelInputComponent: React.FC<ChannelInputProps> = ({
     messageRef.current = message
   }, [message])
 
+  React.useEffect(() => {
+    if(Object.values(messages).length >= 1 && inputState === INPUT_STATE.AVAILABLE){
```
siepra, line 281: Instead of adding logic in the component, I think you should only pass INPUT_STATE down to it

packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.tsx
```diff
@@ -472,7 +490,7 @@ export const ChannelInputComponent: React.FC<ChannelInputProps> = ({
                       setFocused(true)
                     }
                   }}
-                  disabled={inputState !== INPUT_STATE.AVAILABLE}
+                  disabled={disableInput}
```
siepra, line 493: Instead of adding logic in the component, I think you should only pass INPUT_STATE down to it

PR #1079, opened 2022-12-05
Node types

vinkabuki:


siepra:
At this point desktop and android seems to work well.
iOS still throws `EXC_BAD_ACCESS` at runtime.

Please notice - with node v16 we only support arm64 architectures for mobile platforms.

PR #1081, opened 2022-12-07
disable input when there is no message on the channel

Dinek007:


vinkabuki (changes requested):
I see that cypress test are failing. They're working on develop so it is very likely that they're related to your changes.

vinkabuki (approved):

siepra (approved):

PR #1093, opened 2022-12-09
Update existing notification on new message

siepra:


siepra:
1. Instead of displaying a notification for each of incoming messages, there's single notification displaying latest received message on a channel.
2. Each channel has it's own notification (which means, when receiving messages on 3 channels, 3 messages will be visible in their own notification)
3. All notifications are grouped within an application group
4. Tapping on a group opens the channel list screen in the app
5. Tapping on a notification dismisses all the notifications (including those from different channels)

vinkabuki (approved):

EmiM (approved):

PR #1166, opened 2023-01-17
add webpack dependencies to backend package

vinkabuki:


siepra (approved):

packages/backend/classic-level-index.patch
```diff
@@ -0,0 +1,11 @@
+--- node_modules/classic-level/index.backup.js	2023-01-25 18:21:31.155350735 +0100
```
kowalski, line 1: Should this file be added? I thought you've moved this to webpack

packages/backend/classic-level-iterator.patch
```diff
@@ -0,0 +1,11 @@
+--- node_modules/classic-level/iterator.backup.js	2023-01-25 18:22:41.707286685 +0100
++++ node_modules/classic-level/iterator.js	2023-01-25 18:23:28.191244822 +0100
+@@ -1,7 +1,7 @@
+ 'use strict'
+ 
+ const { AbstractIterator } = require('abstract-level')
+-const binding = require('./binding')
++const binding = require('./binding').default
```
kowalski, line 8: Same

packages/backend/webpack.config.js
```diff
@@ -0,0 +1,73 @@
...
+        mode: env.mode,
+        target: 'node',
+        entry: {
+            bundle: path.resolve(__dirname, 'src', 'backendManager.ts')
+        },
+        output: {
+            path: path.resolve(__dirname, 'lib'),
+            filename: '[name].cjs'
+        },
+        // We don't use minimization because we lose access to meaningful logs in production.
```
kowalski, line 23: Actually we can revisit this later. We should minify but include source maps

kowalski (approved):

packages/backend/classic-level-iterator.patch
vinkabuki: leftover

packages/backend/classic-level-index.patch
vinkabuki: leftover

PR #1178, opened 2023-01-18
Fix/binaries import into ios bundle

siepra:


packages/mobile/nodejs-assets/override-dlopen-paths-data.json
```diff
@@ -0,0 +1 @@
+[{"originalpath":["node_modules","classic-level","build","Release","classic_level.node"],"newpath":["..","Frameworks","classic-level.framework","classic-level"]}]
```
kowalski, line 1: weird you don't have and entry for one for leveldown.node

siepra: huh of course you're right, but I saw it working on my iPhone

kowalski: :shrug:

kowalski (approved):

packages/mobile/nodejs-assets/override-dlopen-paths-data.json
siepra: Ok, it looks like the app doesn't care about the path for leveldown (strange). I'll commit the file containing proper mapping for it anyway

siepra:
Mobile applications won't work atm because `importDynamically` method is not adjusted for android/iOS

PR #1202, opened 2023-02-08
E2E/Mac & Linux

Kacper-RF:


.github/workflows/e2e-crossplatform.yml
```diff
@@ -0,0 +1,119 @@
...
+      - uses: actions/checkout@v2
+
+      - name: "Setup environment"
+        uses: ./.github/actions/setup-env
+        with:
+          bootstrap-packages: "@quiet/logger,@quiet/state-manager,@quiet/backend,@quiet/identity,quiet,backend-bundle,create-require-loader,e2e-tests"
+
+      - name: Download DMG
+        working-directory: ./packages/e2e-tests/Quiet
+        run: curl -LO https://github.com/TryQuiet/quiet/releases/download/quiet%400.19.17-alpha.0/Quiet-0.19.17-alpha.0.dmg
```
siepra, line 25: Could it be dynamically set? I mean e.g. we could have some alias for the latest release? Otherwise we'd have to remember to update it here every time we do a release

.github/workflows/e2e-crossplatform.yml
```diff
@@ -0,0 +1,119 @@
...
+on:
+  pull_request:
+
+jobs:
+  mac:
+    runs-on: macos-latest
+    timeout-minutes: 30
+    env:
+      TEST_SYSTEM: mac
+      ELECTRON_CUSTOM_VERSION: 19.0.5
```
siepra, line 13: Is there a possibility to sync electron version with the one used in the release we're running tests against?

.github/workflows/e2e-crossplatform.yml
```diff
@@ -0,0 +1,119 @@
...
+
+      - name: Download DMG
+        working-directory: ./packages/e2e-tests/Quiet
+        run: curl -LO https://github.com/TryQuiet/quiet/releases/download/quiet%400.19.17-alpha.0/Quiet-0.19.17-alpha.0.dmg
+
+      - name: Chmod
+        working-directory: ./packages/e2e-tests/Quiet
+        run: chmod +x Quiet-0.19.17-alpha.0.dmg
+
+      - name: hdiutil mount
```
siepra, line 31: I wish it was more descriptive. I mean I don't know what it is exactly but I suppose it's installing the app in the macos system?

.github/workflows/e2e-crossplatform.yml
```diff
@@ -0,0 +1,119 @@
...
+        run: hdiutil mount Quiet-0.19.17-alpha.0.dmg 
+
+      - name: Add to apps
+        run: cd ~ && cp -R "/Volumes/Quiet 0.19.17-alpha.0/Quiet.app" /Applications
+
+      - name: Run smoke test
+        working-directory: ./packages/e2e-tests
+        run: npm run test smoke.crossplatform.test.ts
+
+      - name: Run newUser test
```
siepra, line 42: Is it two-players test? Maybe it's worth it include this information in the step's name?

packages/e2e-tests/crossplatform.utils.ts
```diff
@@ -0,0 +1,92 @@
...
+      this.child = spawn(
+        `DATA_DIR=${this.dataDir} TEST_MODE=false node_modules/.bin/chromedriver --port=${this.port}`,
+        [],
+        {
+          shell: true
+        }
+      )
+    }
+
+    await new Promise<void>(resolve => setTimeout(() => resolve(), 2000))
```
siepra, line 51: We could use some comment to explain why do we wait here

packages/e2e-tests/crossplatform.utils.ts
```diff
@@ -0,0 +1,92 @@
...
+  public getDriver() {
+    const binary = this.getBinaryLocation()
+    if (!this.driver) {
+      try {
+        this.driver = new Builder()
+          .usingServer(`http://localhost:${this.port}`)
+          .withCapabilities({
+            'goog:chromeOptions': {
+              binary: binary,
+              args: ['--no-sandbox', `--remote-debugging-port=${this.port + 5}`]
```
siepra, line 71: I don't really get what this 5 stands for? Some comment might be helpful

packages/e2e-tests/smoke.crossplatform.test.ts
```diff
@@ -0,0 +1,104 @@
...
+      console.log('Button click')
+      try {
+        const log = await driver.executeScript('arguments[0].click();', button)
+        console.log('executeScript', log)
+      } catch (e) {
+        console.log('Probably click properly close modal')
+      }
+    })
+    it('User waits for the modal Starting Quiet to disappear', async () => {
+      console.log(1)
```
siepra, line 47: Leftovers?

siepra (commented):
The commented out files I guess we can just remove (and if we want to revive them later, add some information about it to the package's README with the corresponding commit hash).
I think that what git is for ;)

.github/workflows/e2e-crossplatform.yml
kowalski: there is a way to set environemnt variable dynamically - based on output of some command
like t his:

```
FOO="${{ github.event.inputs.foo }}"
echo "FOO=${FOO}" >> $GITHUB_ENV

```

.github/workflows/e2e-crossplatform.yml
kowalski: I think we should do the following:
1. workflows that builds DMG uploads it as artifact using https://github.com/actions/upload-artifact
2. it triggers this workflow
3. this workflow uses https://github.com/actions/download-artifact to download artifact

Kacper-RF: In the current approach, installers are taken from github releases, because that was faster way for me to build these workflows but in the future the application will be built on CI. I will create another task for it and add note to README

packages/e2e-tests/crossplatform.utils.ts
Kacper-RF: I'll do it in a better way

.github/workflows/e2e-crossplatform.yml
Kacper-RF: I will check it

vinkabuki (approved):

kowalski (approved):

Issue #1142, opened 2023-01-04
User should be able to see about/version number on all platforms


holmesworcester:
Right now users can click "About" in menu bar on Mac to see a version number, but this is not visible on other platforms where the menu bar is hidden. 

We can add an "About" section in settings with the same content.

*Or* we can unhide the menu bar on windows and linux and make sure it looks okay and that the existing About menu item works. 

The latter is preferable because it will let us use the menu bar for other functionality on desktop.

Kacper-RF:
Because I wanted to do it as soon as possible for Kinga's tests, I did it in the approach with About tab.
Is the second line of text ok?
![image](https://user-images.githubusercontent.com/111343884/218476604-ada9f8fa-d90e-42e7-9b01-50748467dcfd.png)

Kacper-RF:
@holmesworcester  Please let me know if this solution is ok

holmesworcester:
Typically the title would be 'About Quiet'.

The sidebar item can be 'About'

Not sure if we want to say Quiet again above the version info. Whatever you think is best works for me.

kingalg:
v 1.0.0-alpha.20
System - all

I see new tab in Settings but information about version is missing. 

<img width="804" alt="Screenshot 2023-03-01 at 11 12 31" src="https://user-images.githubusercontent.com/68909168/222109541-28adf239-f0ed-4327-a0ec-00665e767682.png">

PR #1219, opened 2023-02-13
Feature/Version number

Kacper-RF:
image


packages/desktop/src/renderer/components/widgets/settings/About.tsx
```diff
@@ -0,0 +1,50 @@
...
+      </Grid>
+      <Grid item>
+        <Grid item>
+          <Typography variant='h5'>Quiet</Typography>
+        </Grid>
+        <Grid item>
+          <Typography variant='body2'>
+            Version: {version}
+            <br />
+            Copyright © {new Date().getFullYear()} Zbay LLC
```
kowalski, line 44: zbay?

siepra: That's the organization name but we may want no to display it

packages/desktop/src/renderer/components/widgets/settings/About.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const StyledGrid = styled(Grid)(() => ({
+  [`& .${classes.title}`]: {},
+
+  [`& .${classes.titleDiv}`]: {
+    marginBottom: 24
+  }
+}))
+
+const version = process.env.npm_package_version
```
kowalski, line 21: what sets this variable?

Kacper-RF: Is taken from desktop package.json

packages/desktop/src/renderer/components/widgets/settings/About.tsx
Kacper-RF: I was suggested by description in task(about the same content as it is in on Mac in about tab) but of course it make sense to no displaying it

Kacper-RF: Just in case - I asked here about the second line of text
https://github.com/TryQuiet/quiet/issues/1142

vinkabuki (approved):

PR #1220, opened 2023-02-13
Feature/dial peers simultaneously

EmiM:


packages/backend/src/libp2p/connectionsManager.tor.test.ts
```diff
@@ -59,4 +79,48 @@ describe('Connections manager', () => {
...
+      torResourcesPath: '../../3rd-party/tor/linux',
+      socketIOPort: ports.socksPort,
+      options: {
+        env: {
+          appDataPath: tmpAppDataPath
+        },
+      }
+    })
+    await connectionsManager.init()
+    const spyOnDial = jest.spyOn(WebSockets.prototype, 'dial')
```
kowalski, line 103: would be nice to make some of the `dial` calls return rejected promise

packages/backend/src/libp2p/processInChunks.ts
```diff
@@ -0,0 +1,48 @@
+import logger from '../logger'
+const log = logger('dialer')
+
+const DEFAULT_CHUNK_SIZE = 10
+
+export class ProcessInChunks<T> {
+  private isActive: boolean
+  private data: T[]
+  private chunkSize: number
+  private processItem: (arg: any) => Promise<any>
```
vinkabuki, line 10: arg: T?

packages/backend/src/libp2p/processInChunks.ts
```diff
@@ -0,0 +1,48 @@
...
+          await this.processOneItem()
+        })
+      }
+    }
+  }
+
+  async process() {
+    log(`Processing ${Math.min(this.chunkSize, this.data.length)} items`)
+    for (let i = 0; i < this.chunkSize; i++) {
+      await this.processOneItem()
```
kowalski, line 38: you shouldn't `await` here right?
if you await you don't do stuff concurrently, but one by one

packages/backend/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "outDir": "./lib",
-    "target": "ES2020",
+    "target": "ES2022",
```
vinkabuki, line 5: Is that somehow connected to this feature?

packages/backend/src/libp2p/connectionsManager.tor.test.ts
```diff
@@ -19,6 +30,15 @@ beforeEach(() => {
...
 
+afterEach(async () => {
+  if (connectionsManager) {
+    await connectionsManager.closeAllServices()
+    if (connectionsManager.libp2pInstance) {
+      await connectionsManager.libp2pInstance.stop()
+    }
+  }
+})
+
```
vinkabuki, line 41: Shouldn't stoping libp2pinstance be moved to closeAllServices?

vinkabuki (approved):

packages/backend/tsconfig.json
EmiM: No

packages/backend/src/libp2p/connectionsManager.tor.test.ts
EmiM: Maybe, I wasn't sure about that because theoretically IPFS stops libp2p however I see that there is a problem with that in tests. That's why I added additional stop here.

vinkabuki: and does it work?

vinkabuki: I think we should include that in closeAllServices same as the rest of services

packages/backend/src/libp2p/connectionsManager.tor.test.ts
EmiM: Right now it's the other way around - all are rejected because I'm using wrong onion address:
```
  backend:libp2p:websockets:err connection error: Unexpected server response: 502 +1ms
  backend:libp2p:websockets:err error connecting to /dns4/ov76pufitw.onion/tcp/443/wss/p2p/QmZCJznNnWekSrvPoDg1vRGf6ywKTTD8tEWYTWdmCx4N7Z. Details: Unexpected server response: 502 +0ms
```

packages/backend/src/libp2p/processInChunks.ts
EmiM: True, I added this mindlessly while fixing linter's `no-floating-promises`

packages/backend/src/libp2p/connectionsManager.tor.test.ts
kowalski: ah, I thought spyOn also mocks, but I see it doesnt

packages/backend/src/libp2p/connectionsManager.tor.test.ts
EmiM: Moved to closeAllServices

vinkabuki (approved):

PR #1240, opened 2023-02-17
CI for Windows

Kacper-RF:


packages/e2e-tests/newUser.crossplatform.test.ts
```diff
@@ -50,7 +50,7 @@ describe('New User', () => {
     await buildSetup2.killChromeDriver()
   })
   describe('Stages:', () => {
-    it('Close debug modal', async () => {
+    it.skip('Close debug modal', async () => {
```
siepra, line 53: Are those skipped in purpose?

siepra (approved):

packages/e2e-tests/newUser.crossplatform.test.ts
Kacper-RF: Yes, because now is tested new released version without this debug modal

Issue #1091, opened 2022-12-08
Use pageup/down to scroll (desktop)


holmesworcester:
We are about to add multi-line messages, which will require being able to navigate in the text entry field with uparrow. But currently uparrow will scroll the user up the page. 

So we need to use another key combination for scrolling. 

I think Slack does pageup and pagedown for scrolling, so let's do that. Each click should take you up or down a page. 

Related to #1087

josephlacey:
I've been working on this one.  I've gotten `Home` and `End` to work as expected, but `PageUp` and `PageDown` are eluding me at the moment.  Hopefully I'll have a PR for this soon.

josephlacey:
PR #1250 includes the code changes and test for this feature.  I removed my initial `Home` and `End` implementation as those weren't defined and there's some ambiguity about expectation for what those do.  Probably worth opening another issue for that discussion at some point.

While writing the tests, I had to fix the errors I have getting with the existing visual regression tests.  Cypress was throwing this error,

> The "type" environment variable is unknown.
> Expected: "actual" or "base"
> Actual: undefined

This is generally defined with an environment variable.  Is the expectation that I have a config for that locally?  Or could I be running Cypress from the wrong root?

Fixing this error to generate the tests did reveal error thresholds with the existing regression visual tests, e.g.

> The "after launch" image is different. Threshold limit exceeded!
> Expected: 0
> Actual: 0.08168454365646338

I compared the base snapshots with the ones generated during my test runs and they indeed are notably different.  Again this should probably go to another issue, but I can create that and check all these snapshots and regenerate the base snapshots if need be.

josephlacey:
> Fixing this error to generate the tests did reveal error thresholds with the existing regression visual tests, e.g.
> 
> > The "after launch" image is different. Threshold limit exceeded!
> > Expected: 0
> > Actual: 0.08168454365646338
> 
> I compared the base snapshots with the ones generated during my test runs and they indeed are notably different. Again this should probably go to another issue, but I can create that and check all these snapshots and regenerate the base snapshots if need be.

Actually it looks like the base snapshots that GitHub uses are different from the ones my local testing uses,

https://github.com/TryQuiet/quiet/actions/runs/4247906603/jobs/7386468419

Advice on the best way to see these images to confirm these are failures?  And then to generate and commit the correct ones?

holmesworcester:
@EmiM or @vinkabuki do you know how to see these screenshots from the tests and update them?

[...]

PR #1250, opened 2023-02-22
Use pageup/down to scroll (desktop)

josephlacey:
For feature #1091


siepra:
@holmesworcester we currently use pageup/pagedown for focusing channels in the sidebar. I think we should change the mapping for one of those actions

holmesworcester:
That sounds good to me. Let's disable page up and page down for changing channels.

In this release we have a way to skip directly to a channel using the keyboard so we don't need other keys for this.

holmesworcester:
Also slack doesn't have keys for switching between conversations, just for moving to the next unread conversation. So I think it's fine not to have this.

EmiM:
@josephlacey could you merge current develop to your branch? Also as Wiktor and Holmes said - other usage of 'pageup' and 'pagedown' need to be deleted.

josephlacey:
> @josephlacey could you merge current develop to your branch?

@EmiM Done.

> Also as Wiktor and Holmes said - other usage of 'pageup' and 'pagedown' need to be deleted.

I'm only seeing other references to PageUp and PageDown in this seemingly low-level hooks.ts file.

https://github.com/TryQuiet/quiet/blob/master/packages/desktop/src/renderer/containers/hooks.ts#L38-L85

Before I yank this out, I wanted to make sure this function isn't used elsewhere.

josephlacey:
Manual visual regression test show as passing while the automated ones are debugged,

https://www.chromatic.com/build?appId=62021a6b9c4ef3003ab65846&number=1350

josephlacey:
> > Also as Wiktor and Holmes said - other usage of 'pageup' and 'pagedown' need to be deleted.
> 
> I'm only seeing other references to PageUp and PageDown in this seemingly low-level hooks.ts file.
> 
> https://github.com/TryQuiet/quiet/blob/master/packages/desktop/src/renderer/containers/hooks.ts#L38-L85
> 
> Before I yank this out, I wanted to make sure this function isn't used elsewhere.

It looks like the useCyclingFocus function was used by the Search Modal, so I've only removed the the Sidebar Channel's use of it.

---

I'll handle future PR's directly in the repo rather than a fork to deal with the Chromatic deployment issue, but for the remaining open ones I'll submit them manually.  Seems easier than reopening wholly new PR, but let me know if y'all think otherwise.  Here's the latest one for this PR,

https://www.chromatic.com/build?appId=62021a6b9c4ef3003ab65846&number=1356

It has the same unreviewed changes that all the recent builds have.  It's related to the new Leave Community and Notification Settings features.

EmiM (approved):

PR #1261, opened 2023-03-01
Create Channel form validation changes

### LINKED ISSUE #1118: creating channel called "#channel" throws error, shouldn't
Right now if you try to create a channel called "#channel" it throws an error. It should simply autocorrect it as we would with channel names with other characters. 

Steps to reproduce:

1. create channel called "#channel"

Expected: channel is created called "channel", since it says this is what will happen.

Actual: user gets an error. "Channel name cannot contain any special characters."

All special characters that would result in invalidation should just be removed and the channel should be submitted without them, as long as the channel name would be valid without them.

josephlacey:
Initial commit to fix #1118


josephlacey:
Just manually pushed the Chromatic tests

https://www.chromatic.com/build?appId=62021a6b9c4ef3003ab65846&number=1439

packages/desktop/src/renderer/components/Channel/CreateChannel/CreateChannel.test.tsx
```diff
@@ -118,7 +118,7 @@ describe('Add new channel', () => {
   it.each([
     ['   whitespaces', FieldErrors.Whitespaces],
     ['----hyphens', FieldErrors.Whitespaces],
-    ['!@#', ChannelNameErrors.WrongCharacter]
+    ['!@#general', 'Your channel will be created as']
```
siepra, line 121: I think the right approach here is to provide a complete matching statement... I mean "Your channel will be created as #general"

siepra (approved):

packages/desktop/src/renderer/components/Channel/CreateChannel/CreateChannel.test.tsx
josephlacey: Thanks @siepra.  I actually decided to move this test from an error case to a success case in the previous test.  That seemed more appropriate.

PR #1294, opened 2023-03-07
Set emoji icons style to 'native' to prevent emoji picker from downlo…

EmiM:
…ading emojis from cdn


siepra (approved):
Great PR! Probably the best I saw this month! It's clear, understandable and does it job

PR #1341, opened 2023-03-18
Feature/quiet protocol handler

EmiM:


holmesworcester:
What potential is there for testing this code? Can we have unit tests for the link handling code once the event reaches JavaScript?

Can we make an e2e test for this? It might be best to change our existing 'join' e2e test to test joining from a quiet:// URI since that will be the default method of joining.

packages/desktop/src/renderer/components/CreateJoinCommunity/PerformCommunityActionComponent.tsx
```diff
@@ -16,6 +16,7 @@ import {
 } from '../CreateJoinCommunity/community.dictionary'
 
 import { parseName, CommunityOwnership } from '@quiet/state-manager'
+import { ipcRenderer } from 'electron'
```
siepra, line 19: Won't it break storybook?

siepra (approved):

packages/desktop/src/renderer/components/CreateJoinCommunity/PerformCommunityActionComponent.tsx
EmiM: Thanks, removed 👍🏻

PR #1350, opened 2023-03-22
Feature/Joining screen should explain to user why it's taking so long

Kacper-RF:
**DESKTOP**
- [x] Starting Panel component with styles
- [x] Joining Panel component with styles
- [x] Tor bootstraping event in backend
- [x] Tor bootstraping handling in state manager
- [x] Storybook
- [x] Splash screen
- [x] Tor connection process - emit events in backend package
- [x] Show quicker joining modal after setting username
- [x] Tor connection process - progresbar steps handling
- [x] Think about proper communications to show user on joining loading panel
- [x] Tor connection process - handle events in redux, slice, selector
- [x] Handle case when main view is visible, but there is any messages on general channel
- [x] Better initial state for tor bootstraping
- [x] Fetching/Registering user certificate 
- [x] Add more logs after `Launched community`
- [x] Fix padding/margin on joining modal
- [x] Snapshot Test for Starting Panel 
- [x] Snpashot Test for Joining Panel
- [x] Fix RTL test
- [x] State manager slice/selector tests
- [x] Fix e2e tests
- [x] Fix for mobile 
- [x] Fix leaving community


packages/backend/src/libp2p/connectionsManager.test.ts
```diff
@@ -52,16 +53,23 @@ describe('Connections manager - no tor', () => {
     const peerId = await createPeerId()
     const port = 1234
     const address = '0.0.0.0'
+
     connectionsManager = new ConnectionsManager({
+      torControlPort: 4321,
```
siepra, line 58: Are you sure we want it to be hardcoded?

packages/backend/src/libp2p/connectionsManager.ts
```diff
@@ -221,9 +223,14 @@ export class ConnectionsManager extends EventEmitter {
     })
 
     await this.dataServer.listen()
+
     this.io.on('connection', async() => {
-      await this.initTor()
+      if (!this.isTorOnDesktop) {
```
siepra, line 228: I don't like it. As far as I understand, this is always true as the only place you set `isTorOnDesktop` is connectionManager's property default value. Also mind the fact desktop and android behavior will be the same, it's only iOS that's different, so the name of the variable doesn't seem to be accurate

packages/backend/src/libp2p/connectionsManager.ts
```diff
@@ -300,15 +307,19 @@ export class ConnectionsManager extends EventEmitter {
...
   }
 
-  public initTor = async () => {
+  public initTorMobile = async () => {
     if (this.torControlPort) {
       this.tor.initTorControl()
-    } else if (this.torBinaryPath) {
-      await this.tor.init()
-    } else {
-      throw new Error('You must provide either tor control port or tor binary path')
```
siepra, line 305: aren't we loosing this throw by separating the method into `initTorDesktop` and `initTorMobile`?

packages/e2e-tests/newUser.crossplatform.test.ts
```diff
@@ -141,7 +142,7 @@ describe('New User', () => {
       await buildSetup2.createChromeDriver()
       driver2 = buildSetup2.getDriver()
       await driver2.getSession()
-      await new Promise<void>(resolve => setTimeout(() => resolve(), 10000))
+      // await new Promise<void>(resolve => setTimeout(() => resolve(), 30000))
```
siepra, line 145: Why is that commented out?

siepra (commented):
My overall feeling is that spinning up tor got a little bit confusing. I wasn't able to fully get what's going on by just reading the diff

EmiM (approved):

packages/backend/src/libp2p/connectionsManager.ts
```diff
@@ -308,18 +308,12 @@ export class ConnectionsManager extends EventEmitter {
...
-    if (this.torBinaryPath) {
-      await this.tor.init()
-    }
-  }
-
-  public initTorMobile = async () => {
     if (this.torControlPort) {
       this.tor.initTorControl()
+    } else if (this.torBinaryPath) {
+        //
```
siepra, line 314: Does anything happen here?

packages/backend/src/storage/storage.ts
```diff
@@ -98,6 +100,7 @@ export class Storage extends EventEmitter {
       // @ts-ignore
       AccessControllers: AccessControllers
     })
+    this.emit(SocketActionTypes.TOR_CONNECTION_PROCESS, TorConnectionProcessInfo.INITIALIZED_STORAGE)
```
vinkabuki, line 103: I don't think it's right that we have storage specific events under TorConnectionProcessInfo => I can imagine possible scenarios that it misleads someone in the future.

siepra (approved):

vinkabuki (approved):

PR #1353, opened 2023-03-22
Changes to mobile ChannelList date and time formatting

### LINKED ISSUE #1022: Dates should match design
In the channel list on mobile, "most recent message sent on..." dates display as e.g. "Nov 10, 12:23PM" when they should display as either "Yesterday" or "Nov 10"

See design: https://www.figma.com/file/y8h6w8PYR9jyI3zjYHL9Cl/Mobile-(and-prototype)?node-id=456%3A18822&t=FRJwrcYEsdOlvFPP-0

image

josephlacey:
Fixes #1022


siepra (commented):
Looks great but I really miss unit tests for that method. I know there is no corresponding test file but that's only because the app was a POC at a time and now as we do it right, we probably wanna include proper tests

josephlacey:
> Looks great but I really miss unit tests for that method. I know there is no corresponding test file but that's only because the app was a POC at a time and now as we do it right, we probably wanna include proper tests

Sure, I can write a unit for this.

josephlacey:
> > Looks great but I really miss unit tests for that method. I know there is no corresponding test file but that's only because the app was a POC at a time and now as we do it right, we probably wanna include proper tests
> 
> Sure, I can write a unit for this.

@siepra I've added a test for this utility component.  My initial approach to the test relied on absolute timestamps, but that required refactoring the source component to include an argument for a 'now' time to compare the 'createAt' time against.  But that didn't seem like the correct approach, so I reverted that and rewrote the test to use relative datetimes.

siepra:
Hmm... I think it's fine, but usually the right way is to mock the Date object with jest

Issue #530, opened 2022-05-26
RTL test for channelInput doesn't work


vinkabuki:
For some reason RTL test for channelInput stopped working. @siepra had this problem before, I somehow managed to create working test but it doesn't work again and it's not clear why

kingalg:
@vinkabuki said that it's something that didn't work a year ago but now it can be closed.

PR #1377, opened 2023-03-28
Traverse previous and next lines with up and down arrow keys

### LINKED ISSUE #1339: Arrow up and down doesn't work correctly in message field
Version: 1.1.0-alpha.5
Systems: all

User should be able navigate multi-line message that he is writing with all four arrows. Currently only arrow left and right are working.

josephlacey:
Fixes #1339 

This is a first draft of a fix.  I feel like it needs feedback before going any further with it. It's not very React-minded and instead just adds a vanilla-JavaScript helper function for the existing keydown callback. I couldn't figure out a more React-minded approach that didn't seem to include more overhead than was necessary, but I might be missing something simple here.  So before I add a test for this, I wanted to make sure this seemed like the right approach.

Also I added some style to the input div to respect pasted-in white space.  This was something that came up during my manual testing, but there might be a reason not to do this.


siepra (commented):
I think it's fine. I know this issue is tricky so the solution you came up with doesn't seem hacky to me

josephlacey:
> I think it's fine. I know this issue is tricky so the solution you came up with doesn't seem hacky to me

Thanks @siepra 

And an update.  I have a working test, but the problem is that the ContentEditable Div doesn't seem to have the same setup in the testing environment as it does in the App, so the traversal code isn't really being tested.  Not sure why the difference during the test, but hopefully I'll have this sorted out by tomorrow.

josephlacey:
> > I think it's fine. I know this issue is tricky so the solution you came up with doesn't seem hacky to me
> 
> Thanks @siepra
> 
> And an update. I have a working test, but the problem is that the ContentEditable Div doesn't seem to have the same setup in the testing environment as it does in the App, so the traversal code isn't really being tested. Not sure why the difference during the test, but hopefully I'll have this sorted out by tomorrow.

@siepra I've included the test I've been banging against for two days, but it still doesn't successfully complete.  I've marked it to be skipped pending further work.  It looks related to this existing issue,  #530.  I dug into the problem a bit and found that it's supposed to work, [according to the feature detailed on this testing library issue](https://github.com/testing-library/user-event/issues/442), but I couldn't get the innerHTML and textContent values to change in the input field.  Using fireEvent and targeting either of those property would set the value of course, but that doesn't simulate the actual typing required for the multi-line test nor does it setup the DOM Element for proper traversal.

I could keep working at this, and by extension [the test mentioned](https://github.com/TryQuiet/quiet/blob/develop/packages/desktop/src/rtl-tests/channel.test.tsx#L510) in #530, but one could also make the argument that both of these are such baseline functionalities that they get tested in the process of development.  Evidenced perhaps by the other issue being backlogged for almost a year now.  What do you think?

siepra:
I think it's fine to skip the tests for now if they cause problems . As you said, we're paying a lot of attention to this component during development and QA sessions anyway. If somehow in the future we'll start running into problems with this thing, we'll try to come up with some automatic tests ideas

packages/desktop/src/rtl-tests/channel.test.tsx
```diff
@@ -563,6 +563,85 @@ describe('Channel', () => {
     `)
   })
 
+  it.skip('user can enter a multi-line message and then traverse it with arrow keys', async () => {
+    const { store, runSaga } = await prepareStore(
+      {},
+      socket // Fork state manager's sagas
```
siepra, line 569: I wonder if we really need to fork sagas here... it'd probably be better to isolate input component for this test. On the other hand it may be useful to check whether keyboard mapping works in wide context - can we achieve it without having to populate the store?

josephlacey:
I've pushed a refactor of the tests for this PR.  The first is simply that a multi-line message is correctly handled.  That's enabled and passes.

The second test is the traversal of the multi-line message and is currently skipped.  The test, as written, mirrors the input field with a multi-line message, but the test fails because the userEvent typing doesn't render the message parts like the app does.  I could rewrite this to pass what the test renders, but that wouldn't test the code, so doesn't seem useful.

packages/desktop/src/rtl-tests/channel.test.tsx
josephlacey: @siepra I've removed this fork and only render the component in the test, but unfortunately it still fails.

PR #1388, opened 2023-03-31
User should be able to share invite link

Kacper-RF:


vinkabuki (approved):

packages/desktop/src/renderer/components/Settings/Tabs/Invite/Invite.component.tsx
```diff
@@ -0,0 +1,88 @@
...
+      <Grid item>
+        <Grid item>
+          <Typography variant='h5'>Your community link</Typography>
+        </Grid>
+        <Grid item>
+          <Typography variant='body2'>
+            Anyone with Quiet app can follow this link to join this community.
+            <br /> Only share with people you trust.
+          </Typography>
+          <a onClick={() => openUrl(invitationLink)}>
```
siepra, line 72: This seems weird... why would someone want to click-open an invitation code to it's own community?

siepra (commented):
I think hiding/revealing the link would be usefull as it contains invitation code. We should probably keep this feature

siepra (approved):

PR #1398, opened 2023-04-04
Feature/invitation link error modals

EmiM:


siepra (approved):
Looks good but I can think of one more use case. What if user tries to use the link to the community he's already a member of? He should probably just be redirected to the app without any error (that's how we do on mobile)

PR #1409, opened 2023-04-05
[FIX]Tor errors on app init

Kacper-RF:


packages/desktop/src/renderer/components/CreateJoinCommunity/JoinCommunity/JoinCommunity.tsx
```diff
@@ -16,14 +22,19 @@ const JoinCommunity = () => {
...
+  const torBootstrapProcessSelector = useSelector(connection.selectors.torBootstrapProcess)
   const [revealInputValue, setRevealInputValue] = useState<boolean>(false)
 
   useEffect(() => {
-    if (isConnected && !currentCommunity && !joinCommunityModal.open) {
+    if (
+      isConnected &&
+      !currentCommunity &&
+      !joinCommunityModal.open &&
+      torBootstrapProcessSelector === 'Bootstrapped 100% (done)'
```
siepra, line 33: Risky. Can you store it in some shared const?

packages/desktop/src/renderer/store/reducers.ts
```diff
@@ -57,7 +76,19 @@ const allReducers = combineReducers(reducers)
 
 export const rootReducer = (state, action) => {
   if (action.type === communities.actions.resetApp.type) {
-    state = undefined
+    const torBootstrapProcess = state.Connection.torBootstrapProcess
```
siepra, line 79: I think it can possibly be misleading for someone in the future. I mean, this code doesn't look like it serves for resetting state anymore - for me it has something to do with monitoring tor connection. Excluding it into a separate function or at least adding some comment could be helpful maybe?

packages/state-manager/src/sagas/identity/identity.transform.ts
```diff
@@ -0,0 +1,16 @@
...
+import { IdentityState } from './identity.slice'
+
+export const IdentityTransform = createTransform(
+  (inboundState: IdentityState, _key) => {
+    return { ...inboundState }
+  },
+  (outboundState: IdentityState, _key) => {
+    return {
+      ...outboundState,
+      isRegisterButtonClicked: false
```
siepra, line 12: I really don't like it that we keep in store the information about component's state

vinkabuki (approved):

holmesworcester:
Does this include a regression test? I saw some changes to tests but didn't see a new test.

vinkabuki (approved):

PR #1415, opened 2023-04-06
Fix/tor flow backend state manager

vinkabuki:


packages/backend/src/libp2p/connectionsManager.ts
```diff
@@ -528,6 +531,7 @@ export class ConnectionsManager extends EventEmitter {
       await this.createCommunity(args)
     })
     this.dataServer.on(SocketActionTypes.LAUNCH_COMMUNITY, async (args: InitCommunityPayload) => {
+      if ([ServiceState.LAUNCHING, ServiceState.LAUNCHED].includes(this.communityState)) return
```
EmiM, line 534: Why are you checking it here if the same condition is being checked in `launchCommunity`?

EmiM (approved):

PR #1418, opened 2023-04-06
Mobile/Tor and state fixes

Kacper-RF:


siepra (approved):
Looks good. The only thing I would change is to rename the IsTorInit enum to TorInitState (it's just a matter of naming convention consistency)

PR #1444, opened 2023-04-12
Channel deletion

siepra:


packages/backend/src/storage/ChannelsAccessController.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const createChannelAccessController = (peerId) => {
+  // @ts-ignore
+  class ChannelsAccessController extends AccessController {
+    static get type() {
+      return type
+    }
+
+    async canAppend(entry, identityProvider) {
+      // Channel deletion WIP
```
siepra, line 17: Is this still WIP?

packages/desktop/libssl.so
siepra, line 1: Are we sure we don't need it anymore? What about Ubuntu 22.04?

packages/desktop/package.json
```diff
@@ -80,7 +80,7 @@
   },
   "homepage": "https://github.com/TryQuiet",
   "@comment version": "To build new version for specific platform, just replace platform in version tag to one of following linux, mac, windows",
-  "version": "1.2.0",
+  "version": "1.2.0-alpha.31",
```
siepra, line 83: Isn't this version behind the full 1.2.0?

packages/state-manager/src/sagas/messages/messages.slice.ts
```diff
@@ -38,6 +39,10 @@ export const messagesSlice = createSlice({
   name: StoreKeys.Messages,
   reducers: {
     sendMessage: (state, _action: PayloadAction<WriteMessagePayload>) => state,
+    deleteMessages: (state, action: PayloadAction<DeleteMessagesPayload>) => {
```
siepra, line 42: I think this name may be confusing. It sounds like it deletes certain amount of messages, How about marking that we delete the whole channel entry for messages?

packages/state-manager/src/sagas/publicChannels/deletedChannel/deletedChannel.saga.test.ts
```diff
@@ -0,0 +1,167 @@
...
+            description: 'Welcome to #photo',
+            timestamp: DateTime.utc().valueOf(),
+            owner: owner.nickname,
+            address: 'photo'
+          }
+        }
+      )
+    ).channel
+  })
+  describe('owner handling', () => {
```
siepra, line 60: I like test names to be more descriptive. If I saw 'owner handling' test failing, I wouldn't know what exactly doesn't work

packages/state-manager/src/sagas/publicChannels/publicChannels.slice.ts
```diff
@@ -29,24 +33,44 @@ const log = logger('publicChannels')
 
 export class PublicChannelsState {
   public currentChannelAddress: string = 'general'
+  public isGeneralRecreation: boolean = false
```
siepra, line 36: Once again it's just a minor naming thing, but 'pendingGeneralChannelRecreation' would be more accurate, in my opinion

packages/desktop/libssl.so
vinkabuki: It was added here by accident

packages/desktop/package.json
vinkabuki: right

packages/state-manager/src/sagas/publicChannels/publicChannels.slice.ts
vinkabuki: +1

packages/state-manager/src/sagas/publicChannels/deletedChannel/deletedChannel.saga.ts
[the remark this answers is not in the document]
Kacper-RF: I will do that in next PR

vinkabuki (approved):

vinkabuki (approved):

packages/backend/src/storage/ChannelsAccessController.ts
holmesworcester: We merged this just so that emi's typescript work could move forward?

PR #1450, opened 2023-04-13
Fix/Error on startup - TOR: Connection already established

Kacper-RF:
- [x] Checking statuses in connectionsManager
- [x] Unit test


packages/backend/src/libp2p/connectionsManager.ts
```diff
@@ -255,12 +254,15 @@ export class ConnectionsManager extends EventEmitter {
         community.peers = sortedPeers
       }
       await this.localStorage.put(LocalDBKeys.COMMUNITY, community)
+      if ([ServiceState.LAUNCHING, ServiceState.LAUNCHED].includes(this.communityState)) return
+      this.communityState = ServiceState.LAUNCHING
       await this.launchCommunity(community)
```
vinkabuki, line 259: This should be moved to the bottom, first we should set status to launching for both community and registrar, then launch them

vinkabuki (approved):

PR #1483, opened 2023-04-25
Remove webflow and related functionality

josephlacey:
Fixes https://github.com/TryQuiet/zbay-website/issues/9


siepra:
This `navigation.js` file is the one I came up with and it doesn't cover smooth transitions. If you @holmesworcester think it's enough, let's just merge it

PR #1522, opened 2023-05-12
Add release build configuration to detoxrc

siepra:
There's an additional action required after merging the branch. Bitrise AVD setup should point to `emulator` device name

### DIFF
--- a/packages/mobile/android/app/src/main/java/com/quietmobile/MainActivity.java
+++ b/packages/mobile/android/app/src/main/java/com/quietmobile/MainActivity.java
@@ -104,13 +104,10 @@ public class MainActivity extends ReactActivity {
         Intent intent = getIntent();
         checkAgainstIntentUpdate(intent);
 
-        Context context = getApplicationContext();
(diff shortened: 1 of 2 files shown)


packages/mobile/android/app/src/main/java/com/quietmobile/MainActivity.java
```diff
@@ -104,10 +104,13 @@ protected void onCreate(Bundle savedInstanceState) {
...
         checkAgainstIntentUpdate(intent);
 
-        if (BuildConfig.SHOULD_RUN_BACKEND_WORKER == "true") {
-            Context context = getApplicationContext();
-            new BackendWorkManager(context).enqueueRequests();
-        }
+        Context context = getApplicationContext();
+        new BackendWorkManager(context).enqueueRequests();
+
+//        if (BuildConfig.SHOULD_RUN_BACKEND_WORKER == "true") {
```
siepra, line 110: The problem with this conditional statement is that detox has problems reading the environment variable. I'd want to see how it works with storybook before removing it entirely.

PR #1535, opened 2023-05-26
bug/post-prerelease-fixes

Kacper-RF:
- [x] Only owner should delete channel from OrbitDB
- [x] Fix files deletion
- [x] Don't run deletion loop in replication saga while channels object from backend is empty

### DIFF
--- a/packages/backend/src/storage/storage.test.ts
+++ b/packages/backend/src/storage/storage.test.ts
@@ -623,16 +623,16 @@ describe('Files deletion', () => {
       storage.deleteFilesFromChannel(messages)
     ).resolves.not.toThrowError()
 
-    await sleep(1000)
-    const isFileExist2 = await storage.checkIfFileExist(realFilePath)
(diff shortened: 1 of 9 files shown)


packages/backend/src/storage/storage.test.ts
```diff
@@ -541,3 +565,77 @@ describe('Users', () => {
...
+   const isFileExist = await storage.checkIfFileExist(realFilePath)
+    expect(isFileExist).toBeTruthy()
+
+    await expect(
+      storage.deleteFilesFromChannel(messages)
+    ).resolves.not.toThrowError()
+
+    await sleep(1000)
+    const isFileExist2 = await storage.checkIfFileExist(realFilePath)
+    expect(isFileExist2).toBeFalsy()
```
EmiM, line 628: Maybe you could use waitForExpect here instead of sleep?

EmiM (approved):

vinkabuki (approved):

PR #1548, opened 2023-06-01
fix selector for mobile channel statuses

vinkabuki:


packages/state-manager/src/sagas/publicChannels/publicChannels.selectors.test.ts
```diff
@@ -391,6 +391,50 @@ describe('publicChannelsSelectors', () => {
     const unreadChannels = publicChannelsSelectors.unreadChannels(store.getState())
     expect(unreadChannels).toEqual([channelId])
   })
+
+  it.only('channelsStatusWithName returns valid data', async () => {
```
EmiM, line 395: `.only` - is it a leftover?

PR #1552, opened 2023-06-02
bug/deleting channels on mobile bugged

Kacper-RF:


packages/state-manager/src/sagas/publicChannels/deleteChannel/deleteChannel.saga.ts
```diff
@@ -24,12 +24,10 @@ export function* deleteChannelSaga(
...
-  if (currentChannel === undefined) {
-    return
-  }
-
-  if (currentChannel.disabled) {
-    return
+  if (currentChannel) {
+    if (currentChannel.disabled) {
+      return
+    }
```
vinkabuki, line 30: you can replace this nested if with nice oneliner if (currentChannel?.disabled) return

packages/mobile/src/screens/ChannelList/ChannelList.screen.tsx
```diff
@@ -33,19 +34,17 @@ export const ChannelListScreen: FC = () => {
   )
 
   const community = useSelector(communities.selectors.currentCommunity)
+  const channelsStatusSorted = useSelector(publicChannels.selectors.channelsStatusSorted)
 
-  const channelsStatusWithName = useSelector(publicChannels.selectors.channelsStatusWithName)
-
```
vinkabuki, line 38: I think we can get rid of that selector and corresponding tests etc

vinkabuki (approved):

packages/mobile/src/screens/ChannelList/ChannelList.screen.tsx
vinkabuki: I think it was used only in that one place

PR #1586, opened 2023-06-22
backend/nest

Kacper-RF:


packages/backend/src/nest/app.module.ts
```diff
@@ -0,0 +1,132 @@
...
+    LocalDbModule,
+    Libp2pModule,
+    TorModule,
+    StorageModule,
+    IpfsModule,
+  ],
+  providers: [
+    {
+      provide: EXPRESS_PROVIDER,
+      useValue: express(),
```
kowalski, line 51: Tehnically better is
```
useFactory: () => express()
```

Does the same, but defers creation of the instance

packages/backend/src/nest/common/logger.ts
```diff
@@ -0,0 +1,9 @@
+import { logger } from '@quiet/logger'
+
+const createLogger = (name: string) => {
+  return {
+    log: logger('backend')(name),
```
kowalski, line 5: maybe just 
```
return logger('backend')(name)
```
?

I've seen in many places things linke `log.log.error(...)` as a result of this

packages/backend/src/nest/connections-manager/connections-manager.service.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+    await connectionsManagerService.closeAllServices()
+    await connectionsManagerService.init()
+    const launchCommunitySpy = jest.spyOn(connectionsManagerService, 'launchCommunity')
+    const launchRegistrarSpy = jest.spyOn(registrationService, 'launchRegistrar')
+    expect(launchCommunitySpy).not.toHaveBeenCalled()
+    expect(launchRegistrarSpy).not.toHaveBeenCalled()
+  })
+
+  // At this moment, that test have to be skipped, because checking statues is called before launchCommunity method
+  // it.skip('community is only launched once', async () => {
```
kowalski, line 187: so it enough to just `it.skip`, don't comment out

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -0,0 +1,611 @@
...
+  ) {
+    super()
+  }
+
+  async onModuleInit() {
+    process.on('unhandledRejection', error => {
+      console.error(error)
+      throw new Error()
+    })
+    process.on('SIGINT', function () {
```
kowalski, line 93: I'm almost positive it's not necessary now that we have nest

packages/backend/src/nest/validation/validators.ts
```diff
@@ -1,8 +1,8 @@
 import _ from 'validator'
 import joi from 'joi'
-import logger from '../logger'
-import { type ChannelMessage, type PublicChannel } from '@quiet/types'
-const log = logger('validators')
+// import logger from '../logger'
```
kowalski, line 3: remove

packages/backend/src/nest/websocketOverTor/utils.ts
```diff
@@ -16,7 +15,7 @@ export function dumpPEM(tag: string, body: string | Certificate | CryptoKey) {
   return Buffer.from(result)
 }
 
-export async function initConnectionsManagerWithTor(connectionsManager: ConnectionsManager, port: number) {
+export async function initConnectionsManagerWithTor(connectionsManager: any, port: number) {
```
kowalski, line 18: actually I prefered without `any` :)

kowalski (approved):

PR #1653, opened 2023-07-19
bug/new/unread message indicator sometimes missing

Kacper-RF:


packages/state-manager/src/sagas/publicChannels/markUnreadChannels/markUnreadChannels.saga.ts
```diff
@@ -23,7 +25,15 @@ export function* markUnreadChannelsSaga(
 
       const statuses = yield* select(publicChannelsSelectors.channelsStatus)
 
-      const joinTimestamp = yield* select(identitySelectors.joinTimestamp)
+      // Fix for users whose has damaged property with join timestamp and problem with proper checking new message
+      let joinTimestamp: number | null | undefined
```
EmiM, line 29: I think it would be a better idea for this to be added as transform instead of in app's logic

siepra: I agree it shouldn't be done in saga, but I'm not certainly convinced if it should be a responsibility of transformer either... I'd rather changed selector to never return `undefined`

siepra (approved):

PR #1659, opened 2023-07-20
tor async happy path

vinkabuki:


EmiM (approved):

packages/state-manager/src/sagas/socket/startConnection/startConnection.saga.ts
```diff
@@ -87,13 +87,15 @@ export function subscribe(socket: Socket) {
     | ReturnType<typeof connectionActions.setTorConnectionProcess>
     | ReturnType<typeof connectionActions.torBootstrapped>
     | ReturnType<typeof connectionActions.connectionManagerInit>
+    | ReturnType<typeof connectionActions.setTorInitialized>
   >(emit => {
     // UPDATE FOR APP
     socket.on(SocketActionTypes.TOR_BOOTSTRAP_PROCESS, (payload: string) => {
```
EmiM, line 93: I would change this event name since it does not inform about process anymore, it's 1/0 information now.

EmiM (approved):

PR #1668, opened 2023-07-26
Chore/immediate UI launch

siepra:


packages/mobile/src/store/nativeServices/events/nativeServicesCallbacks.ts
```diff
@@ -70,7 +70,7 @@ export const deviceEvents = () => {
         emit(network.actions.removeInitializedRegistrars())
       }),
       nativeEventEmitter?.addListener(NativeEventKeys.AppResume, () => {
-        emit(navigationActions.navigation({ screen: ScreenNames.SplashScreen }))
+        // emit(navigationActions.navigation({ screen: ScreenNames.SplashScreen }))
```
Kacper-RF, line 73: Should it stay or be removed?

packages/mobile/src/components/CreateCommunity/CreateCommunity.component.tsx
```diff
@@ -12,6 +12,7 @@ export const CreateCommunity: FC<CreateCommunityProps> = ({
   createCommunityAction,
   redirectionAction,
   networkCreated,
+  ready = true,
```
Kacper-RF, line 15: Why provide that props with default value ?

packages/mobile/src/store/nativeServices/events/nativeServicesCallbacks.ts
siepra: It stays

packages/state-manager/src/sagas/appConnection/connection.slice.ts
```diff
@@ -8,6 +8,7 @@ export class ConnectionState {
   public uptime = 0
   public peersStats: EntityState<NetworkStats> = peersStatsAdapter.getInitialState()
   public isConnectionManager = false
+  public isTorInitialized = false
   public torBootstrapProcess = 'Bootstrapped 0% (starting)'
```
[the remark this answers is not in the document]
siepra: It's a leftover from cherry-picking commits

packages/mobile/src/components/CreateCommunity/CreateCommunity.component.tsx
siepra: Because we don't want to risk blocking a component without specified reason eg. in case selector failed

PR #1672, opened 2023-07-27
Chore/decrease ios startup time

siepra:


siepra:
There are a few things here that still needs our attention:
1. **Leaving community stopped working**
2. **Detox test for minimizing app works but marks failure**
3. "Join/Create community" and "Username registration" screens are greying out input instead of being temporarily covered by splash screen when coming back from idle state
4. Greyed out inputs are blinking when coming back from idle state

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -113,6 +117,57 @@ export class StorageService extends EventEmitter {
 
     this.attachFileManagerEvents()
     await this.initDatabases()
+
+    this.startIpfs()
+  }
+
+  private startIpfs() {
```
kowalski, line 124: Async

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -113,6 +117,57 @@ export class StorageService extends EventEmitter {
...
+    this.ipfs
```
kowalski, line 125: Return

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -113,6 +117,57 @@ export class StorageService extends EventEmitter {
 
     this.attachFileManagerEvents()
     await this.initDatabases()
+
+    this.startIpfs()
```
kowalski, line 121: Void

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -137,12 +193,12 @@ export class StorageService extends EventEmitter {
     await this.createDbForChannels()
     this.logger('2/6')
     await this.createDbForCertificates()
-    this.logger('3/6')
-    await this.createDbForMessageThreads()
+    // this.logger('3/6')
```
kowalski, line 196: If this is no longer needed, it should be removed. If you plan to bring it back in future please add a line of comment explaining why this is commented out

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -137,12 +193,12 @@ export class StorageService extends EventEmitter {
...
+    // await this.createDbForMessageThreads()
     this.logger('4/6')
     await this.initAllChannels()
-    this.logger('5/6')
-    await this.initAllConversations()
+    // this.logger('5/6')
```
kowalski, line 200: Same as above

kowalski (approved):

EmiM (approved):

Issue #1423, opened 2023-04-07
Small typo: "never mind"


holmesworcester:
Steps to reproduce

1. join community on mobile
2. leave community

typo in "are you sure?" prompt. "Never mind' should be "Nevermind"

leblowl:
Actually, I'm noticing this issue mentions mobile. It looks like this is only fixed for desktop currently

PR #1675, opened 2023-07-28
Fix typo "Nevermind -> "Never mind"

hongha912:
Fix this first issue https://github.com/TryQuiet/quiet/issues/1423

The issue mentioned "Never mind" should be "nevermind", but [Grammarly](https://www.grammarly.com/blog/nevermind-or-never-mind/#:~:text=Never%20mind%20tells%20someone%20to,pay%20something%5D%20no%20nevermind.%E2%80%9D) attests that the other way round is more correct, so I fixed the only instance of the other way instead. 

image


holmesworcester:
@hongha912 Sorry for the not-yet-documented hiccup! The visual regression tests are failing because we use an external service for running react storybook tests, and there's an additional step or two to get them to run. I'm not sure exactly what's needed and we should document this better--it came up with @josephlacey too. But here's what I see. 

First I think you need to sign in to Chromatic with Github:
image

Then I see this: 
image
https://www.chromatic.com/review?appId=62021a6b9c4ef3003ab65846&number=1675&type=linked

...Which I think means you need to follow these instructions for it to run the tests on this PR. 

@josephlacey is this right? 

We should add something to the [desktop README](https://github.com/TryQuiet/quiet/blob/develop/packages/desktop/README.md) about this once we know the steps to get it working.

josephlacey:
@holmesworcester @hongha912 

> ...Which I think means you need to follow these instructions for it to run the tests on this PR.
> 
> @josephlacey is this right?

Yes, this looks correct.  The missing part is the Chromatic project token.  Outside contributors can't access the GitHub Secret to run an automated version of that test.  I have it stored as a local environment variable, `$CHROMATIC_PROJECT_TOKEN`, so I just run `npm run chromatic` with the branch checked out that's being tested, usually just a second step after pushing commits to the PR.   That script handles all the details, and it's pretty seamless.  The longer command in the example from Chromatic seems to provide more granular control, which probably works fine too.

Github will then register that a test related to the PR has been sent manually and include it in the testing report.  The Visual Regression failure never goes away, but that's not a big deal.
 
> We should add something to the [desktop README](https://github.com/TryQuiet/quiet/blob/develop/packages/desktop/README.md) about this once we know the steps to get it working.

Yeah, we held this because we hadn't settled on a policy of how to handle the token.  Not sure if including that token in a public issue is a security issue or not.  The testing itself seems harmless, but maybe someone could send more tests than the plan includes and exceed the limit of tests for the month?  Regardless if you want to make an issue and assign it me, I can add the relevant parts to the README.

holmesworcester:
Let's write it up and invite people to request the token in their first PR.

holmesworcester:
@hongha912 I just shared the Chromatic project token with you on Signal. 

Thanks for the clarification, @josephlacey!

hongha912:
Thanks both! So, I've gone through these instructions. I was able to invoke a build locally by setting the CHROMATIC_PROJECT_TOKEN here https://www.chromatic.com/build?appId=62021a6b9c4ef3003ab65846&number=1849

Seems like I'll need to set the token as a secret in my forked repo. https://github.com/TryQuiet/quiet/actions/runs/5791705156/job/15696918936?pr=1675 also points to some failures that might be secret-related? Are there any other credentials I have to set?

Also, is forking the usual way you let contributors contribute? Maybe I was supposed to push directly to your repo? (I tried that at first and was rejected).

Kacper-RF:
Hi @hongha912, thanks for contributing to Quiet !

I restarted e2e workflows, and i will check if mac will pass tests. For Windows we need to figure out how to run workflow without this token.

Kacper-RF:
Please update your branch with newest `develop` branch, this could help with failing E2E tests for MacOS, because we released new alpha version and value in your package.json is lower `1.7.1-alpha.0` then on current develop branch  `1.7.1-alpha.1` , so autoupdater try to update it on workflow.

For Windows I'm currently working on solution that for E2E workflow no secret will be needed, because we only build that version for testing purposes.

Kacper-RF:
I merged to develop my PR with building Windows installer on E2E workflow without any secrets so it should help with failing Windows tests.

hongha912:
Apologies for keep coming back to this. I've added the Chromatic as a repository secret in my forked repo (CHROMATIC_PROJECT_TOKEN), but the Chromatic job is still failing with the same error.

https://github.com/TryQuiet/quiet/actions/runs/5834060524/workflow?pr=1675

Seeing the workflow file above, I thought I was doing the correct thing, but is the setup different?

Kacper-RF:
We saw changes on Chormatic and they are already approved.

We probably need to investigate if there's any possibility that `Visual regressions` workflow will not fail when some external collaborator wants to create a PR.

Please update Your branch with the newest develop(mac e2e tests failing), and pull request will be approved.

holmesworcester:
@Kacper-RF can you create an issue for whatever investigation you'd like to do? Will a future contributor run into this same  problem?

leblowl:
I think we can probably just merge this even with the visual regression tests not working, since it's such a small change.

hongha912:
@leblowl oh yeah totally! I don't think I have write access so someone in Quiet would need to do it!

PR #1726, opened 2023-08-22
test: add test suite for channelAddress

UlisesGascon:


packages/common/src/channelAddress.test.ts
```diff
@@ -0,0 +1,29 @@
...
+  })
+})
+
+describe('Get Channel Name From Channel Id', () => {
+  it('Should return the channel name', () => {
+    const channelId = 'rockets_1faff74afc8daff3256275ce89d30528'
+    const channelName = 'rockets'
+    expect(getChannelNameFormChannelId(channelId)).toEqual(channelName)
+  })
+  it('Should return the channel id fi does not match the structure', () => {
```
EmiM, line 23: Thank you for adding tests :)
 
There is a typo "fi" -> "if". Also I believe that the function should be named `getChannelNameFromChannelId` ("Form" -> "From"). I'd appreciate if you fixed that too.

UlisesGascon: Yeah, there is a typo in the function name. I fixed it in https://github.com/TryQuiet/quiet/pull/1727. Also, I fixed the typo that I introduced with the `fi`. If you want, we can close this PR and focus on the other one that includes the commits from this PR too.

PR #1731, opened 2023-08-22
add missing edge test cases for invitation codes

UlisesGascon:


packages/common/src/invitationCode.test.ts
```diff
@@ -18,7 +18,15 @@ describe('Invitation code helper', () => {
     expect(invitationDeepUrl('validCode')).toEqual('quiet://?code=validCode')
   })
 
+  it('builds proper invitation deep url with empty code', () => {
+    expect(invitationDeepUrl()).toEqual('quiet://?code=')
```
EmiM, line 22: Invitation URL (both share url and deep url) is being changed in another PR

UlisesGascon: Oh! I was not aware, I will close this PR then 👍

Issue #1732, opened 2023-08-22
Implementing OpenSSF Scorecard


UlisesGascon:
Hi team! I think it would be beneficial for the project if we follow the recommendations from the [OpenSSF Scorecard](https://github.com/ossf/scorecard).

### Context

Many projects are part of this initiative, such as [Node.js](https://kooltheba.github.io/openssf-scorecard-api-visualizer/#/projects/github.com/nodejs/node), [ipfs](https://kooltheba.github.io/openssf-scorecard-api-visualizer/#/projects/github.com/ipfs/ipfs), and many more. Even if the project is not actively promoting the initiative, the Scorecard will index them once they reach a certain level of popularity.

If you are unfamiliar with the tool, here is some additional information:

- [What is Scorecard?](https://github.com/ossf/scorecard#what-is-scorecard)
- [You should use the OpenSSF Scorecard](https://blog.ulisesgascon.com/openssf-scorecard-in-nodejs)

Overall, the recommendations suggested in the scorecard are good for increasing security in Open Source projects (branch protection, pinning dependencies, tracking vulnerabilities, security policies...).

### Next Steps

I will create some PRs that will help us speed up the scoring process, and I will also include a proper workflow for submitting the project scoring.

holmesworcester:
This is great. Thank you!

I'm curious what steps you would recommend after this one, to protect against vulnerable dependencies and secure against supply chain attacks.

We're very aware that there's a lot more work to be done here, so I'm curious what order of steps you recommend.

UlisesGascon:
> I'm curious what steps you think we should take after this one, to protect against vulnerable dependencies and secure against supply chain attacks.
>
> We're very aware that there's a lot more work to be done here, so I'm curious what you recommend.

Thanks for the feedback, @holmesworcester! 

**TL:DR;**

While the Scorecard offers us many things and a way to [monitor the evolution](https://github.com/marketplace/actions/openssf-scorecard-monitor), for a long-term objective, we can try to achieve [the CII Best Practices](https://bestpractices.coreinfrastructure.org/en).

Basically, the idea is to sign up for the project and fill out the forms for the different levels (passing, silver, and gold). While this is a quite long process in terms of researching and checking that we have the best practices in place, it is a very systematic way to increase security in the community without feeling overwhelmed.

**CII Implementation details**

[...]

PR #1734, opened 2023-08-22
Enable dependabot

UlisesGascon:
### Main Changes

This PR will enable [Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file). While the suggested frequency is daily, I suggest weekly or even a custom CRON Job to control the noise (if you are not used to it) 🙂. It will automatically generate PRs with the dependency upgrades.

### Context

Related TryQuiet/quiet#1732

### Changelog

- 1b283ac feat: enable dependabot by @UlisesGascon

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -60,15 +60,17 @@ updates:
     schedule:
       interval: weekly
 
-  - package-ecosystem: docker
-    directory: /packages/integration-tests
-    schedule:
-      interval: weekly
+  # Currently, the integration tests pkg is waiting for a clean up.
+  # @SEE:  https://github.com/TryQuiet/quiet/pull/1734#discussion_r1302831794
+  # - package-ecosystem: docker
+  #   directory: /packages/integration-tests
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,106 @@
...
+    schedule:
+      interval: weekly
+
+  - package-ecosystem: docker
+    directory: /packages/integration-tests
+    schedule:
+      interval: weekly
+
+  - package-ecosystem: npm
+    directory: /packages/integration-tests
```
EmiM, line 69: I would not check integration-tests right now, we have to clean up the package first. I believe that it was not adjusted after we introduced nest to the project. I am afraid that it would be messy if dependabot checked it right now but maybe I am wrong

UlisesGascon: Resolved in c9387db

EmiM (approved):

PR #1735, opened 2023-08-23
Pin dependencies for GH Actions and Docker

UlisesGascon:
### Main Changes

Pinned dependencies for GitHub actions and Docker (4a739cc & 043d901)

**Important**

This PR has mapped the current version in usage for each dependency and pinned it in the `Dockerfile` and `GitHub Actions` workflow files. This makes the build more reproducible (immutable dependencies) and stable and also makes it easier to track changes in the dependencies.

While upgrading dependencies might seem complicated as it currently uses hashes, it is not. Because the dependencies can be upgraded automatically using dependabot #1734. Additionally, the dependencies include a human-readable version in the hash, so it is easy to know which version is being used:

```
steps:- uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # v2.7.0
```

```
FROM node:18.12.1@sha256:e9ad817b0d42b4d177a4bef8a0aff97c352468a008c3fdb2b4a82533425480df
```

### Context

Related TryQuiet/quiet#1732

### Changelog

- 4a739cc chore: pin dependencies for GitHub actions by @UlisesGascon
- 043d901 chore: pin Docker dependencies by @UlisesGascon


holmesworcester:
@UlisesGascon --  @EmiM just sent a note in Quiet wondering if you could resolve conflicts in this PR. Just wanted to note that here in case you weren't getting Quiet notifications, which are still pretty noisy ironically :)

UlisesGascon:
Conflict solved in d07df2f 🎉

leblowl:
I fixed up the conflicts over here: #2340

Issue #1649, opened 2023-07-19
User should be able to tap/click on "unregistered"  label to learn what it means


vinkabuki:
When we let users join without the community owner being online, we label users who are not registered  with an "unregistered" label so that it's clear their names are not final yet. 

Users should be able to click or tap on this label to see an explanation of what it means. Designs:

https://www.figma.com/file/TV9pF84Ob8pLYRLu83gNol/Joining-when-owner-is-offline?type=design&node-id=311-5980&mode=design&t=zJWpEmDpd1LSWwEO-4

![Image](https://github.com/TryQuiet/quiet/assets/213678/f089b23b-b64c-491b-af2a-ef089a7b7691)

kingalg:
@holmesworcester @vinkabuki - could you please provide some designs and more detailed description?

holmesworcester:
I didn't create this issue so I'm actually not sure what is meant here. I think this was created when we split the epic and it's based on the designs, but because there's no description I don't know what it means :)

We need to add a few lines of issue description even if it seems obvious!

Moving to 'blocked'

vinkabuki:
It was just copy-pasted from #1340, so I also didn't know details.

kingalg:
Version: 2.0.0-alpha.12
System: Windows, MacOS, Linux

User is able to click on "unregistered" to get details. 
Inside popup "OK" and X (close) buttons are working.

<img width="798" alt="Screenshot 2023-09-21 at 11 51 00" src="https://github.com/TryQuiet/quiet/assets/68909168/142a0735-ac7d-4452-97b9-e843521ff501">

Issue can be closed after **checking this on mobiles** that are not yet ready.

kingalg:
mobile@2.0.1-alpha.7 (iOS 325)
System: android and ios

Button works correctly on mobile devices. I'm not sure about matching design. Is the one on figma intended to cover half of the screen and our implementation is wrong or what we did is correct? @holmesworcester 
Our implementation:
<img width="328" alt="Screenshot 2023-10-17 at 12 47 15" src="https://github.com/TryQuiet/quiet/assets/68909168/9a2cadc6-76f6-4230-bfd8-0db42c51e31d">
Figma design:
<img width="258" alt="Screenshot 2023-10-17 at 12 48 18" src="https://github.com/TryQuiet/quiet/assets/68909168/eaacc2bb-43b6-4985-8df9-9664d547c73d">

holmesworcester:
[...]

Issue #1559, opened 2023-06-02
Warn user of duplicate unregistered names


holmesworcester:
As part of #1340 we are letting unregistered users send messages. We should use these designs to warn users if an unregistered user has the same name as a registered user **or** as another unregistered user: 

In the message screen, all duplicate usernames should have a "duplicate" warning:

https://www.figma.com/file/TV9pF84Ob8pLYRLu83gNol/Joining-when-owner-is-offline?type=design&node-id=311-7523&mode=design&t=eS9WDLYpQNod1Acl-4

![Image](https://github.com/TryQuiet/quiet/assets/213678/d92202e5-8b11-43ad-a1a4-caeebf320e17)

Tapping on this warning should take users to an explanation and warning:

https://www.figma.com/file/TV9pF84Ob8pLYRLu83gNol/Joining-when-owner-is-offline?type=design&node-id=320-2662&mode=design&t=eS9WDLYpQNod1Acl-4

![Image](https://github.com/TryQuiet/quiet/assets/213678/ee795120-8229-4c96-a471-f4d6d5ef64ca)

kingalg:
@holmesworcester this issue is about unregistered user having the same username as registered user. Is it also possible that two unregistered users will have the same names? I can't find issue for this and I was thinking if this is technically possible or not?

holmesworcester:
It's definitely possible. Unregistered users having the same name can be handled with the same warning as registered/unregistered.

kingalg:
quiet@2.0.3-alpha.0

Done on Desktop and Mobile. All issues discovered in this area has been reported in separate github issues.

Issue #1558, opened 2023-06-02
Unregistered users should be obvious


holmesworcester:
Unregistered users should have the note "Unregistered" next to their name

https://www.figma.com/file/TV9pF84Ob8pLYRLu83gNol/Joining-when-owner-is-offline?type=design&node-id=571-2124&mode=design&t=C8hfFvMP9S6qUm4E-4

![Image](https://github.com/TryQuiet/quiet/assets/213678/999aab88-4bec-4ccb-a1b9-c275535c84f4)

Kacper-RF:
@holmesworcester I have one question - what if the user is unregistered and used a duplicate username, which label is more important in this case?

holmesworcester:
Duplicate.

kingalg:
Version: 2.0.0-alpha.12
System: Windows, MacOS, Linux

Unregistered users get "Unregistered" label and in case it's duplicated unregistered username they get "Duplicate".

There is a bug and system messages sometimes shows as "Unregistered" - it's part of the issue https://github.com/TryQuiet/quiet/issues/1842 to be handled in current sprint.

<img width="425" alt="Screenshot 2023-09-21 at 11 21 59" src="https://github.com/TryQuiet/quiet/assets/68909168/2f84c1af-e0b2-48ca-a38f-459aea077926">

User can click on "Unregistered" label and get more details.

<img width="581" alt="Screenshot 2023-09-21 at 11 26 09" src="https://github.com/TryQuiet/quiet/assets/68909168/f156f745-24c2-4409-8e02-7f7f228a5129">

Issue will be closed **after mobile versions are ready and check**

PR #1762, opened 2023-09-05
Feature/1558/1559/1649

Kacper-RF:
### Pull Request Checklist

Only desktop part at this moment

https://github.com/TryQuiet/quiet/issues/1558
https://github.com/TryQuiet/quiet/issues/1559
https://github.com/TryQuiet/quiet/issues/1649


packages/desktop/src/renderer/components/widgets/channels/ChannelMessages.tsx
```diff
@@ -152,6 +153,7 @@ export const ChannelMessagesComponent: React.FC<IChannelMessagesProps & FileActi
                     downloadFile={downloadFile}
                     cancelDownload={cancelDownload}
                     onMathMessageRendered={onMathMessageRendered}
+                    // place for userLabel props
```
siepra, line 156: Is this comment neccessary?

packages/desktop/src/renderer/components/widgets/channels/BasicMessage.tsx
```diff
@@ -207,6 +210,11 @@ export const BasicMessageComponent: React.FC<BasicMessageProps & FileActionsProp
                       {infoMessage ? 'Quiet' : messageDisplayData.nickname}
                     </Typography>
                   </Grid>
+                  {userLabel && (
+                    <Grid>
+                      <UserLabelContainer type={userLabel} username={messageDisplayData.nickname} />
```
siepra, line 215: Why did you add "container" suffix?

siepra: Ok, I see. Won't it break storybook/snapshot testing? I guess there should probably not be a separate container for user label only.

packages/desktop/src/renderer/sagas/modals/modals.types.ts
```diff
@@ -21,4 +21,6 @@ export enum ModalName {
   searchChannelModal = 'searchChannelModal',
   warningModal = 'warningModal',
   channelCreationModal = 'channelCreationModal',
+  duplicateModal = 'duplicateModal',
```
siepra, line 24: I'm missing the information what specifically is duplicated. I think naming it `duplicatedUserModal` or `duplicatedUsernameModal` would be better

packages/desktop/src/renderer/sagas/modals/modals.types.ts
```diff
@@ -21,4 +21,6 @@ export enum ModalName {
...
+  unregisteredModal = 'unregisteredModal',
```
siepra, line 25: Same as above

packages/desktop/src/renderer/components/widgets/userLabel/UserLabel.component.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+const UserLabel: React.FC<UserLabelProps> = ({
+  type,
+  username,
+  unregisteredUsernameModalHandleOpen,
+  duplicatedUsernameModalHandleOpen,
+}) => {
+  const isUnregistered = type === UserLabelType.UNREGISTERED
+  const handleOpen = isUnregistered ? unregisteredUsernameModalHandleOpen : duplicatedUsernameModalHandleOpen
```
EmiM, line 61: How does it look when user is both unregistered and has duplicated username?

Kacper-RF: I built it to provide one label (just like on Figma), we can decide what is more important in that situation and display that label as quick solution.

For long term solution in next iteration we can work on possibility to show 2 different labels

siepra (approved):

EmiM (approved):

PR #1766, opened 2023-09-06
fix: Add patch to requirements documentation

leblowl:
I noticed that `patch` was not installed on my system, leading to Electron being
imported in the backend, which then caused an error: "Electron failed to install
correctly, please delete node_modules/electron and try installing again". The
build steps are currently written such that they succeed even if the patch
command fails. This works for now since it makes them idempotent, since patch
fails if it cannot apply changes (e.g. the changes are already applied).
However, it also hides the case where patch isn't installed. I'd like to ensure
patch is installed and provide an error otherwise, but for now we can just
update the documentation to let users know it's required.

### Pull Request Checklist

### DIFF
--- a/packages/backend/README.md
+++ b/packages/backend/README.md
@@ -4,10 +4,11 @@ Communication manager for Quiet project. Uses OrbitDB, Libp2p, Tor and websocket
 
 ### Installation
 
-Requirements: 
+Requirements:
 - node@12
 - typescript
 - ts-node
+- patch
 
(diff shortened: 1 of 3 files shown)


siepra (approved):

leblowl:
I reverted some of the changes in this PR because it was failing on Windows, we can revisit this later to implement an automated solution if it comes up again. For now adding to the documentation seems sufficient.

PR #1767, opened 2023-09-06
Clarify autoupdate language

### LINKED ISSUE #1540: Clarify autoupdate language
Version - it's a long known issue, no specific version, but when I'm creating this task we are testing v. quiet@1.2.1-alpha.7

What happens: if the app restarted when the new version exist we update the app. Use can't do anything to prevent this update. 

What should happen: update should take place only when the user choose to update the app through our updated pop-up.

leblowl:
When a new update is available, it is applied automatically on restart. The update modal now let's you know that an update is available and allows you to restart now or later.

Fixes #1540

### Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,3 +9,5 @@
 * Fixed bug with displaying incorrect default settings tab.
 
 * Fixed LoadingPanel useEffect bug.
+
+* Clarify autoupdate language in update modal to let users know that the app will update on restart.
--- a/packages/desktop/src/renderer/components/widgets/update/UpdateModal.test.tsx
+++ b/packages/desktop/src/renderer/components/widgets/update/UpdateModal.test.tsx
@@ -84,7 +84,7 @@ describe('UpdateModal', () => {
                 style="width: 600px;"
               >
                 <div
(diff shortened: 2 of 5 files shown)


leblowl:
Would you like for me to add changes like this to the CHANGELOG?

holmesworcester:
We're trying to do a better job of maintaining the changelog and just getting a feel for it. I think it's worth adding small things, yes.

holmesworcester:
Do we know why the tests are failing? I see one failure for chromatic, which makes sense, but what about the 'two clients' test? @Kacper-RF did you document the setup process for chromatic? I've shared the secret with Lucas.

Kacper-RF:
We have some unexpected problems with joining to community on Linux which shows very randomly on e2e tests and real app, we are already looking at this issue.

Please add Lucas to organization TryQuiet as member, and then problem with `Visual regressions/chromatic` should be solved.

@leblowl please run linter before next commit to pass `Check` step

holmesworcester:
Do you mean in github or in Chromatic?

Kacper-RF:
In GitHub, in Chromatic as far as I remember it should be possible to login via GitHub account.

leblowl:
Closing/reopening this to test Chromatic

PR #1799, opened 2023-09-12
Chore/replace disabling inputs with splash

siepra:
### Pull Request Checklist


leblowl:
Why are tests being moved to the __tests__ directory?

siepra:
We don't really have any strict policy on tests location. I decided to use `toMatchSnapshot`instead of the inline one, because the snapshot is very long and it makes the test harder to read. At the same time I thought that moving it to a separate directory will make the whole component directory less messy

Kacper-RF (commented):
Could you leave the version information on the splash screen until there is an "About" tab in mobile settings where the user can see the current version? I think this is important for QA and future debugging purposes.

leblowl:
idea: I might squash commits to keep the git history on develop a little cleaner/easier to undo

siepra:
@leblowl I always squash, although recently I've been thinking about some more strict naming policy like prefixes etc.

Kacper-RF (approved):

leblowl:
@siepra nice, naming policy for commits? Like https://www.conventionalcommits.org/en/v1.0.0/?

Issue #540, opened 2022-05-27
User should be able to write `:smile:` for 😄 and use other emoji codes


pdurbin:
As a keyboard junkie, in Element, Slack, and GitHub I very frequently write `:smile:` to get the 😄 emoji. I would prefer not to reach for the mouse to click the emoji button in Quiet:

<img width="184" alt="Screen Shot 2022-05-27 at 9 31 57 AM" src="https://user-images.githubusercontent.com/21006/170709461-4b162a57-a7e2-4162-a4c7-fc2cea461c32.png">

I'm aware that a workaround on Mac is to hit Ctrl-Cmd-Spacebar to bring up a native emoji picker...

<img width="788" alt="Screen Shot 2022-05-27 at 9 35 14 AM" src="https://user-images.githubusercontent.com/21006/170710150-ea0f2fd0-43ff-447c-b244-dd1a27ed4067.png">

... but it's faster and easier for me to just type `:smile:` like I do in other places.

Tab completion and search/autocomplete would be nice (see https://github.blog/2012-10-12-emoji-autocomplete/ ). That is, here on GitHub I only have to type `:smi` and then hit the Tab key to select from a list. Element and Slack work the same way.

![Screen Shot 2022-05-27 at 9 45 52 AM](https://user-images.githubusercontent.com/21006/170712012-0f448ac2-31f1-4912-8bdc-0fe252f86062.png)

Some places such as https://github.com/ikatyang/emoji-cheat-sheet seem to refer to these as emoji shortcodes for anyone searching issues for that term.

holmesworcester:
Thanks for this! 

Just noting for when we get to this that the visual autocomplete interacts with @ mentions, as the UI is similar, so we might want to tackle it at the same time. This library might work for both, or we could use the existing @ mention code, though it might not consider all the cases: https://github.com/webscopeio/react-textarea-autocomplete. 

We could also look at what other editors or open source chat apps use for this.

pdurbin:
https://github.com/nolanlawson/emoji-picker-element seems promising. I heard about it at https://shoptalkshow.com/451/ when they interviewed @nolanlawson.

holmesworcester:
This seems like a good proposal! https://chat.openai.com/share/1491b99e-d50c-463f-af00-d8e389644bc1

holmesworcester:
@pdurbin should we assign this to you?

pdurbin:
[...]

PR #1822, opened 2023-09-17
write `:smile:` for 😄, `:tada:` for 🎉, `:rocket:` for 🚀 #540

pdurbin:
### Pull Request Checklist

This is just a draft pull request for now. @holmesworcester and I were talking about it.

It's for this issue:

- #540

So far I'm only supporting three emoji codes (smile, tada, and rocket). And there's no preview, no picker. I was just happy to get something working. 😅 

Here are a couple screenshots:

Screenshot 2023-09-16 at 11 00 15 PM

Screenshot 2023-09-16 at 11 00 21 PM

### DIFF
--- a/packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.tsx
+++ b/packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.tsx
@@ -481,6 +481,7 @@ export const ChannelInputComponent: React.FC<ChannelInputProps> = ({
         } else if (inputStateRef.current === INPUT_STATE.AVAILABLE) {
           e.preventDefault()
           const target = e.target as HTMLInputElement
+          target.innerText = replaceEmojicodesWithEmoji(target.innerText);
           onChange(target.innerText)
           onKeyPress(target.innerText)
           setMessage('')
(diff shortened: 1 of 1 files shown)


holmesworcester:
Looks great! I think if we had a complete emojimap and a couple tests this would be great!

pdurbin:
Ha. Thanks. Just hacking for fun. 😜 

For the complete emojimap, we should probably tap in to our existing emoji library, right? It looks like @siepra added https://www.npmjs.com/package/emoji-picker-react in 654b315. I see the data I want over at https://github.com/ealush/emoji-picker-react/blob/4.0.0/src/data/emojis.json for example. In GitHub I type `:sweat_smile:` to get 😅 and that JSON file has pretty much the same, but with spaces instead of underscores and no colons...

    {
      "n": [
        "sweat smile",
        "smiling face with open mouth and cold sweat"
      ],
      "u": "1f605",
      "a": "0.6"
    },

... but we can probably work with that. From the emoji-picker-react docs I'm not sure how to get at the data. I'm not sure where to pass in `sweat smile` to get the emoji I want, I mean. This issue suggests that maybe the data needs to be exported:

- https://github.com/ealush/emoji-picker-react/issues/349

Does that sound right?

Also, where should I add tests, please?

Finally, two tests are failing in this PR:

- Check / check (ubuntu-20.04) (pull_request)
- Visual regressions / chromatic-deployment (ubuntu-20.04) (pull_request)

Is this expected for outside contributors? Thanks!

holmesworcester:
I would be okay with duplicating the data by hand if that's the best solution.

Another thing we could do is look at how matternost solves this problem. There might be another library that is useful.

pdurbin:
I'm not actively working on this so I'm going to close this pull request to reduce noise and clutter. Maybe a future developer and at least see which files need to be touched by looking at it.

PR #1843, opened 2023-09-21
feature/Users who request already-taken usernames should be invited to request a new username

Kacper-RF:
### Pull Request Checklist


CHANGELOG.md
```diff
@@ -33,3 +33,5 @@
 * C4 for Quiet architecture. Context and Container diagrams.
 
 * Invite tab as default in settings
+
```
EmiM, line 36: This should be put into [unreleased] section on top of the changelog

packages/state-manager/src/sagas/identity/identity.selectors.ts
```diff
@@ -37,6 +38,27 @@ export const csr = createSelector(communitiesSelectors.currentCommunityId, selec
...
 })
 
+export const usernameTaken = createSelector(currentIdentity, csrsMapping, (identity, csrs) => {
+  const userCertificate = identity?.userCertificate
+  if (userCertificate) return false
+
+  const username = identity?.nickname
+  if (!username) return false
+
+  const allUsernames: string[] = Object.values(csrs).map(u => u.username)
```
leblowl, line 48: These are just CSRs? Shouldn't we be checking against signed certificates? Since a username isn't valid until the certificate has been signed by the owner.

Kacper-RF: Thanks for catching this

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -849,12 +830,6 @@ export class StorageService extends EventEmitter {
       this.logger('Certificate is either null or undefined, not saving to db')
       return false
     }
-    const verification = await verifyUserCert(payload.rootPermsData.certificate, payload.certificate)
```
EmiM, line 852: Why did you remove that?

packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -1,178 +1,68 @@
...
   }
 
   public async registerOwnerCertificate(payload: RegisterOwnerCertificatePayload): Promise<void> {
-    let cert: string
-    try {
-      cert = await registerOwner(payload.userCsr.userCsr, payload.permsData)
-    } catch (e) {
-      this.logger.error(`Registering owner failed: ${e.message}`)
+    // It should not be here.
+    this._permsData = payload.permsData
```
EmiM, line 45: Why it's being set here? Did you remove calling LAUNCH_REGISTRAR? I think permsData was set on that event.

packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -1,178 +1,68 @@
...
-      if (this.pendingPromise) return
-      this.pendingPromise = this.registerUser(req.body.data)
-      const result = await this.pendingPromise
-      if (result) {
-        res.status(result.status).send(result.body)
+    this.on(
+      RegistrationEvents.REGISTER_USER_CERTIFICATE,
+      async (payload: { csrs: string[]; certificates: string[] }) => {
+        // Lack of permsData means that we are not the owner of the community
+        if (!this._permsData) return
```
leblowl, line 23: Also I think it would be good to move this into `issueCertificates` to keep all the logic together.

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -397,36 +397,17 @@ export class StorageService extends EventEmitter {
...
-      this.emit(StorageEvents.REPLICATED_CSR, allCsrs)
+      const allCertificates = this.getAllEventLogEntries(this.certificates)
+      this.emit(StorageEvents.REPLICATED_CSR, { csrs: allCsrs, certificates: allCertificates })
       await this.updatePeersList()
     })
     this.certificatesRequests.events.on('write', async (_address, entry) => {
       const csr: string = entry.payload.value
       this.logger('Saved CSR locally')
-      this.emit(StorageEvents.REPLICATED_CSR, [csr])
+      this.emit(StorageEvents.REPLICATED_CSR, { csrs: [csr], certificates: [] })
```
leblowl, line 410: Why use an empty certificates list here?

packages/backend/src/nest/storage/storage.service.ts
leblowl: I'd recommend adding a access controller for that verifies certificate validity. This sort of validation needs to happen on each peer when receiving (replicating) data.

packages/desktop/src/renderer/components/CreateUsername/CreateUsernameComponent.tsx
```diff
@@ -188,77 +231,137 @@ export const CreateUsernameComponent: React.FC<CreateUsernameComponentProps> = (
...
+                        field.onBlur()
+                      }}
+                      value={field.value}
+                    />
+                  )}
+                />
+
+                {!isNewUser && (
+                  <Typography variant='caption' style={{ marginTop: 8 }}>
+                    Your username will be public, but you can choose any name you like. No spaces or special characters.
```
leblowl, line 304: @holmesworcester seems like "public" is confusing here

packages/desktop/src/renderer/components/CreateUsername/CreateUsernameComponent.tsx
```diff
@@ -188,77 +231,137 @@ export const CreateUsernameComponent: React.FC<CreateUsernameComponentProps> = (
...
+                    Lowercase letters and numbers only.
```
leblowl, line 305: This validation should be added to certificatesRequests access controller if we'd like to enforce it.

packages/mobile/src/components/Registration/UsernameRegistration.component.tsx
```diff
@@ -40,6 +47,12 @@ export const UsernameRegistration: FC<UsernameRegistrationProps> = ({
     const parsedName = parseName(name)
     setUserName(parsedName)
     setParsedNameDiffers(name !== parsedName)
+    if (registeredUsers && !isNewUser) {
+      const allUsersArr = Object.values(registeredUsers).map(user => user.username)
+      if (allUsersArr.includes(name)) {
```
leblowl, line 52: Can use a Set here

packages/state-manager/src/sagas/identity/identity.selectors.ts
```diff
@@ -37,6 +38,22 @@ export const csr = createSelector(communitiesSelectors.currentCommunityId, selec
   return identities[id]?.userCsr
 })
 
+export const usernameTaken = createSelector(currentIdentity, certificatesMapping, (identity, certs) => {
+  const userCertificate = identity?.userCertificate
+  if (userCertificate) return false
+
+  const username = identity?.nickname
+  if (!username) return false
```
leblowl, line 46: Should this be `if (username) return false`?

leblowl (changes requested):
Thanks for working on this @Kacper-RF, I have some initial thoughts and a little bit more left to review.

packages/desktop/src/renderer/components/CreateUsername/CreateUsernameComponent.tsx
holmesworcester: My memory is that we added this when someone was surprised that their name was visible to others, but I may be misremembering. Let's remove it.

holmesworcester: "You can choose any username you like. No spaces or special characters."

packages/backend/src/nest/registration/registration.functions.ts
```diff
@@ -37,252 +17,70 @@ class UserCsrData {
...
+  const pendingCsrs = payload.csrs.filter(csr => {
+    const username = getReqFieldValue(parsedCsrs[csr], CertFieldsTypes.nickName)
 
-  const registrarResponse: UserCertificatePayload = await response.json()
+    if (!username) return false
+    if (certNames.includes(username)) return false
+    if (pendingNames.includes(username)) return false
+    pendingNames.push(username)
+    return true
+  })
```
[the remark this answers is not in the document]
vinkabuki: I did that at first  and it didn't work because await in callback function in filter method

packages/backend/src/nest/storage/storage.service.ts
vinkabuki: @EmiM Because it doesn't make sense to verify that on write, only owner adds certificates and he trusts himself. 
@leblowl I can do that, but keep in mind that it will break the current community, adding access controller to existing db means creating new database and losing access to the old set of certificates - should I do that now? in other words - after releasing our community will break

packages/backend/src/nest/registration/registration.service.ts
vinkabuki: Yeah and it was too late in case owner registering his own cert - should I figure this out now?

packages/state-manager/src/sagas/identity/identity.selectors.ts
vinkabuki: no, this is for TS mainly, and what you're proposing is that if user picked any username that means that this username is not taken which is obviously not true.

packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -1,178 +1,68 @@
...
-    this._app.post('/register', async (req, res): Promise<void> => {
-      if (this.pendingPromise) return
-      this.pendingPromise = this.registerUser(req.body.data)
-      const result = await this.pendingPromise
-      if (result) {
-        res.status(result.status).send(result.body)
+    this.on(
+      RegistrationEvents.REGISTER_USER_CERTIFICATE,
+      async (payload: { csrs: string[]; certificates: string[] }) => {
+        // Lack of permsData means that we are not the owner of the community
```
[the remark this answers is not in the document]
vinkabuki: Only after manipulating the source code - as well as many many other things in code, practically you could add this kind of note to every method in the codebase - should it be our standard practice?

packages/backend/src/nest/storage/storage.service.ts
vinkabuki: I'll unify that with replicate.progress behavior, it's because only owner needs csrs and certificates for registrar purpose and since he's always the first in community - there are no certificates yet

packages/backend/src/nest/storage/storage.service.ts
vinkabuki: @leblowl I think this needs a separate discussion on how to refactor storage module (it was planned for a very long time already), and how to actually design it to make it stronger in terms of both backward compatibility and upgrades. I can share my thoughts on that. I learned that the hard way by being unaware that adding access controllers means creating new db and I've dropped all the channels in community.

packages/desktop/src/renderer/components/CreateUsername/CreateUsernameComponent.tsx
Kacper-RF: I talked with Bartek about general topic of Access Controllers and we think it's worth doing it as a separate task/PR and first thinking about all the places where to add them.

packages/backend/src/nest/registration/registration.service.ts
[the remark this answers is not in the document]
leblowl: I think it is helpful to add notes throughout the codebase whenever there is ambiguity or complexity that isn't completely obvious from the code. I think the Linux kernel is a great example of code that uses comments effectively: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/mm/zswap.c?id=ca56489c2f1df516801eeff06129d2e9e282ab7b#n632 ... it can be much easier for someone reading the code later to understand what is happening and why.

If having permsData doesn't unconditionally mean that we are the owner of the community, then we shouldn't label it as such. If there is more to the story, I think we should say so.

Security only comes from server-side operations. So we can't trust a client to always send us perfect data. Even if it requires them to rewrite their client-side code in order to do so. It's like a cashier not counting cash at the grocery store. In the p2p model, each peer is their own server. So before a peer accepts anything, they need to verify for themselves that it's acceptable.

packages/backend/src/nest/registration/registration.functions.ts
```diff
@@ -37,252 +17,70 @@ class UserCsrData {
...
-  requestTimeout = 120000,
-  socksProxyAgent: Agent
-): Promise<RegistrationResponse> => {
-  const controller = new AbortController()
-  const timeout = setTimeout(() => {
-    controller.abort()
-  }, requestTimeout)
+export const extractPendingCsrs = async (payload: { csrs: string[]; certificates: string[] }) => {
+  const certNames: string[] = []
+  const pendingNames: string[] = []
```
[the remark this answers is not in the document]
leblowl: @vinkabuki Any thoughts on this?

packages/backend/src/nest/registration/registration.functions.ts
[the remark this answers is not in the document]
leblowl: One solution to that: https://stackoverflow.com/a/42497383

Another option is to simplify like so:
```typescript
  const pendingCsrs: CertificationRequest[] = []

  for (csr of payload.csrs) {
    const parsedCsr = await loadCSR(csr)
    const username = getReqFieldValue(parsedCsr, CertFieldsTypes.nickName)

    if (username && !certNames.includes(username) && !pendingNames.includes(username)) {
      pendingNames.push(username)
      pendingCsrs.push(csr)
    }
  }

```

packages/backend/src/nest/storage/storage.service.ts
leblowl: @vinkabuki I'm okay with implementing this in a separate PR. We should make an issue.

packages/identity/src/common.ts
```diff
@@ -149,3 +142,29 @@ export const getReqFieldValue = (
...
+
+// generalize to certificateByField
+export const certificateByUsername = (username: string, certificates: string[]): string | null => {
+  /**
+   * Check if given username is already in use
+   */
+  for (const cert of certificates) {
+    const parsedCert = parseCertificate(cert)
+    const certUsername = getCertFieldValue(parsedCert, CertFieldsTypes.nickName)
+    if (certUsername?.localeCompare(username, undefined, { sensitivity: 'base' }) === 0) {
```
leblowl, line 165: Why are we using localeCompare here?

packages/state-manager/src/sagas/identity/registerUsername/registerUsername.saga.ts
```diff
@@ -37,19 +42,43 @@ export function* registerUsernameSaga(socket: Socket, action: PayloadAction<stri
...
+        peerId: identity.peerId.id,
+        dmPublicKey: identity.dmKeys.publicKey,
+        signAlg: config.signAlg,
+        hashAlg: config.hashAlg,
+        existingKeyPair,
+      }
 
-  if (userCsr === null) {
+      userCsr = yield* call(createUserCsr, payload)
+      // console.log({ userCsr })
```
leblowl, line 76: Let's remove this comment

packages/state-manager/src/sagas/identity/registerUsername/registerUsername.saga.ts
```diff
@@ -37,19 +42,43 @@ export function* registerUsernameSaga(socket: Socket, action: PayloadAction<stri
...
   }
 
-  let userCsr = null
+  let userCsr = identity.userCsr
 
-  // Reuse the same csr if nickname hasn't changed
-  if (identity.nickname === nickname) {
-    userCsr = identity.userCsr
-  }
+  if (userCsr !== null) {
```
leblowl, line 53: Can userCsr also be undefined?

packages/desktop/src/renderer/components/CreateUsername/CreateUsernameComponent.tsx
leblowl: I agree, I added some issues to track this:
- https://github.com/TryQuiet/quiet/issues/1893
- https://github.com/TryQuiet/quiet/issues/1899

packages/backend/src/nest/storage/storage.service.ts
leblowl: https://github.com/TryQuiet/quiet/issues/1893
https://github.com/TryQuiet/quiet/issues/1899

packages/desktop/src/renderer/components/widgets/usernameTakenModal/UsernameTakenModal.container.tsx
```diff
@@ -0,0 +1,49 @@
...
+      )
+    },
+    [dispatch]
+  )
+
+  useEffect(() => {
+    if (isUsernameTaken) {
+      usernameTakenModal.handleOpen()
+
+      //IMPORTANT: temporary solution if somehow user recive once wrong information from selector
```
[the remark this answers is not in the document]
Kacper-RF: Yes, that's true, but I've had weird situations where somehow a user who shouldn't get this modal got this -> the selector returned true once, but after that everything was fine because it returned false, so that's why I had to add this line in else

packages/identity/src/common.ts
vinkabuki: I don't even know what localeCompare is, I just moved the method from backend package to common package without paying attention to its implementation

packages/backend/src/nest/registration/registration.service.ts
vinkabuki: Maybe I was not clear: we need both. LAUNCH_REGISTRAR sets permsData if owner already have certificate, if he doesn't have a certificate it's being set when issuing certificate. It should be set by launch_registrar always in my opinion but this involve additional work of figuring out what's wrong in state-manager and make a scope of this PR much bigger possibly, lmk if you want mi to chase it now @leblowl

packages/backend/src/nest/registration/registration.service.ts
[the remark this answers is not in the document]
vinkabuki: Ok that's good idea, I've added a comment

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -178,10 +174,6 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
     if (this.tor && !options.saveTor) {
       await this.tor.kill()
     }
-    if (this.registrationService) {
-      this.logger('Stopping registration service')
-      await this.registrationService.stop()
-    }
```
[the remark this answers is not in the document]
vinkabuki: Because in the old model this service needed to be initialized - start listening on express server, add routing etc. Now it has just methods, so 'stop' method became empty.

packages/backend/src/nest/registration/registration.service.ts
leblowl: Ok, I think that makes sense. Can we add some more detail to the comment:
```
    // It should not be here.
```
to explain a little more why it shouldn't be there.

packages/state-manager/src/sagas/socket/startConnection/startConnection.saga.ts
```diff
@@ -246,12 +246,6 @@ export function subscribe(socket: Socket) {
           ownerCertificate: payload.network.certificate,
         })
       )
-      emit(
-        communitiesActions.storePeerList({
-          communityId: payload.communityId,
-          peerList: payload.network.peers,
-        })
-      )
```
leblowl, line 254: Why are we removing this?

leblowl (changes requested):

packages/state-manager/src/sagas/socket/startConnection/startConnection.saga.ts
vinkabuki: You noticed that we do not use peers anymore in the registrar for the owner - it's state-manager part of it, we were passing empty array here.

packages/identity/src/common.ts
vinkabuki: ok, thanks

packages/backend/src/nest/registration/registration.functions.ts
```diff
@@ -37,252 +17,68 @@ class UserCsrData {
...
-    data: {
-      communityId: communityId,
-      payload: registrarResponse,
-    },
+    if (username && !certNames.has(username) && !pendingNames.has(username)) {
+      pendingNames.add(username)
+      pendingCsrs.push(csr)
+    }
   }
+  return pendingCsrs
```
leblowl, line 52: I have one more thought on this. I think we need to take a little more care here. We only want to process a single CSR per public key and we want to process the latest CSR submitted for each public key, yet at the same time, we want to process CSRs in order. Otherwise, if we don't process the latest CSR per public key, then nobody would be able to change names because we would always process their first submitted CSR.

It seems like we need to reverse the CSR list, find the first CSR for each public key (which gives us the most recent CSR for each public key). Then reverse that list so that it is back in order. Then filter CSRs to remove CSRs for usernames that are already taken. Then remove duplicates by taking the first CSR with a specific name.

leblowl (changes requested):

leblowl (approved):

PR #1863, opened 2023-09-26
Chore/lerna publishing config

siepra:
### Pull Request Checklist

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -5,7 +5,7 @@
   "version": "independent",
   "command": {
     "version": {
-      "allowBranch": ["master", "develop", "workflows/require-e2e-before-build-release"],
+      "allowBranch": ["master", "develop", "workflows/require-e2e-before-build-release", "chore/lerna-publishing-config"],
       "conventionalCommits": true,
       "createRelease": "github"
     },
--- a/packages/e2e-tests/package.json
+++ b/packages/e2e-tests/package.json
@@ -3,6 +3,7 @@
   "version": "2.0.1-alpha.1",
   "description": "For running end to end tests we use  Selenium WebDriver (https://www.selenium.dev/documentation/webdriver/) with Electron ChromeDriver (https://www.npmjs.com/package/electron-chromedriver) in Jest",
   "main": "index.js",
+  "private": true,
   "scripts": {
     "build": "tsc -p tsconfig.build.json",
     "prepare": "npm run build",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "name": "root",
   "private": true,
   "scripts": {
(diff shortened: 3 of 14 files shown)


lerna.json
```diff
@@ -5,7 +5,7 @@
   "version": "independent",
   "command": {
     "version": {
-      "allowBranch": ["master", "develop", "workflows/require-e2e-before-build-release"],
+      "allowBranch": ["master", "develop", "workflows/require-e2e-before-build-release", "chore/lerna-publishing-config"],
```
leblowl, line 8: Should we remove these two? `workflows/require-e2e-before-build-release` and `chore/lerna-publishing-config`

packages/desktop/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "quiet",
+  "name": "@quiet/desktop",
```
leblowl, line 2: I like this since we will now have @quiet/desktop and @quiet/mobile. I think it's more clear

packages/e2e-tests/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "2.0.1-alpha.1",
   "description": "For running end to end tests we use  Selenium WebDriver (https://www.selenium.dev/documentation/webdriver/) with Electron ChromeDriver (https://www.npmjs.com/package/electron-chromedriver) in Jest",
   "main": "index.js",
+  "private": true,
```
leblowl, line 6: How does private affect things? Does this prevent Lerna from bumping versions or just from publishing?

lerna.json
siepra: I don't think so. Of course I'll remove the last one after I transform this draft to PR, but `workflows/require-e2e-before-build-release` I'm not sure whether we still use. @Kacper-RF can you confirm?

packages/desktop/package.json
siepra: Yes! I only want to make sure it won't break anything so whoever wants to jump in and help me with testing, please do

packages/e2e-tests/package.json
siepra: It doesn't do anything until you apply `--no-private` flag to the `publish` or `version` command. I think it's best to just keep those "internal" packages marked as private but to continue to version them as we already do

lerna.json
Kacper-RF: I think we don't use it at the moment, we triggering E2E tests before releasing but from `build-release.yml` workflow as separate jobs.

siepra: Ok, then I'll remove it from here

leblowl (approved):

Issue #1818, opened 2023-09-15
Lerna bootstrap failure on windows


Ahmedouu:
tried this a few times already on linux and windows and it keeps failing just on windows.

![image](https://github.com/TryQuiet/quiet/assets/33392644/a0b9620c-92f4-41c1-870e-e6bdbdb4caa3)

EmiM:
Could you go to packages/backend and run `npm run prepare`? It will tell us more what happened. It may be a problem with patch.

yaronarik:
I had the same issue



![npm run prepare](https://github.com/TryQuiet/quiet/assets/94235955/798d4778-53b7-40fb-94e3-6b98ff4ee900)

vinkabuki:
You're missing patch lib.
Quick solution is to remove "applyPatches" in prepare script in backend package. applyPatches is just for mobile platforms.

Ahmedouu:
@vinkabuki  I tried that but prepare keeps failing to run cause of the websockts module: 
![image](https://github.com/TryQuiet/quiet/assets/33392644/bdef30d6-0331-4eaf-8363-41120db7c244)

vinkabuki:
@Ahmedouu It is just a warning, do you see something like "webpack 5.75.0 compiled with 6 warnings in 10857 ms" ? You should, it's the last output line after running npm run prepare

Ahmedouu:
yeah I see it then I start the desktop package I get this websocket error:

![image](https://github.com/TryQuiet/quiet/assets/33392644/f37d1da4-4d9c-4d0d-b406-291c9dd04bb1)

then we I try to bootstrap again i get this:
![image](https://github.com/TryQuiet/quiet/assets/33392644/0a159c20-040d-471a-92bd-e213253dae70)

so I removed the prepare script and managed to bootstrap with no errors but still got the websocket error when I start the desktop app..

Ahmedouu:
PS C:\Users\USER\projects\quiet\packages\backend> npm run prepare

> @quiet/backend@2.0.0-alpha.3 prepare
> npm run webpack


> @quiet/backend@2.0.0-alpha.3 webpack
> webpack --env mode=development && cp ./lib/bundle.cjs ../backend-bundle/bundle.cjs

assets by path nest/ 38.2 KiB 67 assets
assets by path *.ts 633 bytes
  asset rn-bridge.d.ts 473 bytes [compared for emit]
  asset backendManager.d.ts 122 bytes [compared for emit]
[...]

PR #1870, opened 2023-09-27
Adjust project bootstrap scripts to be windows-friendly

EmiM:
### Pull Request Checklist


Kacper-RF (commented):
Setup environment step in E2E and check workflow failed

packages/backend/package.json
```diff
@@ -9,14 +9,18 @@
...
-		"webpack": "webpack --env mode=development && cp ./lib/bundle.cjs ../backend-bundle/bundle.cjs",
+		"webpack": "run-script-os",
+		"webpack:default": "webpack --env mode=development && cp ./lib/bundle.cjs ../backend-bundle/bundle.cjs",
 		"webpack:prod": "webpack --env mode=production && cp ./lib/bundle.cjs ../backend-bundle/bundle.cjs",
-		"applyPatches": "patch -f -p0 < ./electron-fetch.patch || true && patch -f -p0 --forward --binary < ./parse-duration.patch || true && patch -f -p0 --forward --binary < ./parse-duration-esm.patch || true",
+		"webpack:windows": "webpack --env mode=development && xcopy .\\lib\\bundle.cjs ..\\backend-bundle\\ /Y",
+		"webpack:prod:windows": "webpack --env mode=production && xcopy .\\lib\\bundle.cjs ..\\backend-bundle\\ /Y",
+		"applyPatches": "run-script-os",
+		"applyPatches:default": "patch -f -p0 < ./electron-fetch.patch || true && patch -f -p0 --forward --binary < ./parse-duration.patch || true && patch -f -p0 --forward --binary < ./parse-duration-esm.patch || true",
+		"applyPatches:windows": "git apply ./electron-fetch-git.patch --whitespace=fix --reject --verbose --no-index --ignore-space-change --inaccurate-eof || cd .",
```
leblowl, line 19: Should we just use git apply on all systems to keep things simple?

packages/mobile/package.json
```diff
@@ -12,12 +12,18 @@
     "lint": "eslint --ext .jsx,.js,.ts,.tsx ./src/ --fix",
     "lint-ci": "eslint --ext .jsx,.js,.ts,.tsx ./src/",
     "gen": "plop",
-    "prepare-backend-assets": "mkdir -p ./nodejs-assets/nodejs-project && cp ../backend/lib/bundle.cjs ./nodejs-assets/nodejs-project/bundle.cjs",
+    "prepare-backend-assets": "run-script-os",
+    "prepare-backend-assets:default": "mkdir -p ./nodejs-assets/nodejs-project && cp ../backend/lib/bundle.cjs ./nodejs-assets/nodejs-project/bundle.cjs",
+    "prepare-backend-assets:windows": "mkdir .\\nodejs-assets\\nodejs-project || cd . && xcopy ..\\backend\\lib\\bundle.cjs .\\nodejs-assets\\nodejs-project\\ /Y",
```
leblowl, line 17: Is this change necessary, doesn't powershell understand both `mkdir`, `cd` and `cp`?

leblowl (commented):
Thanks for looking into this @EmiM!

leblowl:
Does this fix: https://github.com/TryQuiet/quiet/issues/1818?

packages/backend/package.json
EmiM: I totally agree that we should use one tool instead of two for the same purpose but **I think** patch files differ in format between those generated by patch and git.
I'd say: let's do this but not in scope of this MR but separately.

EmiM: What's more - I still have a problem with handling git's patch correctly (on Windows at least). Applying works fine on windows machine that patch was created on but fails on CI or another windows machine. I suppose git may be changing encoding? I am not sure yet.

packages/mobile/package.json
EmiM: `mkdir` differs in behaviour on Windows - it throws error and stops the process when directory already exists. That's why the `|| cd .` hack is here.
Also - yes, powershell has `cp` command but according to people's outputs in the task I saw that they were using regular cmd and cmd in windows does not have `cp`.
Of course we could add suggestion/requirement to the readme to use powershell but I decided that adjusting scripts to both cmd and powershell is not over the top since other commands (like mkdir) also work differently on Windows and Unix-like OSs.

EmiM:
@leblowl almost but I added a note to desktop readme. There still may be an issue with patch.

vinkabuki (approved):

vinkabuki:
LGTM, I've tested it on my local windows machine which used to fail - now it works, I am able to open and use the app

packages/backend/package.json
leblowl: Ok sounds good. Does the issue on CI break our Windows CI in any way?

packages/mobile/package.json
leblowl: Ok, sounds good. Another alternative for the future could be having prepare scripts written in Javascript, so that we can maintain the same code for all platforms.

leblowl:
I'd recommend squashing commits to keep the git history cleaner on develop

packages/backend/package.json
EmiM: It should not. E2E tests pass, other checks too.

Issue #615, opened 2022-06-15
User should be able to add a custom profile image


holmesworcester:
Being able to have custom profile images will make things way more comfortable and fun, and it will make the the conversation view in Quiet look a lot better too.

Requirements:
1. Users should be able to add a profile image
2. Users should be able to change their profile image
4. Both registered and unregistered users should be able to add and change their profile image
3. No one should be able to update the profile image for another user (we should check that profile image changes are signed with the user's message-signing key)
4. Profile images should be visible on desktop and mobile

Nice-to-have (break out into separate issues if not done in this round of work:)

1. Mobile user should be able to add profile image
2. Desktop (and mobile) users should be able to crop profile images

When we release this we have to prioritize #614 so that people can't register the name Quiet, set their profile image to the Quiet logo, and spoof system messages.

holmesworcester:
We'll use whatever libraries are convenient for cropping, so we just need where this displays, and how it appears in onboarding. (What's the best way to encourage using a profile photo?) 

Do Discord and Slack make it easy to pick an avatar from a list, if you don't want to add a photo?

jgaylor:
[Figma](https://www.figma.com/file/BnANosC1KGMUvm8oU2Dr0i/Custom-profile-image?type=design&node-id=13%3A37&t=JFXK1Y663Oj9RwIh-1) and [Loom](https://www.loom.com/share/1088f957a2884fc0b74fbb09b9a54578)

holmesworcester:
This is great. And we'll use whatever the library offers for cropping, e.g.

kingalg:
desktop@2.1.1-alpha.0

It's working. I feel it's more like a draft of a feature than a real feature, and changing things listed in https://github.com/TryQuiet/quiet/issues/2251 would certainly help. But there is no reason to keep this issue open as its requirements have been fulfilled.

Danie10:
I see a profile image could be added on the desktop app, but I could not find a way to add it on the Android mobile app? Am I missing something as the issue was marked completed?

kingalg:
Hi @Danie10, mobile was a nice-to-have for this issue, and we decided to split it and make another issue. If you are interested, you can track this for mobile in https://github.com/TryQuiet/quiet/issues/2065. You are not missing anything by not seeing an option to add a profile picture in your Android app. Feel free to let me know if you have any additional questions or spot something that looks out of order!

Danie10:
[...]

PR #1923, opened 2023-10-05
Add user profile feature

leblowl:
Sorry about the large PR, I can split it up if desired.

This PR adds a user profile feature for desktop.

On the frontend, the user profile sidebar panel simply shows the current user's
name and profile photo. Clicking on the profile photo brings up a context menu
that shows the profile and profile actions (only 'edit profile' for now).
Clicking the 'edit profile' action changes the context menu view such that you
can edit the profile photo. This commit also adds user profile photos to channel
messages. A jdenticon is showed in place of user profile photo if it doesn't
exist.

On the backend, there is a new shared orbitdb user profiles key/value store. The
key/value store maps user public keys with profiles. It uses a custom index so
that we can filter invalid entries thoroughly. In state-manager, a new
saveUserProfile saga is added which signs the profile data and sends it to the
backend.

Relevant chromatic stories:
- https://www.chromatic.com/component?appId=62021a6b9c4ef3003ab65846&csfId=components-sidebarcomponent&buildNumber=2100&k=65272e0e2c99834a79dc0451-1200px-interactive-true&h=29&b=-1
- https://www.chromatic.com/component?appId=62021a6b9c4ef3003ab65846&csfId=components-userprofilemenuprofileview&buildNumber=2100&k=65272e0e2c99834a79dc0445-1200px-interactive-true&h=29&b=-1
- https://www.chromatic.com/component?appId=62021a6b9c4ef3003ab65846&csfId=components-userprofilemenueditview&buildNumber=2100&k=65272e0e2c99834a79dc0444-1200px-interactive-true&h=29&b=-1

Solution for #615

### Pull Request Checklist

screen1
screen


packages/desktop/src/renderer/components/ContextMenu/ContextMenu.component.tsx
```diff
@@ -79,64 +76,66 @@ export const ContextMenu: FC<ContextMenuProps> = ({ visible, handleClose, title,
...
+        padding: 0,
+        width: '100%',
+      }}
+    >
+      {items.map((item, index) => {
+        return <ContextMenuItem {...item} key={index} />
+      })}
+    </List>
+  )
+}
```
leblowl, line 126: I've separated these out to make the context menu more general purpose. Not all context menu views will have item lists.

packages/desktop/src/renderer/sagas/notifications/notifications.click.test.ts
```diff
@@ -16,7 +16,6 @@ const notification = jest.fn().mockImplementation(() => {
 window.Notification = notification
 
 jest.mock('../../../shared/sounds', () => ({
-  ...jest.requireActual('../../../shared/sounds'),
```
leblowl, line 19: Was causing issues after upgrading Jest

packages/desktop/webpack/webpack.config.renderer.dev.js
```diff
@@ -17,7 +17,8 @@ module.exports = {
     extensions: ['.ts', '.tsx', '.js', '.jsx'],
     plugins: [
       new TsconfigPathsPlugin({ configFile: 'tsconfig.build.json' })
-    ]
+    ],
+    conditionNames: ["import", "node"]
```
leblowl, line 21: Changed due to issues with @ipld/dag-cbor. Moving the logic from state-manager to the backend in a future PR will make these changes unnecessary.

packages/state-manager/src/sagas/publicChannels/publicChannels.selectors.ts
```diff
@@ -200,35 +205,53 @@ export const dailyGroupedCurrentChannelMessages = createSelector(displayableCurr
...
+  dailyGroupedCurrentChannelMessages,
+  userProfiles,
+  (groups: MessagesGroupsType, userProfiles: Record<string, UserProfile>) => {
+    const result: MessagesDailyGroups = {}
+    for (const day in groups) {
+      result[day] = groups[day].reduce((merged: DisplayableMessage[][], message: DisplayableMessage) => {
+        if (!merged.length) {
+          // Add profile photo to user's first message
+          if (message.pubKey in userProfiles) {
+            message.photo = userProfiles[message.pubKey].profile.photo
```
leblowl, line 223: The diff is large, but I really just added this line and a similar one below

packages/state-manager/src/sagas/users/userProfile/saveUserProfile.saga.ts
```diff
@@ -0,0 +1,68 @@
...
+  const userProfile: UserProfile = {
+    profile: profile,
+    profileSig: signature,
+    pubKey,
+  }
+
+  console.log('Saving user profile', userProfile)
+
+  yield* apply(socket, socket.emit, applyEmitParams(SocketActionTypes.SAVE_USER_PROFILE, userProfile))
+}
```
leblowl, line 68: Moving most of this logic to the backend in a future PR will make it easier to test and review.

packages/backend/src/nest/storage/UserProfileStore.ts
```diff
@@ -0,0 +1,261 @@
...
+      }
+    }
+  }
+}
+
+export class UserProfileKeyValueIndex extends KeyValueIndex<UserProfile> {
+  constructor() {
+    super(UserProfileStore.validateUserProfileEntry)
+  }
+}
```
leblowl, line 261: The bulk of the backend changes are here. I've used a custom index here so that we can filter entries before they are indexed, otherwise invalid entries could overwrite valid entries because the index is a hash map. I'm doing this instead of using an access controller because it looks like access controllers only validate heads and we really need to validate each entry.

packages/state-manager/src/sagas/users/userProfile/saveUserProfile.saga.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import { config } from '../../users/const/certFieldTypes'
+import { identitySelectors } from '../../identity/identity.selectors'
+import logger from '../../../utils/logger'
+import { usersActions } from '../users.slice'
+import { type Socket, applyEmitParams } from '../../../types'
+
+const log = logger('saveUserProfileSaga')
+
+const fileToBase64String = (file: File): Promise<string> => {
```
siepra, line 19: Wouldn't it be more universal if we put this method in common package?

packages/state-manager/src/sagas/users/userProfile/saveUserProfile.saga.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+    reader.onerror = reject
+    reader.readAsDataURL(file)
+  })
+}
+
+export function* saveUserProfileSaga(socket: Socket, action: PayloadAction<{ photo?: File }>): Generator {
+  const identity = yield* select(identitySelectors.currentIdentity)
+
+  if (!identity?.userCsr || !action.payload.photo) {
```
siepra, line 38: I'm not sure if this is the case so don't bother if it's not important in this context but if you want to make sure user is a part of a community you should not only check the presence of csr, but also a replication of the community's metadata. On current develop there's a "communityMembership" selector in the identity structure

packages/state-manager/src/sagas/users/userProfile/saveUserProfile.saga.ts
```diff
@@ -0,0 +1,68 @@
...
+      } else {
+        reject(new Error(`Failed to read file, result: ${reader.result}`))
+      }
+    }
+    reader.onerror = reject
+    reader.readAsDataURL(file)
+  })
+}
+
+export function* saveUserProfileSaga(socket: Socket, action: PayloadAction<{ photo?: File }>): Generator {
```
siepra, line 35: I think I'd be good to add test for this saga

packages/state-manager/src/sagas/publicChannels/publicChannels.selectors.ts
siepra: How about moving it to the `displayableCurrentChannelMessages` selector so that each message has a sender photo assigned? This way it wouldn't be necessary to duplicate this statement a few lines below

packages/identity/src/verifySignature.ts
```diff
@@ -12,3 +12,14 @@ export const verifySignature = async (
   const algorithm = getAlgorithmParameters(config.signAlg, 'verify')
   return await crypto.verify(algorithm.algorithm as Algorithm, publicKey, signature, Buffer.from(message))
 }
+
+export const verifyDataSignature = async (
+  signature: ArrayBuffer,
+  data: Uint8Array,
```
siepra, line 18: As the only difference here is data type, why not to extend the type and keep it a single method? 
```
data: string | Uint8Array
```

packages/identity/src/sign.ts
```diff
@@ -11,3 +11,12 @@ export const sign = async (message: string, privKey: CryptoKey): Promise<ArrayBu
 
   return await crypto.sign(algorithm.algorithm as Algorithm, privKey, messageBuffer)
 }
+
+export const signData = async (data: Uint8Array, privKey: CryptoKey): Promise<ArrayBuffer> => {
```
siepra, line 15: It wouldn't be necessary to duplicate method if we extended data type
```
data: string | Uint8Array
```

siepra:
I wouldn't be myself if I didn't ask: what about mobile?

siepra:
I roughly looked at desktop package and UI tests and it looks fine, but I think you should definitely merge develop into your branch as there're changes to the sidebar component there.

Also now as we have working Detox tests on CI, I'm happy to see them passing whenever possible :)

holmesworcester:
I think we should make two separate issues for mobile. One would be to display avatars if they exist. The other would be to add and update them.

On mobile the minimum viable feature for the latter requires quite a bit more work. We need to implement our design for picking photos, and I think we need compression too: a mobile user will never know how to resize a photo manually on their phone, or if they do we shouldn't ask them to. It's already a stretch on desktop but on mobile an image size limit would feel too broken to be releasable, in my view.

The other reason to not go farther now is that this issue was not top of backlog and was chosen as an onboarding task that would give Lucas a good tour of how data moves through the app. I'd rather have Lucas move on to more fundamental stuff on mobile like iOS push notifications experiments.

packages/backend/src/nest/storage/UserProfileStore.ts
```diff
@@ -0,0 +1,261 @@
...
+  }
+  const chars = atob(base64Data)
+  const bytes = new Array(chars.length)
+  for (let i = 0; i < chars.length; i++) {
+    bytes[i] = chars.charCodeAt(i)
+  }
+  return new Uint8Array(bytes)
+}
+
+export class UserProfileStore {
```
Kacper-RF, line 57: In overall I will suggest to do it as injectable provider like another services.
https://docs.nestjs.com/providers

Then you have to import that in `providers` array in storage.module.ts and inject in StorageService

Some advantages of this approach

-We will have consistency between other services and will be consistent with the NestJS documentation
-in future You can inject via constructor another providers like options, or other services
-it will be easier to test that part of code in future with another services

packages/identity/src/sign.ts
leblowl: Same comment above.

packages/state-manager/src/sagas/users/userProfile/saveUserProfile.saga.ts
leblowl: I've tested this manually. I suggest that we consider adding a test in a separate PR. I would like to refactor this by moving it to the backend soon.

packages/state-manager/src/sagas/users/userProfile/userProfile.selectors.ts
```diff
@@ -0,0 +1,23 @@
...
+export const userProfiles = createSelector(usersSlice, users => users.userProfiles)
+
+export const myUserProfile = createSelector(userProfiles, currentIdentity, (userProfiles, identity) => {
+  if (identity?.userCsr) {
+    const pubKey = pubKeyFromCsr(identity.userCsr.userCsr)
+    return userProfiles[pubKey]
+  }
+  return undefined
+})
+
```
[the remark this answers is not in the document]
leblowl: I don't think it's worth adding tests for something like this. I've verified it's working locally.

packages/state-manager/src/sagas/users/userProfile/saveUserProfile.saga.ts
holmesworcester: Generally we want to make sure the tests prove the feature is correct, before we deploy. 

If we think there's a refactor imminent, maybe we could write a test that would work across the refactor, such as an e2e test? Since we don't know when we'll do the refactor, I don't think it's a good reason to stop using testing as we have been.

leblowl: Talking with @holmesworcester, it sounds like the team has made a decision to go all in on automated testing, so I'll add a test here

packages/state-manager/src/sagas/users/userProfile/userProfile.selectors.ts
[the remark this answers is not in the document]
leblowl: I am still not convinced a test here is worth it, but it sounds like there is a team policy around this, so I'll add one.

packages/backend/src/nest/storage/UserProfileStore.ts
```diff
@@ -0,0 +1,261 @@
...
+        return
+      }
+      await this.store.put(userProfile.pubKey, userProfile)
+    } catch (err) {
+      logger.error('Failed to add user profile', err)
+    }
+  }
+
+  public static async validateUserProfile(userProfile: UserProfile) {
+    // FIXME: Add additional validation to verify userProfile contains
```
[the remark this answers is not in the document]
leblowl: I can add a card for it. Basically, I don't know if there is anything enforcing `userProfile` to be a `UserProfile`. I haven't seen anything that would. I think typescript only works at the language level, so it wouldn't let you write a function that references another function if the types were incompatible, but on boundaries (like the network layer) Typescript isn't checking an object at runtime when it comes over the wire. I don't think it's a big deal, because I think the most important validation (that I've thought of) is here, and this function should throw an exception if a field doesn't exist or there is a type mismatch. It's just if we perform object validation, we can provide a more detailed message and catch errors earlier and that's why I've skipped it for now. We can always do better, but I think this is a good start.

packages/backend/src/nest/storage/UserProfileStore.ts
```diff
@@ -0,0 +1,261 @@
...
+    // required fields
```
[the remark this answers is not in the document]
leblowl: I commented above. It might be best to continue this discussion over here, all together: https://github.com/TryQuiet/quiet/pull/1923#discussion_r1363721810

packages/backend/src/nest/storage/UserProfileStore.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  public async close() {
+    await this.store?.close()
+  }
+
+  public async addUserProfile(userProfile: UserProfile) {
+    logger('Adding user profile')
+    try {
+      if (!UserProfileStore.validateUserProfile(userProfile)) {
+        // TODO: Send validation errors to frontend or replicate
```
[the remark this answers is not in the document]
leblowl: https://github.com/TryQuiet/quiet/issues/2043

packages/backend/src/nest/storage/UserProfileStore.ts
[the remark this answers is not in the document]
leblowl: https://github.com/TryQuiet/quiet/issues/2044

packages/backend/src/nest/storage/UserProfileStore.ts
```diff
@@ -0,0 +1,261 @@
...
+      // https://security.stackexchange.com/a/135636
+      const photoBytes = base64DataURLToByteArray(profile.photo)
+      if (!isPng(photoBytes)) {
+        logger.error('Expected PNG for user profile photo', userProfile.pubKey)
+        return false
+      }
+
+      // 200 KB = 204800 B limit
+      //
+      // TODO: Perhaps the compression matters and we should check
```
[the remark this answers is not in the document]
leblowl: https://github.com/TryQuiet/quiet/issues/2045

packages/backend/src/nest/storage/UserProfileStore.ts
```diff
@@ -0,0 +1,261 @@
...
+}
+
+/**
+ * Modified from:
+ * https://github.com/orbitdb/orbit-db-kvstore/blob/main/src/KeyValueIndex.js
+ *
+ * Adds validation function that validates each entry before adding it
+ * to the index.
+ *
+ * TODO: Save latest entry and only iterate over new entries in updateIndex
```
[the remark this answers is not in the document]
leblowl: https://github.com/TryQuiet/quiet/issues/2047

packages/backend/src/nest/storage/UserProfileStore.ts
```diff
@@ -0,0 +1,261 @@
...
+}
+
+/**
+ * Takes a base64 data URI string that starts with 'data:*\/*;base64,'
+ * as returned from
+ * https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsDataURL
+ * and converts it to a Uint8Array.
+ */
+export const base64DataURLToByteArray = (contents: string): Uint8Array => {
+  const [header, base64Data] = contents.split(',')
```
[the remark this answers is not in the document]
leblowl: I think these are helpful to read:
- https://security.stackexchange.com/questions/135513/what-could-an-img-src-xss-do
- https://security.stackexchange.com/a/124843
- https://security.stackexchange.com/questions/258384/can-xss-ever-occur-in-an-img-tags-src-attribute

I don't think JS is allowed in a src tag in Chrome. With external HTTP requests, I think it's possible that someone does a CSRF, but since Quiet is running in it's own isolated browser, the user shouldn't have any session data for other sites. In this PR, we are checking that the binary data starts with the magic byte sequence of a PNG. Unless there is a way for the browser to interpret data starting with that magic byte sequence as some other type of data, I think it should be sufficient to start. Another thing might be adding a CSP and checking that the data url starts with `image/png`. If you can find any other resources that help us improve the safety of the application, we can apply the lessons we learn.

packages/e2e-tests/src/tests/userProfile.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    const text = await elem.getText()
+    expect(text).toEqual(users.owner.messages[0])
+
+    const fullMessages = await generalChannelUser1.getUserMessagesFull(users.owner.username)
+    const img = await fullMessages[1].findElement(By.tagName('img'))
+    await users.user1.app.driver.wait(until.elementIsVisible(img))
+    const imgSrc = await img.getAttribute('src')
+    expect(imgSrc).toEqual(expectedImgSrc)
+  })
+})
```
holmesworcester, line 187: I notice that all of our other multiple client e2e tests are part of a single test run where we test that a few different basic things are working, and that this is the only e2e test that only tests a single feature. 

What are the pros and cons of having a separate run vs. including this test as part of our main run?

For example, I suppose there is an upside to having this test be isolated from other tests, in that perhaps it protects it from giving a false negative.

On the other hand, having this test be part of the multiple clients test means that we are only doing the most time consuming and brittle piece of the process (invitation and initial connection over Tor) once.

@EmiM and @Kacper-RF what do you think?

holmesworcester (commented):
This makes sense to me and I think I understood the important stuff about how it works. Great work! 

I'll approve because I don't think any of my comments required specific changes, but I'm interested in your responses to the questions I raised.

packages/backend/src/nest/storage/UserProfileStore.ts
leblowl: > Got it, so this is what we'll have to do pretty much whenever we want to validate data?

What do you mean?

> Did we create an issue for orbitdb about this yet?

Not yet, I think Haad would prefer to discuss things like this over Matrix, so I intend to do that soon.

packages/e2e-tests/src/tests/userProfile.test.ts
Kacper-RF: Yes exactly the most consuming thing beside whole stuff with building/setting environment on CI is joining to community.

In overall my vision for e2e tests is to include as many cases in one single test to save time when joining.
This has one drawback - maintaining and adding test cases to such a long scenario test is sometimes a bit of a pain.

I think we can watch how long it will take to pass all test files (including the one with the test user profile) on CI and decide if we should move those test cases to `multiple Clients.test.ts`

packages/backend/src/nest/storage/UserProfileStore.ts
holmesworcester: >> Got it, so this is what we'll have to do pretty much whenever we want to validate data?
> What do you mean?

I mean, it looks like we're iterating through every item and validating it, is that right? When do we have to do our own validation in this way? On messages? On channels? 

> Not yet, I think Haad would prefer to discuss things like this over Matrix, so I intend to do that soon.

Yes let's ask soon!

packages/backend/src/nest/storage/UserProfileStore.ts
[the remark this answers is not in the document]
leblowl: I think the most important thing with the `img` tag and untrusted input is making sure the `src` URL is safe (e.g. actually points to an image). Otherwise, I trust the browser to safely display all images. Of course, it may not. But I don't think we need to be too concerned about that right now.

In order to check the `src` URL (which is user-supplied), we verify that it starts with `data:image/png;base64`. We additionally check the magic byte sequence of the payload which might be unnecessary.

I looked at Signal, but I didn't find much in relation to validating images. It does look like they use HTML canvas to resize images which might be a simpler alternative to ImageMagick.

Currently, the user supplies the `src` URL and we validate it. However, another option is to only send the image data instead of the full data URL and then we prepend the `data:image/png;base64,` part ourselves. Then we aren't validating that the URL is correct - we are creating the URL ourselves with the base64 encoded data that the user provides. So less chance that validation breaks.

packages/backend/src/nest/storage/UserProfileStore.ts
leblowl: > I mean, it looks like we're iterating through every item and validating it, is that right? When do we have to do our own validation in this way? On messages? On channels?

Yes that's what we are doing. I think if you want to make sure you validate everything, then this is one way to do it.

packages/backend/src/nest/storage/UserProfileStore.ts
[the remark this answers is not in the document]
holmesworcester: That sounds like a good idea! It gives the attacker less control, right?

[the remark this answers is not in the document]
leblowl: Not really, I think it's just about reducing risk of an if statement not working.

If you have some `<input>` and you are checking it with all these if statements where the final result is that you say that the data is "good" or "safe", then if one of your if statements doesn't work, then you might say the data is "safe" when it might not be.

If, instead of checking input, you just take input and combine it with other pieces of data you wrote to produce a final output that you believe is "safe" or safer, then I think it reduces the risk of missing a conditional.

For example, let's say there are a variety of images that we support. The user could provide a full URL, like `data:image/png;base64,<data>` and we could check to make sure the URL matches our conditions and if so, allow the use of that URL in the application. Or the user could provide a content-type, like `image/png` and `<data>` and we can look up in an object (`{'image/png': 'data:image/png;base64,'}`), the URL `prefix` for a given content-type and then construct the URL with that `prefix` and `<data>`, like `'<prefix><data>'`, or in this case, `'data:image/png;base64,<data>'`. Because we are only using the prefixes that we wrote (as long as we make sure the code actually uses the prefix), the prefix will only ever be a value we wrote or `undefined` if the content-type doesn't exist (in which case we can throw an error).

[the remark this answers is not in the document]
holmesworcester: It sounds like you've thought about this and have made all the changes you wanted, so let's resolve this!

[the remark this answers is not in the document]
leblowl: I added some additional tests around the existing validation

packages/identity/src/verifySignature.ts
siepra: I personally think it's best to do it now. Otherwise the extra issue will get low priority and it's likely we'll postpone it until we forget it

packages/state-manager/src/sagas/users/userProfile/saveUserProfile.saga.ts
siepra: It's not only the team policy but also one important thing about sagas (and one of the main reasons we use them) is that they're extremely easy to test so it comes with a very low cost

PR #1940, opened 2023-10-09
Secure backend socket.io from other applications that can access localhost i.e. browser

Kacper-RF:
### Pull Request Checklist


packages/desktop/src/main/main.ts
```diff
@@ -30,6 +30,8 @@ const updaterInterval = 15 * 60_000
 export const isDev = process.env.NODE_ENV === 'development'
 export const isE2Etest = process.env.E2E_TEST === 'true'
 
+const SOCKET_IO_TOKEN = generateJWT()
```
leblowl, line 33: Why not just use a random number as a token with basic auth? What benefit does a JWT provide here?

siepra (approved):
Ideally those utility functions for eg. generating random string should be moved to some external files/classes on mobile (obj-c code is already complex enough to be hard to read).
It will work though

packages/mobile/ios/Utils.swift
```diff
@@ -0,0 +1,17 @@
...
+
+    for _ in 0..<length {
+        let rand = Int(arc4random_uniform(UInt32(letters.count)))
+        let character = letters[letters.index(letters.startIndex, offsetBy: rand)]
+        randomString.append(character)
+    }
+
+    return randomString
+}
+}
```
leblowl, line 17: Same comment as above

> We should probably use something like: https://developer.apple.com/documentation/security/1399291-secrandomcopybytes
>
> and maybe call it `generateSecret` to match other implementations.

Also it looks like this is a duplicate of the same function in AppDelegate.m

leblowl (changes requested):
Nice job getting this all working! I have a few suggestions, but looks good overall.

packages/backend/src/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+  socketIOSecret: string
+}
+
+// concept
+export const validateOptions = (_options: commander.OptionValues) => {
+  const options = _options as Options
+  if (!options.socketIOSecret) {
+    throw new Error('socketIOSecret is missing in options')
+  }
+}
```
leblowl, line 23: I would probably call this file config.ts or options.ts, might make it easier to find

leblowl (approved):
Looks good, thanks!

PR #1979, opened 2023-10-17
Fix/774

EmiM:
### Pull Request Checklist

### DIFF
--- a/packages/desktop/package-lock.json
+++ b/packages/desktop/package-lock.json
@@ -20,6 +20,7 @@
         "keymirror": "0.1.1",
         "pkijs": "^3.0.8",
         "react-qr-code": "^2.0.11",
+        "sanitize-html": "2.11.0",
         "socket.io-client": "4.6.0"
       },
       "devDependencies": {
@@ -68,6 +69,7 @@
         "@types/react-virtualized": "^9.21.13",
         "@types/redux-actions": "^2.6.1",
         "@types/redux-debounced": "^0.2.19",
+        "@types/sanitize-html": "2.9.2",
         "@types/tmp": "^0.2.3",
         "about-window": "^1.14.0",
         "chromatic": "^6.4.3",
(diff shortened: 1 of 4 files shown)


siepra (approved):

holmesworcester:
Are we using the default options for sanitize-html? Should we pass it maximally restrictive options? 

Are event handlers like onclick allowed at all now?

Are there any features that currently depend on this field being html? Is there a reason not to change it to a textarea?

holmesworcester:
>> Are event handlers like onclick allowed at all now?

> I'm not sure what you mean in this context

I mean in the default options, there are `nonBooleanAttributes` which include event dandlers like `onclick`. Does this mean those attributes are filtered out or allowed? Can someone still insert HTML with an onclick attribute? What can this do?

> @holmesworcester these are the default allowed tags: https://github.com/apostrophecms/sanitize-html#default-options I think it's enough. Technically you still can inject script via [(and probably many other ways) but it would be dangerous mostly when it'd be rendered as message received from someone.](https://github.com/TryQuiet/quiet/pull/[XSS])

I don't understand the second link. Did you intend to link to other ways?

> If you mean to change whole ContentEditable to textarea then that can be a slightly bigger task. Whatever we decide it's still a good idea to sanitize input.

If you and Lucas are sure that this is fixed we don't have to change it. If there are doubts and no features depend on this (image copy/paste?) let's change it.

> Note: Imo we should also (or first of all) add sanitizing on backend because one can omit the frontend layer.

Yes, we should do this now for a complete fix.

leblowl:
It is definitely possible to use a `textarea` and I think it would eliminate the need to sanitize HTML. For example, the text box we are typing into in Github is a `textarea`. However, `textarea` is only for plain unstyled text. In Github, typing @ mentions still opens a little autocomplete tooltip, but the actual mention is not styled until submitting the comment. So this is one approach. If you want inline styling, then I think most sites with a WYSIWYG input box use `contenteditable`. This is how Slack's input box and Gmail's compose email box both work. With `contenteditable` we have to deal with sanitizing HTML.

Perhaps it's possible also to have a hidden/invisible `textarea` with a formatted div that overlays it, however I think that quickly gets complicated since you now have to emulate all the behavior of a normal `textarea`, like a cursor (https://ckeditor.com/blog/ContentEditable-The-Good-the-Bad-and-the-Ugly/).

So there are at least two routes which I think are viable:
1. Use `textarea` and skip WYSIWYG text formatting
2. Use `contenteditable` and sanitize HTML input extremely well

@holmesworcester I think that's a product decision, but given that Slack and most other popular pieces of software support WYSIWYG, I think it's probably most approachable and familiar for the majority of users. Though a `textarea` is simpler to maintain and safer in general.

If we go with `contenteditable` it requires fewer changes to our current code base because we are already using it. We simply have to sanitize input extremely well. Doing it incorrectly is pretty bad. And so for that, I'd highly recommend following something like [OWASP guidelines](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#html-sanitization)  which recommends using the [DOMPurify](https://github.com/cure53/DOMPurify) library. It looks like TinyMCE (a commercial WYSIWYG editor) also uses DOMPurify (https://www.tiny.cloud/blog/input-sanitization/).

leblowl:
> Note: Imo we should also (or first of all) add sanitizing on backend because one can omit the frontend layer.
@EmiM I'm curious what you mean by this. Do you mean sanitizing each message as we receive it on the backend?

EmiM:
@holmesworcester 
> I don't understand the second link. Did you intend to link to other ways?

Yeah, sorry, I did not intend to link anything, github linkified it because I did not put it in codeblock. I fixed the comment already (it supposed to be `<a href="[XSS]"></a>`)

> If you and Lucas are sure that this is fixed we don't have to change it.

Lukas made some good points and I agree, this is mainly a product decision. However, as I mentioned previously, replacing contenteditable with textarea may need some extra changes and should not be done as a hotfix (in my opinion).

@leblowl 
> Do you mean sanitizing each message as we receive it on the backend?

Yes because I would not trust the client on sanitizing input and by "client" I mean "frontend layer" in our case. If you agree with me we can add that but as part of other task, as Holmes suggested.

[DOMPurify](https://github.com/cure53/DOMPurify) looks really good.

leblowl:
Also another thing to consider is that it looks like we do not actually use rich text when displaying messages, so the `contenteditable` field doesn't really serve any purpose currently. The text may look bold in the message box, but when the message is sent it only sends the plain text (https://github.com/TryQuiet/quiet/blob/d58b32c26787a04f0faf936f3e36af4c4ab61827/packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.tsx#L491)

leblowl:
@EmiM I think sanitizing on the backend for each message received becomes more important if we are actually injecting messages into the DOM (e.g. to display rich text messages) via `dangerouslySetInnerHTML`. Currently, it looks like we just take the message contents and render it via a markdown component: https://github.com/TryQuiet/quiet/blob/d58b32c26787a04f0faf936f3e36af4c4ab61827/packages/desktop/src/renderer/components/widgets/channels/TextMessage.tsx#L135 which doesn't look like it uses `dangerouslySetInnerHTML`. I think as long as we are not using `dangerouslySetInnerHTML`, we don't have to worry too much about sanitizing messages because messages won't get parsed by the browser. I think there are still a few things we have to be careful about with React: https://stackoverflow.com/a/51852579.

@holmesworcester And now that I am looking into the code more it looks like we already support markdown on `develop`. I think markdown and WYSIWYG conflict with each other. So it should probably be one or the other. If we want to support markdown, then let's just use a `textfield`.

holmesworcester:
Yes, we have no commitment to WYSIWYG. So textarea is good for now and we can revisit this when we add @ mentions.

EmiM:
So what's the final decision here? We're closing this PR and opening new one with contenteditable -> textarea?

EmiM:
I am closing this PR as we decided to replace contenteditable with textarea: https://github.com/TryQuiet/quiet/pull/1986

Issue #1981, opened 2023-10-18
Duplicate - Unregistered users mixed up


kingalg:
Desktop: Version: 2.0.1-alpha.7
Mobile: mobile@2.0.1-alpha.7 (iOS 325)

This is rather complicated issue and also I'll try to describe it as clear as possible, don't hesitate to contact me if you need some additional explanation. 

I'm going to use names of specific systems but I was able to reproduce this on all systems in different combinations several times.

Legend:
Owner - Quiet user that created the community that the rest of the users want's to join
test1 - Windows user username that joined the community when the owner was online
Android - android user 
iOS - iOS user

1. Owner creates community.
2. Windows user join with username "test1" - joining is successful and username is registered.
3. Owner goed offline.
4. Windows user "test1" send invitation code to Android and iOS users.
5. Android tries to join with username "test1".
6. Android gets the popup with information that username is already taken and choose username "test2".
7. Android join community with username "test2" and tag "Unregistered".
8. iOS user tries to join with username "test2".
9. iOS join community with username "test2".

What should happen - both Android and iOS users are able to use the application with username "test2" and tag "Duplicate".
**What happen - on Windows and Android both users (Android and iOS) are able to use the application and have an username "test2" with tag "Duplicate", however - on iOS Android user has a name "test1" (that he tried to register with initially) and tag "Duplicated" and iOS user has username "test2" and tag "Unregistered".** 

Additional info: what happen when owner goes online? Usernames are still mixed up. Android user gets a popup informing that they need to change username - this popup doesn't work and makes app crush. Upon restart the app Android user can register new username.

If we are using desktop instead of mobile it's still bugged but in different way. Let's say that we used MacOS instead of iOS and Linux instead of Android"
Linux user gets an popup that his username will be registered once the owner is online (which is bad because of 2 reasons - Owner is already online and this user is consider by our system duplicated one so they username will not be registered). MacOS user gets username "test2" and no tag, Linux user has a tag "duplicate" and with next restart of the app is asked to choose different username.

holmesworcester:
Asking users to change the username on restart is fine I think, if they change to a name that is taken and we don't catch this.

When we ask users to change a name are we checking their input against names we know are taken?

If users just restarted maybe they don't know what names are taken yet.

kingalg:
[...]

PR #2002, opened 2023-10-24
Fix/1981/Duplicate - Unregistered users mixed up - filter CSRs on replicated event

Kacper-RF:
### Pull Request Checklist

Issue:https://github.com/TryQuiet/quiet/issues/1981


siepra (approved):

EmiM (approved):

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -419,6 +423,25 @@ export class StorageService extends EventEmitter {
     this.logger('STORAGE: Finished creating certificatesRequests db')
   }
 
+  public async filterCsrs(allCsrs: string[]): Promise<string[]> {
```
leblowl, line 426: I would probably make this "getCsrs" and include
```
const allCsrs = this.getAllEventLogEntries(this.certificatesRequests)
```
inside the function, since I think the only time we want to retrieve CSRs is if they are filtered.

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -419,6 +423,25 @@ export class StorageService extends EventEmitter {
...
+    const filteredCsrsMap: Map<string, string> = new Map()
+
+    await Promise.all(
+      allCsrs.map(async csr => {
+        const parsedCsr = await loadCSR(csr)
+        const pubKey = keyFromCertificate(parsedCsr)
+
+        if (filteredCsrsMap.has(pubKey)) {
+          filteredCsrsMap.delete(pubKey)
+        }
```
leblowl, line 436: I don't think there is any need for  this, just ```filteredCsrsMap.set(pubKey, csr)```

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -419,6 +423,25 @@ export class StorageService extends EventEmitter {
...
+
+        filteredCsrsMap.set(pubKey, csr)
+      })
+    )
+
+    return [...filteredCsrsMap.values()]
```
leblowl, line 442: I think we can just do ```...filteredCsrsMap.values()```

leblowl (changes requested):
Thanks, looks good! I just have some small suggestions

packages/backend/src/nest/storage/storage.service.ts
siepra: Don't you mean `filteredCsrsMap.values()`?

packages/backend/src/nest/storage/storage.service.ts
Kacper-RF: This is quite important. 

We want to have correct order in that data structure when user didn't immediately change his name, because he can use app with duplicated name.

For example map of users pubkey=> csr(and i added nickname)

`{"pubkey1" => "csr1/owner", "pubkey2" => "csr2/alice", "pubkey3" => "csr3/alice(duplicated)", "pubkey4" => "csr4/jimmy"}`


Case that we dont delete existed key but replace it  will be this: 
Example alice(duplicated) change name after `jimmy` joined to community to `jimmy`

`{"pubkey1" => "csr1/owner", "pubkey2" => "csr2/alice", "pubkey3" => "jimmy", "pubkey4" => "csr4/jimmy(duplicated)"}`


But we want something like this:

`{"pubkey1" => "csr1/owner", "pubkey2" => "csr2/alice", "pubkey4" => "csr4/jimmy", "pubkey3" => "jimmy(duplicated)"}`

packages/backend/src/nest/storage/storage.service.ts
Kacper-RF: I want to return csrs  in that format `[val1, val2, val3...]`

leblowl: Yes, sorry, I meant `filteredCsrsMap.values()` and sorry again, I was looking at Object.values (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values) which returns an array. Map.values returns an iterator (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/values). So the initial line looks correct.

packages/backend/src/nest/storage/storage.service.ts
leblowl: Ok, got it, so the insertion order is based on when the key was originally inserted. That makes sense, nice catch!

leblowl (approved):

PR #2067, opened 2023-11-14
feature/Display "<user> has joined <community>" message when unregistered user joins[NEW]

Kacper-RF:
### Pull Request Checklist


packages/state-manager/src/sagas/communities/communities.selectors.test.ts
```diff
@@ -145,6 +145,11 @@ describe('communitiesSelectors', () => {
...
 
     const factory = await getFactory(store)
+    store.dispatch(
+      usersActions.storeUserCertificate({
```
siepra, line 149: Are you sure it's not there already? My memory is factory-girl should do it automatically, unless you set `certificate` as `null` while mocking data

packages/state-manager/src/sagas/communities/saveCommunityMetadata/saveCommunityMetadata.saga.ts
```diff
@@ -22,4 +23,12 @@ export function* saveCommunityMetadataSaga(
...
     })
   )
+
+  const community = yield* select(communitiesSelectors.currentCommunity)
+  if (!community) return
+  const isOwner = community.CA
+
+  if (!isOwner) {
+    yield* put(publicChannelsActions.sendUnregisteredInfoMessage())
+  }
```
siepra, line 33: I guess you can simplify it as just
```
const community = yield* select(communitiesSelectors.currentCommunity)
if (community?.CA) return
yield* put(publicChannelsActions.sendUnregisteredInfoMessage())
```

siepra: Although the cleanest solution would be to make a dedicated saga, hooking up on `communitiesActions.saveCommunityMetadata` or some other action (eg. replicating channels as it's required to send a message)

packages/state-manager/src/sagas/messages/verifyMessage/verifyMessages.saga.ts
```diff
@@ -1,14 +1,50 @@
...
+import { communitiesSelectors } from '../../communities/communities.selectors'
 
 export function* verifyMessagesSaga(
   action: PayloadAction<ReturnType<typeof messagesActions.incomingMessages>>['payload']
 ): Generator {
   const messages = action.payload.messages
 
   for (const message of messages) {
+    const generalChannel = yield* select(publicChannelsSelectors.generalChannel)
+    if (!generalChannel) return
```
siepra, line 17: Doesn't it stop executing for loop for channels other than 'general'? I think we do want to verify messages incoming in other channels as well

packages/state-manager/src/sagas/messages/verifyMessage/verifyMessages.saga.ts
```diff
@@ -1,14 +1,50 @@
...
+
+      while (true) {
+        ownerNickname = yield* select(communitiesSelectors.ownerNickname)
+        if (ownerNickname) {
+          break
+        }
+        yield* delay(500)
+      }
+
+      if (username !== ownerNickname) {
```
siepra, line 38: Filtering out owner's nickname should happen on sending the message (by the owner) not receiving it by other people

packages/state-manager/src/sagas/messages/verifyMessage/verifyMessages.saga.ts
```diff
@@ -1,14 +1,50 @@
...
 ): Generator {
   const messages = action.payload.messages
 
   for (const message of messages) {
+    const generalChannel = yield* select(publicChannelsSelectors.generalChannel)
+    if (!generalChannel) return
+
+    if (message.type === MessageType.Info && message.channelId === generalChannel.id) {
+      const getMessagesFromGeneralByPubKey = yield* select(
+        messagesSelectors.getMessagesFromChannelIdByPubKey(generalChannel.id, message.pubKey)
```
siepra, line 21: You can use action's payload instead of fetching all the messages from the store.
Filter the payload by type (info) and pubkey (other than owner's) and compare it against the fixed format.

siepra: This way you wuldn't even need to check whether messages are coming in general channel at all (security benefits)

packages/state-manager/src/sagas/messages/verifyMessage/verifyMessages.saga.ts
```diff
@@ -1,14 +1,50 @@
...
+      )
+
+      const allUsers = yield* select(usersSelectors.allUsers)
+
+      const username = allUsers[message.pubKey].username
```
siepra, line 26: Use a dedicated selector for this purpose

packages/state-manager/src/sagas/messages/messages.selectors.ts
```diff
@@ -45,6 +45,22 @@ export const publicChannelMessagesEntities = (address: string) =>
     return channelMessagesAdapter.getSelectors().selectEntities(channelMessagesBase.messages)
   })
 
+export const getMessagesFromChannelIdByPubKey = (channelId: string, pubkey: string) =>
```
siepra, line 48: I don't think it's actually necessary to have such selector. See my comments in `verifyMessages.saga`

packages/e2e-tests/src/tests/invitationLink.test.ts
```diff
@@ -182,12 +182,15 @@ describe('New user joins using invitation link while having app opened', () => {
 
     it('Owner sees that guest joined community', async () => {
       console.log('Invitation Link', 21)
-      const generalChannel = new Channel(ownerApp.driver, 'general')
+      const generalChannel = new Channel(guestApp.driver, 'general')
```
EmiM, line 185: Why 'guestApp' driver? This step should check if owner sees that guest joined community. If the logic changed I'd also change test step name

packages/state-manager/src/sagas/communities/communities.selectors.test.ts
Kacper-RF: Without this I get undefined

packages/state-manager/src/sagas/communities/saveCommunityMetadata/saveCommunityMetadata.saga.ts
Kacper-RF: I really like your solution, it's very clean, only 3 lines of code

packages/common/src/messages.test.ts
```diff
@@ -0,0 +1,50 @@
+import { PublicChannelStorage } from '@quiet/types'
+import { generateChannelId } from './channelAddress'
+import { userCreatedChannelMessage, userJoinedMessage, verifyUserInfoMessage } from './messages'
+
+describe('messages helper', () => {
+  const username = 'johnny'
+
+  const generalChannel: PublicChannelStorage = {
```
siepra, line 8: Ideally you'd use factory-girl for mocking channel objects. It's minor though.
```
const generalChannel = (await factory.build<typeof publicChannels.actions.addChannel>('PublicChannel', {
      id: generateChannelId('general'),
      name: 'general',
      owner: username
    })).payload
```

packages/state-manager/src/sagas/messages/verifyMessage/verifyMessages.saga.ts
```diff
@@ -1,14 +1,49 @@
...
+        yield* delay(500)
+      }
+      const channel = yield* select(publicChannelsSelectors.getChannelById(message.channelId))
+      if (!channel) return
+
+      const expectedMessage = yield* call(verifyUserInfoMessage, user.username, channel)
+
+      if (message.message !== expectedMessage) {
+        console.error(`${user.username} tried to send a malicious info message`)
+        return
```
siepra, line 43: I'm not sure if you should use `return` or rather change the `isVerified` prop of the returned status object. Lacking verification status for a message may have consequences in the other parts of state-manager.

packages/state-manager/src/sagas/messages/verifyMessage/verifyMessages.saga.ts
```diff
@@ -1,14 +1,49 @@
...
+import { usersSelectors } from '../../users/users.selectors'
+import { verifyUserInfoMessage } from '@quiet/common'
 
 export function* verifyMessagesSaga(
   action: PayloadAction<ReturnType<typeof messagesActions.incomingMessages>>['payload']
 ): Generator {
   const messages = action.payload.messages
 
   for (const message of messages) {
+    let ownerData = yield* select(usersSelectors.ownerData)
```
siepra, line 16: Code starting from line 16 to 24 should be moved out of the `for` loop as it doesn't depend on the message's data and makes no sense to repeat on each iteration

packages/state-manager/src/sagas/messages/verifyMessage/verifyMessages.saga.test.ts
```diff
@@ -0,0 +1,300 @@
...
+          timestamp: DateTime.utc().valueOf(),
+          owner: alice.nickname,
+          id: generateChannelId('sport'),
+        },
+      })
+    ).channel
+  })
+
+  it('verify standard message ', async () => {
+    if (!alice.userCsr?.userCsr) throw Error('no Alice userCsr')
```
siepra, line 82: Funny thing to do in test. Verifies mocked data validity. What does it result with when running with no csr generated? Will it mark the test as failed?

packages/state-manager/src/sagas/messages/verifyMessage/verifyMessages.saga.test.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    const message: ChannelMessage = {
```
siepra, line 84: Wonder why didn't you use factory-girl to mock `message` object as you did for users and channels?

packages/state-manager/src/sagas/messages/messages.selectors.test.ts
```diff
@@ -47,6 +52,18 @@ describe('messagesSelectors', () => {
       id: community.id,
       nickname: 'john',
     })
+
+    devChannel = (
```
siepra, line 56: Is it a leftover?

packages/state-manager/src/sagas/communities/saveCommunityMetadata/saveCommunityMetadata.saga.ts
```diff
@@ -22,4 +23,8 @@ export function* saveCommunityMetadataSaga(
       ownerCertificate: action.payload.ownerCertificate,
     })
   )
+
+  const community = yield* select(communitiesSelectors.currentCommunity)
```
siepra, line 27: This is way better, although to avoid confusion I'd have rather emit `sendUnregisteredInfoMessage` action directly in `startConnection.saga.ts:261` and moved checking the presence of community.CA inside `sendUnregisteredInfoMessageSaga` (it wouldn't be my first shot to look for sending message from within saveCommunityMetadataSaga)

packages/state-manager/src/sagas/publicChannels/createGeneralChannel/sendInitialChannelMessage.saga.ts
```diff
@@ -21,7 +22,7 @@ export function* sendInitialChannelMessageSaga(
   const message =
     pendingGeneralChannelRecreation && isGeneral
       ? `@${user?.nickname} deleted all messages in #general`
```
siepra, line 24: Don't we want to provide a pattern for this kind of message in the same manner?

packages/state-manager/src/sagas/publicChannels/sendUnregisteredInfoMessage/sendUnregisteredInfoMessage.saga.ts
```diff
@@ -0,0 +1,25 @@
...
+import { communitiesSelectors } from '../../communities/communities.selectors'
+import { identitySelectors } from '../../identity/identity.selectors'
+import { userJoinedMessage } from '@quiet/common'
+
+export function* sendUnregisteredInfoMessage(): Generator {
+  const community = yield* select(communitiesSelectors.currentCommunity)
+  const identity = yield* select(identitySelectors.currentIdentity)
+  const generalChannel = yield* select(publicChannelsSelectors.generalChannel)
+
+  if (!community?.name || !identity || !generalChannel) return
```
siepra, line 14: Don't we risk not sending this 'welcome' message at all, depending on the race condition of replicating channels data and entering saga? I guess what we need is to wait for actual general channel replciation before entering the saga rather than exiting in case of lack of general channel data in store

packages/state-manager/src/sagas/communities/saveCommunityMetadata/saveCommunityMetadata.saga.ts
siepra: Actually, we must depend on channels rather than community metadata replication event

packages/state-manager/src/sagas/publicChannels/sendUnregisteredInfoMessage/sendUnregisteredInfoMessage.saga.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function* sendUnregisteredInfoMessage(): Generator {
+  const community = yield* select(communitiesSelectors.currentCommunity)
+  if (community?.CA) return
+
+  const identity = yield* select(identitySelectors.currentIdentity)
+  let generalChannel = yield* select(publicChannelsSelectors.generalChannel)
+
+  if (!generalChannel) {
+    yield* take(publicChannelsActions.channelsReplicated)
```
siepra, line 18: How about you not only checked for the occurence of an action but also if the payload contains general channel? Then you could remove the condiiton in line 23

siepra (approved):

PR #2072, opened 2023-11-15
feat: add Certificates store and validation

siepra:
### Pull Request Checklist


packages/backend/src/nest/registration/registration.validators.ts
```diff
@@ -3,6 +3,7 @@ import { registerDecorator, ValidationArguments, ValidationOptions } from 'class
 import Logger from '../common/logger'
 
 const logger = Logger('registration.validators')
+
 export function IsCsr(validationOptions?: ValidationOptions) {
```
siepra, line 7: Do I get it right, that we want to introduce the same method for certificate? Also, what are we particularly interested in validating inside `CertificateContainsFields`?

packages/backend/src/nest/storage/certificates/certificates.store.ts
```diff
@@ -0,0 +1,138 @@
...
+            logger('Saved certificate locally')
+
+            emitter.emit(StorageEvents.LOAD_CERTIFICATES, {
+                certificates: await this.getCertificates(),
+            })
+
+            // await this.updatePeersList()
+        })
+
+        this.store.events.on('ready', async () => {
```
siepra, line 54: Can we get rid of this event as it probably duplicates what's being done on `db.load`?

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -313,94 +309,88 @@ export class StorageService extends EventEmitter {
...
-
-  public async createDbForCertificates() {
-    this.logger('createDbForCertificates init')
-    this.certificates = await this.orbitDb.log<string>('certificates', {
-      replicate: false,
-      accessController: {
-        write: ['*'],
-      },
-    })
-    this.certificates.events.on('replicate.progress', async (_address, _hash, entry, _progress, _total) => {
```
siepra, line 359: Do we need to hook up on `replicate.progress`? I'd heard it's not reliable anyway and we process data on other events anyway

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -313,94 +309,88 @@ export class StorageService extends EventEmitter {
...
-
-      const username = getCertFieldValue(parsedCertificate, CertFieldsTypes.nickName)
-      if (!username) {
-        this.logger.error(
-          `Certificates replicate.progress: could not parse certificate for field type ${CertFieldsTypes.nickName}`
-        )
-        return
-      }
-
-      this.userNamesMap.set(key, username)
```
siepra, line 373: Should it be moved to certificate's store? What does it do anyway?

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -313,94 +309,88 @@ export class StorageService extends EventEmitter {
...
-        write: ['*'],
-      },
-    })
-    this.certificates.events.on('replicate.progress', async (_address, _hash, entry, _progress, _total) => {
-      const certificate = entry.payload.value
-
-      const parsedCertificate = parseCertificate(certificate)
-      const key = keyFromCertificate(parsedCertificate)
-
-      const username = getCertFieldValue(parsedCertificate, CertFieldsTypes.nickName)
```
siepra, line 365: How does it relate to the certificates validation we're going to perform in certificate's store? Is it still something we need or can we ignore it while moving this part of code?

packages/backend/src/nest/storage/storage.service.ts
EmiM: So I don't remember about it not being reliable but I think listening on 'replicated' requires changing code a bit because replicate.progress gave you the information about certain entry.
If we **don't have to** change it right now I would do it in next iteration because the change itself can introduce regressions.

packages/backend/src/nest/storage/storage.service.ts
EmiM: I don't remember exactly but it was probably for faster access of usernames. Instead of parsing all certs every time we keep map of pKey-username in memory.

packages/backend/src/nest/storage/certificates/certificates.store.ts
EmiM: Yeah, if we do the same on 'ready' and on other event that calls "db.load" then yes, there is a duplication because db.load triggers 'ready'

packages/backend/src/nest/storage/storage.service.ts
siepra: There was a need to move caching usernames anyway so I removed hooking up on `replicate.progress` in-the-process

packages/backend/src/nest/storage/storage.service.ts
siepra: Yes, also it's being used for push notifications (only). I moved it to certificate's store

packages/backend/src/nest/storage/storage.service.ts
siepra: See the comments above

packages/backend/src/nest/storage/certificates/certificates.store.ts
siepra: Ok, now it emits `emitter.emit(SocketActionTypes.CONNECTION_PROCESS_INFO, ConnectionProcessInfo.LOADED_CERTIFICATES)` but doesn't trigger certificates loading

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -314,92 +309,41 @@ export class StorageService extends EventEmitter {
...
 
-  public async updatePeersList() {
-    const allUsers = this.getAllUsers()
-    const registeredUsers = this.getAllRegisteredUsers()
-    const peers = [...new Set(await getUsersAddresses(allUsers.concat(registeredUsers)))]
-    console.log('updatePeersList, peers count:', peers.length)
-    const community = await this.localDbService.get(LocalDBKeys.COMMUNITY)
-    this.emit(StorageEvents.UPDATE_PEERS_LIST, { communityId: community.id, peerList: peers })
-  }
+  public async updatePeersList() {}
```
siepra, line 333: We no longer need a list of certificates to update peers list (we use csrs instead) so to make things simpler for merging with https://github.com/TryQuiet/quiet/pull/2071 I removed certain methods so we can just accept incoming changes when resolving conflicts

siepra:
At this stage, I've moved logic related to certificates out of storage service into their own store.
Next steps are:
1. Merge https://github.com/TryQuiet/quiet/pull/2071 into the branch
2. Write tests for `certificates.store`
3. Provide certificates validation

siepra:
At this step automatic tests are passing and the app seemed to work fine when tested manually.

packages/backend/src/nest/storage/storage.service.spec.ts
```diff
@@ -446,14 +454,13 @@ describe('StorageService', () => {
       }
     )
 
-    it('Certificates and peers list are updated on write event', async () => {
+    it.skip('Certificates and peers list are updated on write event', async () => {
```
siepra, line 457: This test is failing due to lack of a good way of injecting dependencies in tests using nest (certificates storage is now separate)

EmiM (approved):
Please make sure that e2e tests pass on at least one desktop OS

PR #2157, opened 2023-12-08
Fixing user registration concurrency issues

leblowl:
### Pull Request Checklist

### DIFF
--- a/packages/backend/src/nest/connections-manager/connections-manager.service.ts
+++ b/packages/backend/src/nest/connections-manager/connections-manager.service.ts
@@ -318,11 +318,25 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
     await this.launchCommunity(payload)
     this.logger(`Created and launched community ${payload.id}`)
     this.serverIoProvider.io.emit(SocketActionTypes.NEW_COMMUNITY, { id: payload.id })
+
+    if (!payload.ownerCsr || !payload.permsData) {
+      throw new Error("Owner CSR and PermsData required to create community!")
+    }
+    // In launchCommunity, we initialize the StorageService and also
+    // the RegistrationService for the first time. And then now, once
+    // those are ready, we register the owner's certificate.
+    const cert = await this.registrationService.registerOwnerCertificate({ communityId: payload.id, userCsr: payload.ownerCsr, permsData: payload.permsData })
+    if (payload.certs) {
+      // Hacking, perhaps make certs.certificate optional
+      payload.certs.certificate = cert || ''
+    }
+    await this.localDbService.put(LocalDBKeys.COMMUNITY, payload)
   }
 
   public async launchCommunity(payload: InitCommunityPayload) {
     this.logger('Launching community: peers:', payload.peers)
     this.communityState = ServiceState.LAUNCHING
+
     // Perhaps we should call this data something else, since we already have a Community type.
     // It seems like InitCommunityPayload is a mix of various connection metadata.
     const communityData: InitCommunityPayload = await this.localDbService.get(LocalDBKeys.COMMUNITY)
@@ -424,6 +438,7 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
       this.serverIoProvider.io.emit(SocketActionTypes.PEER_DISCONNECTED, payload)
     })
     await this.storageService.init(_peerId)
+    await this.registrationService.init(this.storageService)
     this.logger('storage initialized')
   }
 
@@ -444,10 +459,6 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
     this.registrationService.on(RegistrationEvents.ERROR, payload => {
       emitError(this.serverIoProvider.io, payload)
     })
-    this.registrationService.on(RegistrationEvents.NEW_USER, async payload => {
-      await this.storageService?.saveCertificate(payload)
-    })
-
     this.registrationService.on(RegistrationEvents.FINISHED_ISSUING_CERTIFICATES_FOR_ID, payload => {
       if (payload.id) {
         this.storageService.resolveCsrReplicatedPromise(payload.id)
--- a/packages/backend/src/nest/registration/registration.service.spec.ts
+++ b/packages/backend/src/nest/registration/registration.service.spec.ts
@@ -10,6 +10,11 @@ import { issueCertificate, extractPendingCsrs } from './registration.functions'
 import { jest } from '@jest/globals'
 import { createTmpDir } from '../common/utils'
 import { RegistrationEvents } from './registration.types'
+import { EventEmitter } from 'events'
+import { CertificatesStore } from '../storage/certificates/certificates.store'
+import { create, IPFS } from 'ipfs-core'
+import OrbitDB from 'orbit-db'
+import { TestConfig } from '../const'
 
 describe('RegistrationService', () => {
   let module: TestingModule
@@ -179,5 +184,72 @@ describe('RegistrationService', () => {
     })
 
     expect(eventSpy).toHaveBeenCalledTimes(3)
+  })
+
+  const createOrbitDbInstance = async () => {
+    const ipfs: IPFS = await create()
+    // @ts-ignore
+    const orbitdb = await OrbitDB.createInstance(ipfs, {
+      directory: TestConfig.ORBIT_DB_DIR,
+    })
+
+    return { orbitdb, ipfs }
+  }
+
+  it('race', async () => {
+    let { orbitdb, ipfs } = await createOrbitDbInstance()
+    let store = new CertificatesStore(orbitdb)
+    let emitter = new EventEmitter()
+    await store.init(emitter)
+
+    registrationService.permsData = permsData
+    // @ts-ignore
+    registrationService.storageService = { certificatesStore: store }
+
+    registrationService.onModuleInit()
+
+    const userCsr = await createUserCsr({
+      nickname: 'alice',
+      commonName: 'nqnw4kc4c77fb47lk52m5l57h4tcxceo7ymxekfn7yh5m66t4jv2olad.onion',
+      peerId: 'Qmf3ySkYqLET9xtAtDzvAr5Pp3egK1H3C5iJAZm1SpLEp6',
+      dmPublicKey: 'testdmPublicKey',
+      signAlg: configCrypto.signAlg,
+      hashAlg: configCrypto.hashAlg,
+    })
+
+    const userCsr2 = await createUserCsr({
+      nickname: 'alice',
+      commonName: 'nnnnnnc4c77fb47lk52m5l57h4tcxceo7ymxekfn7yh5m66t4jv2olad.onion',
+      peerId: 'QmffffffqLET9xtAtDzvAr5Pp3egK1H3C5iJAZm1SpLEp6',
+      dmPublicKey: 'testdmPublicKey',
+      signAlg: configCrypto.signAlg,
+      hashAlg: configCrypto.hashAlg,
+    })
+
+    const certificates: string[] = []
+    let calls = 0
+
+    registrationService.on('new', (p: any) => {
+      console.log(p)
+      calls++
+    })
+
+    registrationService.emit(
+      RegistrationEvents.REGISTER_USER_CERTIFICATE,
+      { csrs: [userCsr.userCsr], certificates: [], id: '1' }
+    )
+
+    registrationService.emit(
+      RegistrationEvents.REGISTER_USER_CERTIFICATE,
+      { csrs: [userCsr2.userCsr], certificates: [], id: '2' }
+    )
+
+    await new Promise(r => setTimeout(r, 10000))
+
+    await store.close()
+    await orbitdb.stop()
+    await ipfs.stop()
+
+    expect(calls).toEqual(1)
   })
 })
--- a/packages/backend/src/nest/registration/registration.service.ts
+++ b/packages/backend/src/nest/registration/registration.service.ts
@@ -99,20 +99,17 @@ export class RegistrationService extends EventEmitter implements OnModuleInit {
 
     const pendingCsrs = await extractPendingCsrs({ csrs: payload.csrs, certificates: certificates as string[] })
 
-    await Promise.all(
-      pendingCsrs.map(async csr => {
-        const result = await issueCertificate(csr, this._permsData)
-        if (result?.cert) {
-          await this.storageService.certificatesStore.addCertificate(result.cert)
-          // Not sure if this is necessary
-          const certs = await this.storageService.certificatesStore.loadAllCertificates()
-          if (!certs.includes(result.cert)) {
-            throw new Error("Cert wasn't added to CertificateStore correctly")
-          }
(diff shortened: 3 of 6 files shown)


packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -72,10 +67,50 @@ export class RegistrationService extends EventEmitter implements OnModuleInit {
...
+      throw new Error("Storage Service must be initialized before the Registration Service")
     }
+
+    // Lack of permsData means that we are not the owner of the
+    // community in the official model of the app, however anyone can
+    // modify the source code, put malicious permsData here, issue
+    // false certificates and try to trick other users. To prevent
+    // that, peers verify that anything that is written to the
+    // certificate store is signed by the owner.
+    if (!this._permsData) {
```
holmesworcester, line 93: Why is this here? Why would we trigger a "finished issuing certificates" event if we are not the owner?

packages/state-manager/src/sagas/identity/savedOwnerCertificate/savedOwnerCertificate.saga.ts
```diff
@@ -10,26 +10,5 @@ export function* savedOwnerCertificateSaga(
...
   action: PayloadAction<ReturnType<typeof identityActions.savedOwnerCertificate>['payload']>
 ): Generator {
-  let communityId: string = action.payload
-
-  if (!communityId) {
-    communityId = yield* select(communitiesSelectors.currentCommunityId)
-  }
-
-  const community = yield* select(communitiesSelectors.selectById(communityId))
-  const identity = yield* select(identitySelectors.selectById(communityId))
```
holmesworcester, line 20: this is moved to registerCertificateSaga for what reason? i don't see any problem but i'm curious!

holmesworcester (commented):
Just two questions...

packages/state-manager/src/sagas/identity/savedOwnerCertificate/savedOwnerCertificate.saga.ts
leblowl: Yea, so we can create the community (setup the storage service and the registration service) and then register the owner's certificate.

packages/backend/src/nest/registration/registration.service.ts
siepra: This statement basically says "You're not the owner? Do not continue issues certificates

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -424,6 +438,7 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
       this.serverIoProvider.io.emit(SocketActionTypes.PEER_DISCONNECTED, payload)
     })
     await this.storageService.init(_peerId)
+    await this.registrationService.init(this.storageService)
```
Kacper-RF, line 441: This is not a recommended way to inject dependencies.

Please add StorageModule in Registration Module:
`imports: [StorageModule]`
And in RegistartionService add:
`private readonly storageService: StorageService`

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -318,11 +318,25 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
     await this.launchCommunity(payload)
     this.logger(`Created and launched community ${payload.id}`)
     this.serverIoProvider.io.emit(SocketActionTypes.NEW_COMMUNITY, { id: payload.id })
+
+    if (!payload.ownerCsr || !payload.permsData) {
+      throw new Error("Owner CSR and PermsData required to create community!")
```
siepra, line 323: I think you're missing `return` here

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -318,11 +318,25 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
...
+    }
+    // In launchCommunity, we initialize the StorageService and also
+    // the RegistrationService for the first time. And then now, once
+    // those are ready, we register the owner's certificate.
+    const cert = await this.registrationService.registerOwnerCertificate({ communityId: payload.id, userCsr: payload.ownerCsr, permsData: payload.permsData })
+    if (payload.certs) {
+      // Hacking, perhaps make certs.certificate optional
+      payload.certs.certificate = cert || ''
```
siepra, line 331: What is it for?

Kacper-RF:
Have you been able to reproduce this issue on your machine? 
If so, do you have exact steps to do this? (For me it's very random)

siepra (commented):
Could you explain how exactly those changes solves the issue?

packages/backend/src/nest/connections-manager/connections-manager.service.ts
EmiM, line 323: This would just show "abnormal backend termination" js error in application. Maybe it should be a `return` instead and emitError?

packages/backend/src/nest/registration/registration.service.ts
EmiM: The name is misleading. AFAIK it's only used for resolving promise in csrReplicatedPromiseMap in storage.

Kacper-RF:
And one very important thing - e2e tests should pass successfully on at least one system before merge.

leblowl:
> Could you explain how exactly those changes solves the issue?

Hi Wiktor, I hope you're doing well my friend. I can. I wrote a little comment in the code, I can quote it here.

>// Apparently, JS will run each function to completion. So assuming
  // we do not have multiple threads, this function should run to
  // completion before it is called again. Because we take the CSRs
  // and then get the latest view of certificates, filter CSRs and
  // then update certificates in a single function, we should never
  // issue a CSR that is contained in the certificates list, at least
  // in this function... as long as there is no other code that
  // updates the certificates list. Something else to consider is that
  // because this function is called asynchronously, there may be
  // several invocations with different CSRs and they may run in an
  // unexpected order. We could address that if it's an issue, but I
  // think that might only affect the order of CSR registration.

I think it might be interested to discuss how race conditions can happen in Javascript. Do you have any thoughts?

leblowl:
I've learned a bit more about how JS works today and I think there is more to it. I don't think it will necessarily run `registerUserCertificates` all the way through synchronously. But in general, I think we can reason about the code we write and if we absolutely want something to run synchronously in a specific order, we can do that.

Kacper-RF:
I checked the e2e multiple test locally from your branch and I think something is broken:
image

packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -18,32 +22,15 @@ export class RegistrationService extends EventEmitter implements OnModuleInit {
   onModuleInit() {
     this.on(
       RegistrationEvents.REGISTER_USER_CERTIFICATE,
-      async (payload: { csrs: string[]; certificates: string[]; id: string }) => {
-        await this.issueCertificates(payload)
+      (payload: { csrs: string[]; certificates: string[]; id: string }) => {
+        this.registrationEvents.push({ csrs: payload.csrs, id: payload.id })
+        this.tryRegisterNextUserCertificates()
```
kowalski, line 27: tryRegisterNextUserCertificates() is async function so it needs to be `await`ed. Otherwise it's a fire-and-forget, which is a very bad idea

I think something may be broken again with our eslint. Do you not get an error on this line? It should complain about missing await. If you *did* want to actually fire-and-forget you should write it as 
```
void this.tryRegisterNextUserCertificates()
```
but this is asually a very bad idea and rarely the behaviour you want

packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -18,32 +22,15 @@ export class RegistrationService extends EventEmitter implements OnModuleInit {
   onModuleInit() {
     this.on(
       RegistrationEvents.REGISTER_USER_CERTIFICATE,
-      async (payload: { csrs: string[]; certificates: string[]; id: string }) => {
-        await this.issueCertificates(payload)
+      (payload: { csrs: string[]; certificates: string[]; id: string }) => {
```
kowalski, line 25: this should still be async function

packages/backend/src/nest/registration/registration.service.spec.ts
```diff
@@ -180,4 +185,71 @@ describe('RegistrationService', () => {
...
+    let { orbitdb, ipfs } = await createOrbitDbInstance()
+    let store = new CertificatesStore(orbitdb)
+    let emitter = new EventEmitter()
+    await store.init(emitter)
+
+    registrationService.permsData = permsData
+    // @ts-ignore
+    registrationService.storageService = { certificatesStore: store }
+
+    registrationService.onModuleInit()
```
kowalski, line 209: missing await

packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -72,10 +70,76 @@ export class RegistrationService extends EventEmitter implements OnModuleInit {
...
+
+      if (this.registrationEvents.length !== 0) {
+        setTimeout(this.tryRegisterNextUserCertificates.bind(this), 0)
+      }
     }
   }
+
+  // Apparently, JS will run each function to completion. So assuming
+  // we do not have multiple threads, this function should run to
+  // completion before it is called again. Because we take the CSRs
```
kowalski, line 91: this is inaccurate

JS will run the function synchronously until the first `await` statement. 
It's a cool excercise to look up the actual implementation of this in C-node. 
What happens underneath is that `await` is actually almost synonymous to `yield`. Async functions are really generators that "generate" Promises. Whenever Promise is returned, the handler applies to it's then() callback the logic that effectively calls `next(resultOfPreviousPromise)` on the generator. 
If you think about it in these terms it becomes obvious what will and what will not get interrupted

packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -72,10 +70,76 @@ export class RegistrationService extends EventEmitter implements OnModuleInit {
...
+  // updates the certificates list. Something else to consider is that
+  // because this function is called asynchronously, there may be
+  // several invocations with different CSRs and they may run in an
+  // unexpected order. We could address that if it's an issue, but I
+  // think that might only affect the order of CSR registration.
+  public async registerUserCertificates(payload: { csrs: string[]; id?: string }) {
+    if (!this.storageService) {
+      throw new Error("Storage Service must be initialized before the Registration Service")
+    }
+    console.log("Registering user certificates")
```
kowalski, line 105: didn't we refactor everthing to use nest logger for logging? If so, please fix this

packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -72,10 +70,76 @@ export class RegistrationService extends EventEmitter implements OnModuleInit {
...
+
+    // Lack of permsData means that we are not the owner of the
+    // community in the official model of the app, however anyone can
+    // modify the source code, put malicious permsData here, issue
+    // false certificates and try to trick other users. To prevent
+    // that, peers verify that anything that is written to the
+    // certificate store is signed by the owner.
+    if (!this._permsData) {
+      if (payload.id) this.emit(RegistrationEvents.FINISHED_ISSUING_CERTIFICATES_FOR_ID, { id: payload.id })
```
kowalski, line 114: I'm starting to think eslint is completely off in this project. Is it not requesting that you wrap this in brackets?

packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -53,15 +40,26 @@ export class RegistrationService extends EventEmitter implements OnModuleInit {
     }
   }
 
-  public async registerOwnerCertificate(payload: RegisterOwnerCertificatePayload): Promise<void> {
+  public async registerOwnerCertificate(payload: RegisterOwnerCertificatePayload): Promise<string | undefined> {
```
kowalski, line 43: `string | undefined` is annoying type to work with

For functions that return stuff conditionally it's generally recommended to go with `string | null`

packages/state-manager/src/sagas/identity/registerCertificate/registerCertificate.saga.ts
```diff
@@ -22,16 +24,30 @@ export function* registerCertificateSaga(
...
+      return
+    }
+
+    const payload: InitCommunityPayload = {
+      id: currentCommunity.id,
+      peerId: identity.peerId,
+      hiddenService: identity.hiddenService,
+      certs: {
+        // Hacking, perhaps make certs.certificate optional
+        certificate: identity.userCertificate || '',
```
kowalski, line 39: call it bikeshedding on my part but `||` is a logic operator, so there is problem with how it's typed
I highly recommend using `??` instead

packages/backend/src/nest/registration/registration.service.ts
leblowl: I think

```
new Promise(something).then(another),
```

is equivalent to 

```
test = async () => {
  await something
  another
}
```

and where the Promise body, `something` is entirely synchronous. I think it's only the promise callback, `another`, which is added to the event loop and is asynchronous. So if there is no line after an `await`, I don't think the `await` does anything. I don't think the `async` keyword actually makes the function async. It's still synchronous unless there is a promise callback or I/O callback or timer. I've been reading a bit more about how JS works internally and posted a lot of links in the Slack internal channel. I also wrote some interested tests over here: https://github.com/TryQuiet/quiet/compare/develop...race

I was confused myself and so I started diving into it a bit more. I don't have the whole picture and am likely still confused about many things, but this is my latest understanding

packages/backend/src/nest/registration/registration.service.ts
leblowl: Yea, this comment is outdated

leblowl:
@siepra 

> How does it differ from the `csrReplicatedPromiseMap` that we already use in `certificatesRequestsStore` which holds execution of the `issueCertificates` method of `registration.service` until its iteration is finished?

I'm not really sure, I'd love to walk through the code with you so we can have a better understanding

packages/backend/src/nest/registration/registration.service.ts
kowalski: > So if there is no line after an await, I don't think the await does anything. I don't think the async keyword actually makes the function async

It's actually the opposite.

If you don't `await` this means that that the outer Promise will fire immediately. If you `await` the inner Promise the outer promise will wait for the inner one.
Even more importantly, if inner Promise throws a failure:
a) with await the outer Promise will be a failuer
b) without await the outer Promise will succeed and the failure from inner Promise will hit the generic error handler of the process

This second point is the main reason for always awaiting stuff. Without it, error handling is a nightmare.

siepra:
@leblowl 

> I don't think it really matters what awaits you have in there, they only affect the order of operations inside the function

With this I disagree, but:

> I also think it might be worth it to refactor a bit like in this PR where we move the synchronization complexity closer to where it matters

Yes, I think I even like your refactor better as it seems more clean and self-explanatory but my point was to raise a question whether it affects the issue we're trying to fix here. I think it doesn't cause it just reinvents something that's been there already.

leblowl:
@siepra I see, thanks for catching that!

PR #2190, opened 2024-01-02
Read messages from a file and add to OrbitDB

leblowl:
@### Pull Request Checklist


packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -791,6 +803,26 @@ export class StorageService extends EventEmitter {
...
+    for (const entry of entries) {
+      const channelId = entry.payload.value.channelId
+      const repo = this.publicChannelsRepos.get(channelId)
+
+      if (repo) {
+        this.logger('Found channel, syncing entries')
+        // @ts-ignore
+        await repo.db.sync([entry])
+      } else {
+        this.logger.error(`Could not add entries. No '${channelId}' channel in saved public channels`)
```
holmesworcester, line 821: So this means that if someone syncs a message entry before they've synced the channel creation orbitdb entry they will get this error. Is that right? 

Can this happen in Quiet as it exists now? Will it fix itself as the user downloads the orbitdb entry that creates the channel?

What about this approach to adding entries to Orbitdb? Is it similarly robust to the channel creation entry arriving after the message? 

Is there a way we can make it robust to that?

leblowl: > So this means that if someone syncs a message entry before they've synced the channel creation orbitdb entry they will get this error. Is that right?
This is really just a demonstration at this point, so the error messaging is mostly for myself, but I think in general the person needs the channel list to be able to add messages to channels.

> Can this happen in Quiet as it exists now? Will it fix itself as the user downloads the orbitdb entry that creates the channel?
I don't think this would happen currently because they need the channel to subscribe to the channel, it might only happen when receiving messages out-of-band.

> What about this approach to adding entries to Orbitdb? Is it similarly robust to the channel creation entry arriving after the message?
This demonstration is not robust to that, but I think it's quite possible to make it... just save messages if the channel doesn't exist and retry when the channel does exist.

holmesworcester: > I don't think this would happen currently because they need the channel to subscribe to the channel, it might only happen when receiving messages out-of-band.

Interesting! So you can potentially receive data for a DB you don't even know about yet. 

We talked about containers and notification scopes--different identifiers for data on the relay that would tell the person fetching it how to treat it. This seems related to that. I wonder if there's anything there. 

Like, with OrbitDB at some point we could decide to simply not subscribe to channels we have not joined or don't care about, for example for performance reasons in a big community. Would we want the ability to ignore certain data on the data relay as well?

Or in the other direction, if our conceptual model for Quiet is currently "there is one state for each community and all community members' clients replicate that state" would it be better/safer/easier for the way we're storing data to reflect that and just be one big DB/DAG? 

Probably we do not need to settle this now but I wanted to mention it because it stands out to me here. (Is it clear what I mean?)

mcginty:
Closing outdated PRs, please re-open if this was a mistake.

Issue #2155, opened 2023-12-08
Community owner will register the same username twice when it should not


kingalg:
Version: 2.0.3-alpha.13
System: MacOS, Windows, Linux
So far issue not spotted on mobile.

**Issue**: more then one user registered with the same username.
It was working before alpha.13. The last time we had a bug related to impersonation attack was around 5 weeks ago and it was fixed.

**Steps to recreate** (this is how I did it, but Wiktor and Kacper replicated it in different ways as well):
1. Create community and register owner on desktop.
2. Make sure that the owner is turn on.
3. Join with another desktop and the same username as the owner. 

**What happens** - both users are registered with the same username and channel users get "possible impersonation attack" popup.

siepra:
HYPOTHESIS:

`certificatesRequestsStore.ts:71` emits `LOADED_USER_CSRS` on database load, which then triggers `REPLICATED_CSR` in `storage.service.ts:334`. It accepts an array of certificates as a payload field, which is being loaded from the certificates store. 

Method `extractPendingCsrs` should filter out CSRs containing usernames already used in registered certificates, but if it received an empty list of certificates, then it won't filter out anything.

My hypothesis is there's a race condition when the certificates store is being loaded AFTER the CSRs store and as a result it isn't providing certificates to verify username uniqueness against.

holmesworcester:
Lucas added this PR: https://github.com/TryQuiet/quiet/pull/2157

holmesworcester:
This is a heisenbug but the warning works correctly (so there isn't a security issue, we do not think) and it happens only in unlikely cases. Moved to next sprint.

holmesworcester:
Noting that we think this is fixed. We should wait to see if it comes up again and reopen. 

@kingalg please reopen if you're still seeing this in the latest release.

PR #2212, opened 2024-01-10
Fix duplicate username registration v2

leblowl:
Another attempt at refactoring the registration service so that we can more easily verify that it does not register duplicate certs. Related to #2155 

### Pull Request Checklist


packages/backend/src/nest/storage/certificates/certificates.store.ts
```diff
@@ -97,6 +92,11 @@ export class CertificatesStore {
   public updateMetadata(metadata: CommunityMetadata) {
     if (!metadata) return
     this.metadata = metadata
+    // Community metadata is required for validating certificates, so
+    // we re-validate certificates once community metadata is set. It
+    // might also make sense to initialize CertificateStore after
+    // CommunityMetadata is received instead of doing this.
+    this.loadedCertificates()
```
leblowl: The previous wait mechanism with setTimeout appeared to be causing another JS timing issue where callbacks were run out of order and caused some certificates to not be sent to the frontend. The other changes in this PR seemed to flush the bug out into the open. So this change is fixing that. Now instead of waiting, we throw an error if community metadata is missing and revalidate when community metadata is available.

packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -4,35 +4,66 @@ import { extractPendingCsrs, issueCertificate } from './registration.functions'
...
+        this.logger('Issuing certificates', event)
+        this.registrationEventInProgress = true
+        await this.issueCertificates({
+          ...event,
+          certificates: (await this.storageService?.certificatesStore.loadAllCertificates()) as string[],
+        })
       }
-    )
+    }
   }
```
leblowl, line 46: Process only one event at a time (`registrationEventInProgress`).

packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -75,10 +99,11 @@ export class RegistrationService extends EventEmitter implements OnModuleInit {
...
 
   public async registerUserCertificate(csr: string): Promise<void> {
-    const result = await issueCertificate(csr, this._permsData)
+    const result = await issueCertificate(csr, this.permsData)
     this.logger('DuplicatedCertBug', { result })
     if (result?.cert) {
-      this.emit(RegistrationEvents.NEW_USER, { certificate: result.cert })
+      // @ts-ignore
+      await this.storageService?.saveCertificate({ certificate: result.cert })
```
leblowl, line 106: We only process the next certificate after we have saved the current batch of certificates and we know we have saved them because we are awaiting it here rather than sending events which are not awaited.

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -618,15 +613,12 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
...
-        this.serverIoProvider.io.emit(SocketActionTypes.RESPONSE_GET_CSRS, { csrs: payload.csrs })
-        this.registrationService.emit(RegistrationEvents.REGISTER_USER_CERTIFICATE, payload)
-      }
-    )
+    this.storageService.on(StorageEvents.REPLICATED_CSR, async (payload: { csrs: string[] }) => {
+      this.logger(`Storage - ${StorageEvents.REPLICATED_CSR}`)
+      this.libp2pService.emit(Libp2pEvents.DIAL_PEERS, await getLibp2pAddressesFromCsrs(payload.csrs))
+      this.serverIoProvider.io.emit(SocketActionTypes.RESPONSE_GET_CSRS, payload)
+      this.registrationService.emit(RegistrationEvents.REGISTER_USER_CERTIFICATE, payload)
+    })
```
leblowl, line 631: Just a small refactor and lint

packages/backend/src/nest/storage/certificates/certificates.store.ts
```diff
@@ -54,27 +50,26 @@ export class CertificatesStore {
...
+    await this.loadedCertificates()
 
     this.logger('Initialized')
   }
 
+  public async loadedCertificates() {
+    this.emitter.emit(StorageEvents.LOADED_CERTIFICATES, {
+      certificates: await this.getCertificates(),
+    })
+  }
```
leblowl, line 72: Just moving this outside the init function so it can be used below

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -72,7 +72,7 @@ export class StorageService extends EventEmitter {
     private readonly localDbService: LocalDbService,
     private readonly orbitDbService: OrbitDb,
     private readonly certificatesRequestsStore: CertificatesRequestsStore,
-    private readonly certificatesStore: CertificatesStore,
+    public readonly certificatesStore: CertificatesStore,
```
leblowl: Making this public so we can access it from the registration service. We can also build wrapper functions in storage service and keep it private.

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -759,11 +754,6 @@ export class StorageService extends EventEmitter {
     })
   }
 
-  public resetCsrAndCertsValues() {
-    this.certificatesRequestsStore.resetCsrReplicatedMapAndId()
-    this.certificatesStore.resetValues()
-  }
```
leblowl, line 782: Is there any need to call `this.certificatesStore.resetValues()`?

EmiM: I think it's needed for restarting Quiet. We do not close backend completely when user leaves community, we just purge data and clean the state, reuse the tor process.

packages/backend/src/nest/storage/storage.service.ts
EmiM: Yeah, I'd prefer to add accessors if you don't mind.

leblowl: Like getters and setters? Or wrapper functions for specific CertificatesStore functions, e.g. `StorageService.getCertificates` which is a wrapper for `CertificatesStore.getCertificates`?

EmiM: Either way is fine for me but I think getter/setter is a slightly better choice in this case.

packages/backend/src/nest/registration/registration.service.ts
```diff
@@ -4,35 +4,75 @@ import { extractPendingCsrs, issueCertificate } from './registration.functions'
...
 import { RegistrationEvents } from './registration.types'
 import Logger from '../common/logger'
+import { StorageService } from '../storage/storage.service'
 
 @Injectable()
 export class RegistrationService extends EventEmitter implements OnModuleInit {
   private readonly logger = Logger(RegistrationService.name)
   public certificates: string[] = []
-  private _permsData: PermsData
+  public permsData: PermsData
```
EmiM: Why is this public? If it's only for setting permsData in test, maybe let's add a setter instead? Or something simliar

EmiM (changes requested):
Just a few comments. I'd also look at e2e tests because none of them passed.

packages/backend/src/nest/registration/registration.service.ts
leblowl: I'm happy to change it back to using a setter as it was before. It's a very minor thing. But just to explain why I changed it: I don't see much benefit of a setter/getter if it's not doing anything beyond setting/getting the value. That's what public is for and doesn't require a variable name like `_permsData`. From a caller's perspective the syntax of a setter and public are the same. If we want to avoid public variables for some reason, then I'd probably prefer using methods like `setPermsData` over JS's setter/getter functionality.

packages/backend/src/nest/storage/storage.service.ts
leblowl: I went with the wrapper here because that approach was already being used in this file

packages/backend/src/nest/storage/storage.service.ts
leblowl: I've added clean methods back in and added them to stores that didn't have them. So the code should be pretty similar to how it was before we refactored things (when everything was in the storage service).

EmiM (approved):

PR #2226, opened 2024-01-15
Fix: Clean up desktop UI console errors/warnings

leblowl:
On desktop, the browser was emitting several errors/warnings that filled up the console. Some of these were related to React and then some of them related to HTML preload. The preload warning says, "A preload for {url} is found, but is not used because the request credentials mode does not match. Consider taking a look at crossorigin attribute." and then a little later it says, "The resource {url} was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally." Removing the crossorigin attribute on the font links removed the preload warning and it looks like everything is preloading correctly to me in dev/production (fonts load at the very beginning and no warnings). Maybe if CORS is not used then it also doesn't want crossorigin attribute.

### Pull Request Checklist

### DIFF
--- a/packages/desktop/src/renderer/components/Settings/SettingsComponent.tsx
+++ b/packages/desktop/src/renderer/components/Settings/SettingsComponent.tsx
@@ -30,17 +30,9 @@ const StyledTabsWrapper = styled(Grid)(() => ({
   width: 168,
 }))
 
-const StyledAppBar = styled(AppBar, { label: 'xxxxx' })(() => ({
+const StyledAppBar = styled(AppBar, { label: 'xxxxx' })(({ theme }) => ({
   backgroundColor: '#fff',
   boxShadow: '0px 1px 3px rgba(0, 0, 0, 0.0)',
-}))
-
-const StyledTabs = styled(Tabs)(({ theme }) => ({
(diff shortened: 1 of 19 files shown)


packages/desktop/src/renderer/components/Settings/SettingsComponent.tsx
```diff
@@ -145,18 +145,18 @@ export const SettingsComponent: React.FC<SettingsComponentProps> = ({
...
-                <Grid style={{ marginTop: '24px', cursor: 'pointer' }}>
-                  <Typography
-                    data-testid='leave-community-tab'
-                    className={classes.leaveComunity}
-                    onClick={leaveCommunityModal.handleOpen}
-                  >
-                    Leave community
-                  </Typography>
-                </Grid>
-              )}
```
leblowl, line 158: Apparently, when nesting certain things inside of material components that are not expecting it, weird errors can happen where it passes attributes to the DOM that it shouldn't. I figured this out by stumbling upon an issue that was describing something similar. I'm not sure where that issue went now, but can probably find it. Of course, it's easy to see how this causes errors by running the current version and the new changes.

siepra (approved):

PR #2227, opened 2024-01-16
refactor: Remove SAVE_OWNER_CERTIFICATE event

leblowl:
The SAVE_OWNER_CERTIFICATE backend handler was only emitting
SEND_COMMUNITY_METADATA so now state-manager emits that event directly
to simplify things. Also the saveOwnerCertToDb saga was only used to
emit the SAVE_OWNER_CERTIFICATE event, so that's been removed and
sendCommunityMetadata is used in its place.

### Pull Request Checklist


packages/state-manager/src/sagas/communities/updateCommunityMetadata/updateCommunityMetadata.saga.ts
```diff
@@ -22,9 +30,10 @@ export function* sendCommunityMetadataSaga(
 
   const communityMetadataPayload: CommunityMetadata = {
     id: community.id,
-    ownerCertificate: community.CA.rootCertString,
-    rootCa: community?.rootCa,
+    ownerCertificate: identity.userCertificate,
```
leblowl, line 33: This appears to be a bug that wasn't caught because we were saving the community metadata twice and the first time we saved it via the SAVE_OWNER_CERTIFICATE event where the bug didn't exist.

packages/backend/src/nest/socket/socket.service.ts
```diff
@@ -160,19 +160,6 @@ export class SocketService extends EventEmitter implements OnModuleInit {
         this.emit(SocketActionTypes.CONNECTION_PROCESS_INFO, ConnectionProcessInfo.REGISTERING_OWNER_CERTIFICATE)
       })
 
-      socket.on(SocketActionTypes.SAVE_OWNER_CERTIFICATE, async (payload: SaveOwnerCertificatePayload) => {
-        this.logger(`Saving owner certificate (${payload.peerId}), community: ${payload.id}`)
-
-        this.emit(SocketActionTypes.SAVED_OWNER_CERTIFICATE, payload)
```
EmiM, line 157: As I understand correctly, this SAVED_OWNER_CERTIFICATE is also not needed anymore because it's emitted in registerOwnerCertificate anyway?

leblowl: Yea, that's my understanding too

Issue #2215, opened 2024-01-12
It shouldn't be possible to create a channel starting with dash. ("-")


kingalg:
Desktop version: 2.0.3-alpha.15 (but according to Emilia it's an old bug)
mobile@2.0.3-alpha.17 (ios 344)

It shouldn't be possible to create a channel starting with dash or space but it is if another special character is used as first and space or dash as second.

If the user tries to use space or dash at the beginning they got correct error message:
<img width="644" alt="Screenshot 2024-01-12 at 14 58 34" src="https://github.com/TryQuiet/quiet/assets/68909168/5dcbf4ab-4113-47db-91a5-aea1610fe9bf">

However if they use any special character that is not allowed in channel name it's suddenly possible to create channel starting with a dash.

<img width="623" alt="Screenshot 2024-01-12 at 15 01 42" src="https://github.com/TryQuiet/quiet/assets/68909168/50822235-03af-4c49-bb9b-e0dd604a68a3">

<img width="431" alt="Screenshot 2024-01-12 at 14 57 17" src="https://github.com/TryQuiet/quiet/assets/68909168/b543c868-7e3a-4225-9d03-79e6ac366d3f">

richardtorres314:
@kingalg I see this was moved to the backlog. Would it be possible for me to work on this? Thank you!

holmesworcester:
@richardtorres314 please do! and generally it's okay to pick something up without our approval. Just ping one of us and we can assign it to you and move to "in progress"

richardtorres314:
@holmesworcester Is this issue good to close? Thank you!

holmesworcester:
Typically we wait until the issue is all the way through our QA process to close it, and @kingalg usually does this.

kingalg:
@richardtorres314, thanks for taking care of this; I confirm that the user is no longer allowed to create a channel starting with a dash. We still need an error message there. Currently, it looks a little broken because the user doesn't get any feedback. It should probably be something like "Name cannot begin with whitespace or dashes," like in the screenshot below. I made a separate issue for that because I see that we are missing an error message also for special characters as well (#2322). If you are interested, you can pick this up, or someone from our team can handle this. 

https://github.com/TryQuiet/quiet/assets/68909168/f70219ff-e7ca-450c-a204-2d1691efb3e1
[...]

PR #2287, opened 2024-02-09
fix: updating channel name creation logic

richardtorres314:
### Pull Request Checklist

- [x] I have linked this PR to related GitHub issue.  #2215

## Before
Before

## After
After


leblowl:
Thanks @richardtorres314 

@holmesworcester I'm happy to merge this as I think it's an improvement, but do we also want an error message in the case where someone provides ".-.-." as a channel name? Currently it looks like the channel must be non empty and cannot start with a space or hyphen. Then for common special characters (like & and .), these get stripped from the channel name and it's not an error. So if we put a channel name like ".-.-." it doesn't start with a dash or space, so it strips the periods and then removes the hyphens and there is no error but it also doesn't allow you to add the channel name because once the hyphens and periods are stripped the channel name is empty.

One easy approach is just provide an error if someone includes special characters, rather than getting fancy and stripping them out.

holmesworcester:
Errors really break the flow though. We should find a library or an open source example for validation in this case and use that. Or enumerate all the cases here, but it's easy to miss one!

richardtorres314:
@holmesworcester totally agree! I could see many cases not covered in our current tests that could lead to this logic needing constant patching.

@leblowl would you happen to know what cases we do/don't want covered? I've found alternatives mentioning that "Channel names have a 21-character limit and can include lowercase letters, non-Latin characters, numbers, and hyphens". Is this similar naming convention we're adopting? Thank you!

leblowl:
@holmesworcester In case there is a misunderstanding, to me it looks like there is still an error in the case of `.-.-.` (the user cannot add a channel with that name), however there is no error message. So do we want a case where the user enters a channel name, doesn't receive any feedback and yet cannot create a channel with that name? I'm happy with however ya'll would like it to be, but just wanted to mention that.

@richardtorres314 I'm not really sure. I'd probably ask @holmesworcester

holmesworcester:
The thing we want to achieve with usernames is uniqueness and some resistance to spoofing usernames with homograph attacks. I think this is less important for channels,  but we decided it would be easiest to use the same validation for now.

We don't want someone to be able to use leading or trailing spaces to spoof a username.

And because the onboarding experience is such a sensitive moment, we don't want to frustrate the user with restrictions that are hard to understand. This is why replacing spaces with hyphens is nice. GitHub uses this approach for repo names, IIRC. In the same spirit, a situation where a user enters some text and gets neither an error nor an automatically modified entry (e.g. hyphens for spaces) is not a good one.

I think these are the only requirements, but I'm happy using something more restrictive too if there's something off the shelf that's easy.

richardtorres314:
@holmesworcester Totally agree with the lack of error messaging! I noticed this was the same experience when a channel name consisted of only special characters: no error yet no channel created (see screenshot below).

Errorless Channel Creation

I'll create an issue for this and move this to the top of my list. Let me know if any additional changes are needed. Thank you!

holmesworcester:
I think creating an issue merging this is a good way! Fwiw, special characters could be turned into hyphens too.

Issue #1987, opened 2023-10-19
Invalid behaviour of pasting multiple files with the same file name


EmiM:
* Copy file with filename A from file explorer
* Paste it to message input- file preview appears
* Don't send the file
* Copy **different** file with same filename A
* Paste it to message input

Expected: second file/image preview appears next to the first preview
What happens: user can see only one preview, file data is replaced underneath to second file data. User can see that this happened only after sending the file to channel

iranimij:
May I fix this issue?

holmesworcester:
That would be awesome!

iranimij:
Would you please assign it to me?

iranimij:
Would you please assign it to me?

holmesworcester:
I just invited you! Ping me when you accept and I'll assign you! Thanks again!

iranimij:
I can not see your invitation

holmesworcester:
Nevermind. Assigned to you!

iranimij:
I tried to replicate the issue, but I couldn't and everything seems ok.

holmesworcester:
@EmiM which platform did you see this on?

iranimij:
Desktop, windows

EmiM:
All desktops, I tagged the issue accordingly

agiledev24:
Hi, @holmesworcester 
I grabbed the first issue and tried fixing it.
Please review!


![Image](https://github.com/TryQuiet/quiet/assets/80997421/5f1f8324-dcff-4098-a40e-0ae3c6e54a90)
![Image](https://github.com/TryQuiet/quiet/assets/80997421/95671102-d118-431f-a51f-f9ed3c0a725a)

holmesworcester:
Thank you! @EmiM would you like to review?

kingalg:
Version: 2.1.1

Thank you @agiledev24! The issue is fixed. 

https://github.com/TryQuiet/quiet/assets/68909168/700cead1-39ab-4246-85b3-7fff705672e3

<img width="410" alt="Screenshot 2024-03-08 at 14 57 47" src="https://github.com/TryQuiet/quiet/assets/68909168/f0df62b3-ee27-4596-bd83-f4f71067a118">

PR #2306, opened 2024-02-19
fix: pasting multiple files #1987

agiledev24:


leblowl (approved):
Looks like an improvement, thanks.

Do you think it also makes sense to remove this logic: https://github.com/TryQuiet/quiet/blob/9693d0d24ffd14cf3db3ec8bbced9b7feafed922/packages/desktop/src/renderer/components/Channel/Channel.tsx#L124-L128
and perhaps move some of this logic out of the component and into `handleClipboardFiles`: https://github.com/TryQuiet/quiet/blob/21aa76162ffc743f8d0b73646254548eb69d6c75/packages/desktop/src/renderer/components/widgets/channels/ChannelInput/ChannelInput.tsx#L375-L378

It looks like perhaps we can also unify the two methods for uploading a file (copy/paste and file manager): https://github.com/agiledev24/quiet/blob/21aa76162ffc743f8d0b73646254548eb69d6c75/packages/common/src/fileData.ts#L4-L13

Just some ideas for the future if you're interested. Thankful for anything that makes the code easier to read.

PR #2307, opened 2024-02-19
Updating channel naming logic

### LINKED ISSUE #2299: Channel creation should show error message if channel cannot be created
Currently, when a channel cannot be created because the channel name does not meet requirements (e.g. contains only special characters), the user does not receive an error message but is not able to create a channel (see below).

Errorless Channel Creation

The proposed fix is to add an error message in those cases.

richardtorres314:
### Pull Request Checklist

#### Before
Before

#### After
After

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 # Fixes:
 
+* Fixes channel name creation logic
 * Remove duplicate introduction messages once again
 * Prevent channel creation with names that start with special character, then a hyphen
 * Choose random ports for Tor services (iOS)
--- a/packages/common/src/naming.test.ts
+++ b/packages/common/src/naming.test.ts
@@ -2,15 +2,11 @@ import { parseName } from './naming'
 
 describe('Naming policy utility functions', () => {
   it.each([
-    ['double-hyp--hens', 'double-hyp-hens'],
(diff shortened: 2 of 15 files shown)


leblowl:
Thanks @richardtorres314. It looks like a few tests fail. And then I was thinking about the case of channel names `Test` and `test`. It looks like we probably want our channel/username comparison to be case insensitive, so `Test === test` and it looks like it was previously. I think just adding the `toLowerCase()` back in there should resolve that.

richardtorres314:
@leblowl I've updated a few of the tests but still have some outstanding failing tests I can't seem to resolve. I also can't seem to build the desktop app to test due to an error that seems to have been introduced a few days ago (error message: `Uncaught Error: You must pass a selector to useSelector`). I'll update this PR when I can rebase off of `develop` and building is resolved.

holmesworcester:
@leblowl @EmiM can you reproduce develop being broken? 

@richardtorres314 can you add any information about your system to help us reproduce the problem if we cannot?

richardtorres314:
@holmesworcester I just found my problem, I switched node environments when working on a different project and failed to revert environments when returning to the project. I can confirm upon switching that all errors have been removed and the app works as expected.

I'll continue working on the tests today. Thank you!

leblowl (approved):
Thanks!

Issue #2332, opened 2024-03-06
Users can't use jpg's for profile photos


holmesworcester:
Steps to reproduce on Ubuntu:

1. Click profile in bottom left
2. Click "Edit photo"
3. Try to select a .jpg 

Expected: it works
Actual: JPG isn't visible in file picker.

kingalg:
Version: 2.2.0-alpha.0

Done.
System: jpg's can be added on Windows, Linux and MacOS and they are visible as avatars on all systems (including iOS and Android).

PR #2353, opened 2024-03-11
fix: Allow JPEG and GIF images as profile photos #2332

islathehut:
* Fixes `UserProfileStore` and `UserProfileContextMenu` to allow JPEG and GIF profile photos as well as PNG.
* Updates E2E tests to include JPEG and GIF cases on profile photo tests
* Adds new backend unit tests for img validation
* Adds new `npm` scripts at the root for conveniently running the desktop app locally instead of having to `cd` into the desktop package folder and linting all packages at once.
* Adds new `npm` scripts for local binary compilation and updates E2E tests allow using local binaries
* Adds documentation for new local binary compilation and testing scripts
* Fixes issues I had while running tor tests locally
* Fix the issue with `ContextMenuProps` having type `unknown` for the field `visible`

GitHub Issue: https://github.com/TryQuiet/quiet/issues/2332

### Pull Request Checklist


islathehut:
Fixes #2332

leblowl (approved):
Thanks. I would move the changelog changes under unreleased, but I can also do it later. Here are some of my thoughts around validation for this feature: https://github.com/TryQuiet/quiet/discussions/2173 in case you are interested ... perhaps the image magic byte sequence check is unnecessary, but it doesn't seem like a big deal.

islathehut:
> Thanks. I would move the changelog changes under unreleased, but I can also do it later. Here are some of my thoughts around validation for this feature: #2173 in case you are interested ... perhaps the image magic byte sequence check is unnecessary, but it doesn't seem like a big deal.

Updated the CHANGELOG files with that formatting.  I think for now we can leave the magic byte checks but would be curious if we could do away with it while still maintaining assurance that images are what they say they are.  Part of me also wonders if we should merge the type check (e.g. `data:image/png;base64`) with the magic byte check for performance (since we'll only have to run one header check per image) and some extra assurance (that way we are validating that not only is the type correct on the uploaded image but it matches the magic byte).

islathehut:
Refactored the photo validation so its consolidated and connected the file type check with the magic number check for added assurance/performance.

EmiM:
Could you add backend unit tests for new file types?

packages/e2e-tests/src/tests/userProfile.test.ts
```diff
@@ -106,11 +106,25 @@ describe('User Profile Feature', () => {
...
+  })
+
+  it('Owner updates their profile photo with JPEG', async () => {
+    const menu = new UserProfileContextMenu(users.owner.app.driver)
+    await menu.openMenu()
+    await menu.openEditProfileMenu()
+    await menu.uploadJPEGPhoto()
+  })
+
+  it('Owner updates their profile photo with GIF', async () => {
```
EmiM, line 120: Could you check locally if userProfile test is passing?
* Last of "Owner updates their profile photo with..." tests uploads gif photo so if the upload suppose to be successful then assertion in "Owner's message with profile photo is visible on channel" should be updated from checking png to gif
* Could you maybe add assertion at the end of each of "Owner updates their profile photo with..." cases to make sure that e.g users do not see upload error message?

EmiM (changes requested):

packages/e2e-tests/src/tests/userProfile.test.ts
islathehut: Added assertions and testing locally now

EmiM (approved):

PR #2354, opened 2024-03-11
feat: type emoji #540

agiledev24:
screenshot

Emoji is shown as we type keyword after colon similar to discord.


EmiM:
@agiledev24 thanks! Is this ready for review? I also see that the check is still failing

Issue #503, opened 2022-05-12
No context menu on right clicking highlighted text


holmesworcester:
User tried to right click on highlighted text to copy a link, but no context menu appeared.

holmesworcester:
related: #568

agiledev24:
Fixed the error and created a PR: 
https://github.com/TryQuiet/quiet/pull/2355
@holmesworcester

holmesworcester:
Awesome and thank you! You might be able to claim credit for a few other open issues with this change too! :)

holmesworcester:
Fixed — closing.

Implemented by 4a71f6840 ("feat(503): Adds desktop context menu", #2719, 2025-03-03), which predates the 9.0 line and ships in `@quiet/desktop@9.0.2`.

`packages/desktop/src/main/main.ts:532` now configures `electron-context-menu`, which supplies a Copy entry for text selections by default, alongside the explicitly enabled link/image entries:

```
contextMenu({
  showInspectElement: false,
  showSaveLinkAs: true,
  showCopyLink: true,
  showSaveImage: true,
  showCopyImage: true,
  showSaveImageAs: true,
})
```

Note: verified from the code and the commit history rather than by running the packaged app. If right-click on selected text still shows no menu on any platform, please reopen with the OS and version.

PR #2355, opened 2024-03-11
fix: show context menu on right click #503

agiledev24:
screenshot2
@holmesworcester

### DIFF
--- a/packages/desktop/package.json
+++ b/packages/desktop/package.json
@@ -80,7 +80,7 @@
   },
   "homepage": "https://github.com/TryQuiet",
   "@comment version": "To build new version for specific platform, just replace platform in version tag to one of following linux, mac, windows",
-  "version": "2.1.2-alpha.9",
+  "version": "2.1.1",
   "description": "Decentralized team chat",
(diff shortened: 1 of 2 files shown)


packages/desktop/package.json
```diff
@@ -80,7 +80,7 @@
   },
   "homepage": "https://github.com/TryQuiet",
   "@comment version": "To build new version for specific platform, just replace platform in version tag to one of following linux, mac, windows",
-  "version": "2.1.2-alpha.9",
+  "version": "2.1.1",
```
leblowl: I think we want to keep it as 2.1.2-alpha.9

leblowl (changes requested):
Thanks. It looks like there might be one small issue in the package.json file and then could you also update the CHANGELOG when you get a moment (add a new line under [unreleased])?

leblowl:
@agiledev24 It looks like the check and desktop tests are failing. If we get those passing, I think we can merge this

holmesworcester:
I'm going to pick this up and try to complete it.

Issue #2358, opened 2024-03-12
Create a documentation site


islathehut:
Currently documentation is pretty sparse and lives in a few places, the goal here is to consolidate the docs in one place that is easy to consume and serve to users.  Documentation should live with the code, be easy to deploy, easy to access, and should (ultimately) make onboarding developers easier.

holmesworcester:
@ikoenigsknecht is this still in waiting for review? i think it was blocked by our gh pages setup or some ambiguity around where to put it, correct?

PR #2359, opened 2024-03-12
Feat: Add documentation site using Docusaurus #2358

islathehut:
This adds [Docusaurus](https://docusaurus.io/) to Quiet, enabling us to have a fully-featured documentation site locally or deployed to production.

I did my best to migrate existing documentation from the markdown files in the repo (including the README) but I'm sure I missed something.  I also added some documentation on how Docusaurus works and is configured with links to the official docs.  

Styling is a rough first try with Quiet colors but we can certainly iterate on it.

Here are some screenshots so folks can get a feel for the styling/setup:

Screenshot 2024-03-12 at 10 58 50 AM

Screenshot 2024-03-12 at 10 59 00 AM

Screenshot 2024-03-12 at 10 59 34 AM

### Pull Request Checklist


islathehut:
Implements #2358

islathehut:
NOTE: Much of this is boilerplate generated by Docusaurus, if you have any questions about anything reach out!

holmesworcester:
This looks great! I think we might want to go a bit further in consolidating docs. Right now it looks like there are duplicates to files like CONTRIBUTING.md and there's a separate docs folder. My worry is that someone might update the docs in either place, leading to lost or conflicting updates. Would you like to propose some consolidation?

islathehut:
> This looks great! I think we might want to go a bit further in consolidating docs. Right now it looks like there are duplicates to files like CONTRIBUTING.md and there's a separate docs folder. My worry is that someone might update the docs in either place, leading to lost or conflicting updates. Would you like to propose some consolidation?

Originally I was gonna delete those but I noticed that at least some are referenced on the website and didn't wanna break things.

holmesworcester:
Then I think we should either:

1. leave them out of Docusaurus so we do not have to update both.
2. update the existing files to not have any content and refer people to Docusaurus.
3. Remove them and update links as best we can. 

For README.md I think we should definitely do #1, since everyone would expect to find a README, and because the Github project page content comes from this. 

Some context on this is that it's very easy for the development instructions to fall out of date when somebody makes changes, it's somewhat hard to catch this systematically because internally we always know "the trick" to get things working, and it's always discouraging for new volunteers to follow our instructions and run into a problem because they're out of date. So we want to avoid having two sets of instructions because then we have even more ways to fail to update the instructions correctly. 

We're also using the [Github Wiki](https://github.com/TryQuiet/quiet/wiki) for docs and page content. How do you think we should integrate that? One solution here would be to distinguish between developer-facing docs and public-facing docs like the [Threat Model](https://github.com/TryQuiet/quiet/wiki/Threat-Model), and move all developer-facing docs into Docusaurus, but keep others in the wiki. How does that sound?

PR #2368, opened 2024-03-15
feat: Add basic mechanism for displaying uncaught exceptions

leblowl:
### Pull Request Checklist

### DIFF
--- a/packages/state-manager/src/sagas/socket/startConnection/startConnection.saga.ts
+++ b/packages/state-manager/src/sagas/socket/startConnection/startConnection.saga.ts
@@ -178,6 +178,7 @@ export function subscribe(socket: Socket) {
       log.error(payload)
       console.error(payload, payload.trace)
       emit(errorsActions.handleError(payload))
+      alert(JSON.stringify(payload))
     })
     // Certificates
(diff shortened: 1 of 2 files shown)


packages/state-manager/src/sagas/socket/startConnection/startConnection.saga.ts
```diff
@@ -178,6 +178,7 @@ export function subscribe(socket: Socket) {
       log.error(payload)
       console.error(payload, payload.trace)
       emit(errorsActions.handleError(payload))
+      alert(JSON.stringify(payload))
```
EmiM, line 181: I understand that this is only temporary for mobile debugging purposes?
Otherwise I would ask why not use our ErrorModal

EmiM (approved):

packages/state-manager/src/sagas/socket/startConnection/startConnection.saga.ts
leblowl: Yea I was thinking to just deploy something quick so I can get feedback from Kinga on monday. Then we can revert the changes and improve it. The changes will only go to the 2.1.2 branch, not develop. We can also spend more time on it now, but that was my plan at least.

Issue #2362, opened 2024-03-14
Update logger output format to include more information


islathehut:
I think using the standard format found in [logf4js](https://stritti.github.io/log4js/docu/users-guide.html) (which is just a copy of the format in log4j) would be a good place to start.  We can tweak from there.

PR #2369, opened 2024-03-18
chore: Add winston logger and update logging in state-manager #2362

islathehut:
## Overview

This adds a more robust logger to the logging package and updates `state-manager` to use the new logger.  The logger is configurable so we can output to console, files, etc. and combine multiple transports.  It also provides colorized formatting and easily enables multi-part logger names (e.g. `state-manager:appConnection:selectors`).

## Debug Flag

I included logic for overriding the log level in a given logger if the `DEBUG` flag contains any part of the logger name.  This makes it backwards-compatible with the logic we have fin the original logger implementation.

## Why `state-manager`?  

I found while debugging connect/disconnect issues that logs from `state-manager` didn't show up in the console with the other logs; I could get them to show up in the browser console but it was inconvenient to jump back and forth and match them up with relevant logs from the `backend`.  This logger allows us to write to a file and then tail that file in the app so we get all of our logs in one place.

### Pull Request Checklist


islathehut:
#2362

PR #2427, opened 2024-04-11
Better peer sorting and updated initial diallng

islathehut:
### Pull Request Checklist

### DIFF
--- a/packages/backend/src/nest/connections-manager/connections-manager.service.tor.spec.ts
+++ b/packages/backend/src/nest/connections-manager/connections-manager.service.tor.spec.ts
@@ -31,6 +31,7 @@ import waitForExpect from 'wait-for-expect'
 import { Libp2pEvents } from '../libp2p/libp2p.types'
 import { sleep } from '../common/sleep'
 import { createLibp2pAddress } from '@quiet/common'
+import { lib } from 'crypto-js'
 
 jest.setTimeout(100_000)
 
@@ -121,7 +122,7 @@ afterEach(async () => {
 })
 
 describe('Connections manager', () => {
-  it('saves peer stats when peer has been disconnected', async () => {
+  it.only('saves peer stats when peer has been disconnected', async () => {
     class RemotePeerEventDetail {
       peerId: string
 
@@ -205,8 +206,8 @@ describe('Connections manager', () => {
     await sleep(5000)
     // It looks LibP2P dials peers initially when it's started and
     // then IPFS service dials peers again when started, thus
-    // peersCount * 2
-    expect(spyOnDial).toHaveBeenCalledTimes(peersCount * 2)
+    // peersCount-1 * 2 because we don't dial ourself (the first peer in the list)
+    expect(spyOnDial).toHaveBeenCalledTimes((peersCount - 1) * 2)
     // Temporary fix for hanging test - websocketOverTor doesn't have abortController
     await sleep(5000)
   })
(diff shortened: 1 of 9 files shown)


packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -517,7 +517,7 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
       agent: this.socksProxyAgent,
       localAddress: this.libp2pService.createLibp2pAddress(onionAddress, peerId.toString()),
       targetPort: this.ports.libp2pHiddenService,
-      peers: peers ?? [],
+      peers: peers ? peers.slice(1) : [],
```
islathehut, line 520: This is to avoid the self-dialing we normally see.

packages/common/src/sortPeers.ts
```diff
@@ -14,17 +13,19 @@ This is the very simple algorithm for evaluating the most wanted peers.
 export const filterAndSortPeers = (
   peersAddresses: string[],
   stats: NetworkStats[],
-  localPeerAddress?: string
+  localPeerAddress?: string,
+  includeLocalPeerAddress: boolean = true
 ): string[] => {
   peersAddresses = filterValidAddresses(peersAddresses)
+  const currentlyConnected = [...stats].filter(peer => peer.connectionTime === 0)
```
islathehut, line 20: When a peer connects we set the `connectionTime` to zero so peers with that value present should be currently connected (or connected the last time we were online).  This means for invitation links the peers are more likely to be online and when we go offline, if we went offline recently, we're more likely to try connecting to online peers (this becomes less likely the longer a user is offline but that's true for connection time filtering, too).

packages/backend/src/nest/libp2p/process-in-chunks.service.ts
```diff
@@ -1,56 +1,88 @@
...
+  private taskQueue: queue<ProcessTask<T>>
   private processItem: (arg: T) => Promise<any>
   private readonly logger = Logger(ProcessInChunksService.name)
   constructor() {
     super()
   }
 
   public init(data: T[], processItem: (arg: T) => Promise<any>, chunkSize: number = DEFAULT_CHUNK_SIZE) {
-    this.data = data
+    this.logger(`Initializing process-in-chunks.service with peers ${JSON.stringify(data, null, 2)}`)
```
holmesworcester, line 27: We don't have to address this now, but eventually we want to sanitize addresses like onion addresses or peerid's from the logs.

packages/common/src/sortPeers.ts
holmesworcester: Should we be saving who we were last connected to, whenever we shut down, or periodically?

islathehut: That is something I also considered, you can give libp2p a datastore so it can persist the peer data it had on shutdown.

packages/backend/src/nest/connections-manager/connections-manager.service.tor.spec.ts
```diff
@@ -121,7 +122,7 @@ afterEach(async () => {
 })
 
 describe('Connections manager', () => {
-  it('saves peer stats when peer has been disconnected', async () => {
+  it.only('saves peer stats when peer has been disconnected', async () => {
```
siepra, line 125: Is this `only` here in purpose?

PR #2438, opened 2024-04-15
feat: 2312 pass invitation data to createNetwork saga and LAUNCH_COMMUNITY…

EmiM:
… payload

### Pull Request Checklist


packages/backend/src/nest/storageServerProxy/storageServerProxy.service.ts
```diff
@@ -0,0 +1,113 @@
...
+class HTTPResponseError extends Error {
+  response: Response
+  constructor(message: string, response: Response) {
+    super(`${message}: ${response.status} ${response.statusText}`)
+    this.response = response
+  }
+}
+
+@Injectable()
+export class ServerProxyService extends EventEmitter {
```
leblowl: I would name this something like `StorageServiceClient`. When I see proxy I think of something like this: https://www.techtarget.com/whatis/definition/proxy-server

packages/desktop/src/renderer/components/CreateJoinCommunity/JoinCommunity/JoinCommunity.tsx
```diff
@@ -20,7 +21,12 @@ const JoinCommunity = () => {
 
   const torBootstrapProcessSelector = useSelector(connection.selectors.torBootstrapProcess)
 
+  const downloadInviteDataError = useSelector(
+    errors.selectors.generalErrorByType(SocketActionTypes.DOWNLOAD_INVITE_DATA)
```
leblowl: Aren't we removing `DOWNLOAD_INVITE_DATA`?

packages/state-manager/src/sagas/appConnection/connection.selectors.ts
```diff
@@ -55,6 +56,15 @@ export const invitationUrl = createSelector(
...
     const pairs = p2pAddressesToPairs(initialPeers)
+    const v2Data = {
+      id: '209348023',
+      ownerCertificate: 'ownerCertificate',
+      rootCa: 'rootCa',
+      ownerOrbitDbIdentity: ownerOrbitDbIdentity,
+      peerList: pairsToP2pAddresses(pairs),
+      psk: communityPsk,
+    }
+    console.log('V2 DATA:', JSON.stringify(v2Data))
```
leblowl: Is this just for testing?

packages/types/src/socket.ts
```diff
@@ -73,6 +73,12 @@ export enum SocketActionTypes {
   PEER_DISCONNECTED = 'peerDisconnected',
   TOR_INITIALIZED = 'torInitialized',
 
+  // ====== Storage server ======
+
+  SET_STORAGE_SERVER_ADDRESS = 'setStorageServerAddress',
+  DOWNLOAD_STORAGE_SERVER_DATA = 'downloadStorageServerData',
+  UPLOAD_STORAGE_SERVER_DATA = 'uploadStorageServerData',
```
leblowl: Are any of these used?

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -424,9 +464,9 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
 
     const community = {
       id: payload.id,
-      peerList: [...new Set([localAddress, ...payload.peers])],
-      psk: payload.psk,
-      ownerOrbitDbIdentity: payload.ownerOrbitDbIdentity,
+      peerList: [...new Set([localAddress, ...metadata.peers])],
+      psk: metadata.psk,
+      ownerOrbitDbIdentity: metadata.ownerOrbitDbIdentity,
```
leblowl, line 469: Since we are adding `inviteData` to the `Community` model, I think we should also store that here.

packages/backend/src/nest/libp2p/process-in-chunks.service.ts
```diff
@@ -71,7 +71,7 @@ export class ProcessInChunksService<T> extends EventEmitter {
   }
 
   public async process() {
-    this.logger(`Processing ${this.taskQueue.length} items`)
+    this.logger(`Processing ${this.taskQueue.length()} items`)
```
leblowl, line 74: Thanks for fixing this

leblowl (commented):
Looks good to me overall, nice work.

packages/backend/src/nest/connections-manager/connections-manager.service.ts
EmiM: I was thinking about it however in theory `inviteData` is only temporary. Once you joined you don't need it anymore.
I can save it here for consistency.

packages/types/src/socket.ts
EmiM: Leftovers 👍🏻 Will be removed

packages/state-manager/src/sagas/appConnection/connection.selectors.ts
EmiM: Yup

packages/desktop/src/renderer/components/CreateJoinCommunity/JoinCommunity/JoinCommunity.tsx
EmiM: Leftovers

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -519,6 +520,7 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
     // Unblock websocket endpoints
     this.socketService.resolveReadyness()
 
+    await this.localDbService.deleteInviteData()
```
leblowl, line 523: I think we should leave inviteData always

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -467,6 +467,7 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
       peerList: [...new Set([localAddress, ...metadata.peers])],
       psk: metadata.psk,
       ownerOrbitDbIdentity: metadata.ownerOrbitDbIdentity,
+      inviteData,
```
leblowl, line 470: I would add a comment here to remind us that current users will not have this field unless we do a migration

leblowl (changes requested):

packages/backend/src/nest/connections-manager/connections-manager.service.ts
leblowl: If we want to keep the `inviteData`,
```
  version?: InvitationDataVersion.v2
  cid: string
  token: string
  serverAddress: string
  inviterAddress: string
```
for some reason because we think the original invite data is important, then we can have `Community.inviteData` and `Community.storageService` with the following schema,
```
  token: string
  serverAddress: string
```

Otherwise, if we don't care about saving the original invite data and de-duplication is our focus, we can have something like `Community.inviterAddress` (inviter's Tor address), `Community.inviteCid` and then `Community.storageService` with the schema above.

I think it's probably easiest to fit with the existing PR, to have `Community.inviteData` and `Community.storageService` and accept that there could be some duplication (`token` and `serverAddress` would exist in both, but the `storageService` field is meant to change, while the `inviteData` represents the original invite data which doesn't change).

@ikoenigsknecht do you have any thoughts?

leblowl: The schema for `InviteData` is here: https://github.com/TryQuiet/quiet/blob/68313c24a4d3942ebc054a450b01e081ce659d68/packages/types/src/network.ts#L31

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -424,9 +464,10 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
 
     const community = {
       id: payload.id,
-      peerList: [...new Set([localAddress, ...payload.peers])],
-      psk: payload.psk,
-      ownerOrbitDbIdentity: payload.ownerOrbitDbIdentity,
+      peerList: [...new Set([localAddress, ...metadata.peers])],
+      psk: metadata.psk,
+      ownerOrbitDbIdentity: metadata.ownerOrbitDbIdentity,
+      inviteData,
```
leblowl, line 470: Sorry for the confusion, I think we should leave this out of the backend model for now and just treat it temporary for now until we are sure we want to store it.

leblowl (approved):

PR #2455, opened 2024-04-18
fix: Various fixes related to peers, CSRs and backend startup

leblowl:
Fixes for the following issues:
- Peers can be deleted if CSRs don't sync
- Backend starting before the frontend is ready, resulting in missed events
- Adding duplicate CSRs

I found an existing bug where we could actually remove peers (originating from the invite link) from the peer list if certificates replicated before CSRs, because we used CSRs to update the peer list. @ikoenigsknecht, this could be related to the issue where you said the dialer stopped working, perhaps it didn't have any peers to dial. In the PR, we use the same definition for all users as the frontend (CSRs + certs) when finding peers and we always retain peers from the invite link even if we don't have their cert or CSR yet.

The peer issue was really easy to reproduce when CSRs didn't replicate, which I was also lucky enough to reproduce. Apparently due to our duplicate CSR issue (I had around 3670 CSRs), that caused CSRs to take 30 minutes or so to replicate. So that's interesting. It looks like we had this saga checkLocalCsr which would re-add a CSR in certain cases or not add the user's CSR at all. In the case where the CSR doesn't get added at all, we could get the disappearing message effect that @kingalg saw because we don't display messages unless we know about the user (CSR or cert must exist). In this PR, we add the CSR once (right after community is launched).

Also the backend was starting before the frontend was ready on mobile only. Thanks @EmiM  for spotting that. This resulted in missing a CHANNEL_SUBSCRIBED event and thus not being able to send a message (having it greyed out indefinitely). It looks like mobile connects with native code to the backend in order to get push notifications which causes the backend to start, then the actual frontend code connects and by then it's missed the CHANNEL_SUBSCRIBED event. So I added a new START event for the frontend to explicitly tell the backend it's ready.

### Pull Request Checklist


packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -729,18 +750,55 @@ export class StorageService extends EventEmitter {
...
+   */
+  // TODO: Add test case around have both certificates and CSRs
+  public async getAllUsers(): Promise<UserData[]> {
+    const csrs = await this.certificatesRequestsStore.getCsrs()
+    const certs = await this.certificatesStore.getCertificates()
+    const allUsersByKey: Record<string, UserData> = {}
+
+    this.logger(`Retrieving all users. CSRs count: ${csrs.length} Certificates count: ${certs.length}`)
+
+    for (const cert of certs) {
```
EmiM, line 759: I'm guessing that retrieving both certs and csrs is for the case when one of those DBs are not fully synced? In any other case you can't have certificate without csr.

leblowl: This mirrors how we do things on the frontend and is helpful in the case where certs replicate before csrs.

packages/state-manager/src/sagas/communities/launchCommunity/launchCommunity.saga.ts
```diff
@@ -63,4 +66,6 @@ export function* launchCommunitySaga(
   }
 
   yield* apply(socket, socket.emitWithAck, applyEmitParams(SocketActionTypes.LAUNCH_COMMUNITY, payload))
+
+  yield* put(identityActions.saveUserCsr())
```
leblowl, line 70: Add the user's CSR once after launching community.

packages/state-manager/src/sagas/identity/registerUsername/registerUsername.saga.ts
```diff
@@ -87,12 +87,24 @@ export function* registerUsernameSaga(
...
+
+  if (community.CA?.rootCertString) {
+    yield* put(communitiesActions.createCommunity(community.id))
+  } else {
+    if (!isUsernameTaken) {
+      yield* put(communitiesActions.launchCommunity(community.id))
+    } else {
+      yield* put(identityActions.saveUserCsr())
+    }
+  }
```
leblowl, line 109: Just inlining `registerCertificate` here to simplify things

packages/state-manager/src/sagas/messages/sendMessage/sendMessage.saga.ts
```diff
@@ -14,28 +14,30 @@ export function* sendMessageSaga(
...
     return
   }
 
+  console.log(`Sending message ${id} to channel ${channelId}`)
+
+  const pubKey = yield* call(pubKeyFromCsr, identity.userCsr.userCsr)
+  const keyObject = yield* call(loadPrivateKey, identity.userCsr.userKey, config.signAlg)
+  const signatureArrayBuffer = yield* call(sign, action.payload.message, keyObject)
+  const signature = yield* call(arrayBufferToString, signatureArrayBuffer)
+  const createdAt = yield* call(getCurrentTime)
```
leblowl, line 40: Re-organizing a bit and adding additional logs.

packages/backend/src/nest/local-db/local-db.service.ts
```diff
@@ -95,21 +95,9 @@ export class LocalDbService {
...
-    if (!peers) {
-      const currentCommunity = await this.getCurrentCommunity()
-      if (!currentCommunity) {
-        throw new Error('No peers were provided and no community was found to extract peers from')
-      }
-      peers = currentCommunity.peerList
-      if (!peers) {
-        throw new Error('No peers provided and no peers found on current stored community')
-      }
-    }
```
leblowl, line 111: Moved this logic into updatePeersList

islathehut (approved):

PR #2493, opened 2024-04-29
refactor: add base class for Stores

EmiM:
### Pull Request Checklist


packages/backend/src/nest/common/types.ts
```diff
@@ -6,11 +6,6 @@ export interface PublicChannelsRepo {
   eventsAttached: boolean
 }
 
-export interface DirectMessagesRepo {
```
adrastaea, line 9: [question] Is this an intended removal? It seems like it wasn't being utilized, but want to be sure that we want it fully removed.

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -309,7 +310,7 @@ export class StorageService extends EventEmitter {
   }
 
   public async updateCommunityMetadata(communityMetadata: CommunityMetadata): Promise<CommunityMetadata | undefined> {
```
adrastaea, line 310: [clarity/maybe bug?] It seems like currently the code in `CommunityMetadataStore` will discard the new metadata when there is an old entry with a `rootCa` and `ownerCertificate`. Should we fix this bug here, change the name of this function to something that suggests that this is a un-overwriteable, or put a comment in here that suggests that the functionality to update existing community metadata is not implemented yet? I'm fine with leaving that to another PR to get this through, but wanted to highlight that `updateCommunityMetadata` is misleading as is.
```ts
      // [Excerpt from CommunityMetadataStore]
      // FIXME: update community metadata if it has changed (so that
      // we can migrate community metadata easily)
      const oldMeta = this.getStore().get(newMeta.id)
      if (oldMeta?.ownerCertificate && oldMeta?.rootCa) {
        return oldMeta
      }
 ```

**Review by adrastaea — APPROVED** [review]
A couple questions, nitpicks, but looks like a great improvement to me!

**Comment by adrastaea** [comment]
Just noticed the changelog needs updated though

File: packages/backend/src/nest/storage/base.store.ts [hunk 3]
```diff
@@ -0,0 +1,27 @@
...
+  getAddress(): Store['address'] {
+    return this.getStore().address
+  }
+
+  async close(): Promise<void> {
+    await this.store?.close()
+  }
+
+  abstract init(): Promise<void>
+  abstract addEntry(entry: V): Promise<V | undefined>
```
leblowl: To me, this abstraction doesn't seem like the best fit. KeyValueStore has a map-like interface (e.g. `setEntry` `getEntry`), while EventStore has a list-like interface (e.g. `addEntry`, `getEntries`). So perhaps an additional base class for each type of store would be more appropriate or just remove `addEntry` from this base class which is probably what I'd recommend for now.

packages/backend/src/nest/storage/base.store.ts
```diff
@@ -0,0 +1,27 @@
...
+    }
+    return this.store
+  }
+
+  getAddress(): Store['address'] {
+    return this.getStore().address
+  }
+
+  async close(): Promise<void> {
+    await this.store?.close()
```
leblowl, line 25: I would add some logs here to indicate that we are closing the store.

packages/backend/src/nest/common/types.ts
holmesworcester: This was a partly finished dm implementation that we decided to remove.

packages/backend/src/nest/storage/base.store.ts
islathehut: Agreed.  We could have two layers of base class:

1. A base store that has the universal stuff like close, init, etc.
2. Individual base classes for each type of store that more closely follows the underlying orbitdb store

packages/backend/src/nest/common/types.ts
EmiM: Yeah, long time gone. When team decides to add DMs it will surely be after few refactorings.

packages/backend/src/nest/storage/base.store.ts
EmiM: I was going for more abstract/general interface, hence the 'addEntry'. It would implement 'setting' for kvstore and 'adding' for event store. However you may have a point so I'll change that, thanks!

EmiM: Btw. do you have a strong preference about `setEntry` params for the base store class of kvstore? If we want it to be closer to orbitdb store method (and more generic) then we should pass key and value here. Currently our stores use key taken from value (e.g meta.id/userProfile.pubKey) so technically we can pass only value.
I'd lean towards key+value.

leblowl: I was thinking about that too. I think probably key+value is easier to read and understand what's going on

islathehut (approved):
Changes look good to me!

Issue #2497, opened 2024-04-30
Clear LevelDB after OrbitDB when leaving a community


leblowl:
Received the following log on Android when leaving a community:
```
04-30 12:56:16.004  4770  4919 E NODEJS-MOBILE: Error: Failed to update peers list - community missing
04-30 12:56:16.004  4770  4919 E NODEJS-MOBILE:     at StorageService.updatePeersList (webpack://@quiet/backend/./src/nest/storage/storage.service.ts?:296:19)
04-30 12:56:16.004  4770  4919 E NODEJS-MOBILE:     at async CertificatesRequestsStore.eval (webpack://@quiet/backend/./src/nest/storage/storage.service.ts?:276:13)
04-30 12:56:16.005  4770  4898 E nodejs  : webpack://@quiet/backend/./src/nest/connections-manager/connections-manager.service.ts?:111
04-30 12:56:16.005  4770  4898 E nodejs  :             throw new Error();
04-30 12:56:16.005  4770  4898 E nodejs  :             ^
04-30 12:56:16.005  4770  4898 E nodejs  : 
04-30 12:56:16.005  4770  4898 E nodejs  : Error
04-30 12:56:16.005  4770  4898 E nodejs  :     at process.eval (webpack://@quiet/backend/./src/nest/connections-manager/connections-manager.service.ts?:111:19)
04-30 12:56:16.005  4770  4898 E nodejs  :     at process.emit (node:events:513:28)
04-30 12:56:16.005  4770  4898 E nodejs  :     at processEmit [as emit] (webpack://@quiet/backend/./node_modules/signal-exit/index.js?:199:34)
04-30 12:56:16.005  4770  4898 E nodejs  :     at emit (node:internal/process/promises:140:20)
04-30 12:56:16.005  4770  4898 E nodejs  :     at processPromiseRejections (node:internal/process/promises:274:27)
04-30 12:56:16.005  4770  4898 E nodejs  :     at processTicksAndRejections (node:internal/process/task_queues:97:32)
```

It looks like we clear/close LevelDB before closing OrbitDB when leaving a community, but it should be the other way around.

kingalg:
@leblowl how can I check if it's done? Was it only the log or was it also causing leaving community to break?
If it was causing leaving a community to break then it's fixed but if I should also check logs, let me know.

kingalg:
mobile@2.2.0-alpha.7 

After consultation with Lucas. Not happening right now, I'll keep an eye on mentioned error.

Issue #2496, opened 2024-04-30
App can get stuck on 'Loading messages' when joining


leblowl:
I think this can happen on any platform, but has happened recently on iOS and Android. The logs for Android show everything working correctly, but the certificates don't sync, even after 10 minutes. 2 peers were connected. I restarted the app and sent a message (restarting allowed me to access the general channel even though nobody was registered due to missing certs) and it took 20 minutes for the message to arrive at another peer. Certs still haven't arrived yet. Looks like slow network conditions. This doesn't happen often for me, but to investigate there are additional logs related to the gossipsub, libp2p, OrbitDB that we can enable.

kingalg:
mobile@2.2.0-alpha.7 | iOS 376

Fixed (or at least I didn't encounter any problems with this in this and previous alpha - joining took a long time but it was completed and synced without need of restarting).

PR #2499, opened 2024-05-01
fix: Reorder the closing of services, prevent sagas running multiple times and close backend server properly

leblowl:
Fixes:

- We were closing LevelDB before closing OrbitDB which resulted in an exception.

  Solution for #2497

- Sagas were running multiple times after leaving a community due to our use of `takeEvery`.
  
  Solution for #2509

- The backend server wasn't getting closed properly when the app went to background on iOS.

  Solution for #2500
  
- There was a bug in the (old) version of OrbitDB that we are using that caused it to send heads in a loop.

  Solution for #2496 and #2512

### Pull Request Checklist


packages/backend/src/backendManager.ts
```diff
@@ -109,12 +109,12 @@ export const runBackendMobile = async () => {
     { logger: ['warn', 'error', 'log', 'debug', 'verbose'] }
   )
 
-  rn_bridge.channel.on('close', async () => {
+  rn_bridge.channel.on('close', () => {
     const connectionsManager = app.get<ConnectionsManagerService>(ConnectionsManagerService)
-    await connectionsManager.pause()
+    connectionsManager.pause()
```
leblowl, line 114: No need for async here since nothing awaits the event handler

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -224,46 +224,42 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
...
     }
     if (this.localDbService) {
       this.logger('Closing local storage')
       await this.localDbService.close()
     }
-    if (this.libp2pService) {
-      this.logger('Stopping libp2p')
-      await this.libp2pService.close()
-    }
   }
```
leblowl, line 249: Re-organizing. Data server should get closed first and then OrbitDB, then LibP2P

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -224,46 +224,42 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
...
-    this.serverIoProvider.io.close()
+  public async closeSocket() {
+    await this.socketService.close()
   }
 
   public async pause() {
     this.logger('Pausing!')
-    this.logger('Closing socket!')
-    this.closeSocket()
+    await this.closeSocket()
```
leblowl, line 257: I remove the log because there is a similar log in `sockerService.close`, but we can keep it too, just let me know.

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -224,46 +224,42 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
...
     this.logger('Pausing libp2pService!')
     this.peerInfo = await this.libp2pService?.pause()
     this.logger('Found the following peer info on pause: ', this.peerInfo)
   }
 
   public async resume() {
     this.logger('Resuming!')
-    this.logger('Reopening socket!')
     await this.openSocket()
```
leblowl, line 265: Again, I remove the log because there is a similar log in socketService.listen. We can keep both also, just let me know.

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -286,20 +282,30 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
...
+      await this.localDbService.close()
+    }
 
-    this.logger('Closing services')
-    await this.closeAllServices({ saveTor: true })
+    this.logger('Resetting tor')
+    this.tor.resetHiddenServices()
 
     this.logger('Purging data')
     await this.purgeData()
```
leblowl, line 311: The important piece here is purging/closing LevelDB after stopping OrbitDB. However, I didn't want to get into a larger refactor so in order to make that work, I copied the contents of `closeAllServices` into this function. Other things still depend on close all services. I think the only difference between the two is that this function purges LevelDB and `closeAllServices` doesn't. So we could add another parameter to `closeAllServices` like `purgeLocalDb: boolean` or something.

packages/mobile/src/store/init/startConnection/startConnection.saga.ts
```diff
@@ -49,17 +61,20 @@ export function* startConnectionSaga(
   })
   yield* fork(handleSocketLifecycleActions, socket, action.payload)
   // Handle opening/restoring connection
-  yield* takeEvery(initActions.setWebsocketConnected, setConnectedSaga, socket)
+  yield* takeLeading(initActions.setWebsocketConnected, setConnectedSaga, socket)
```
leblowl, line 54: This only needs to happen in sequence

packages/mobile/src/store/init/startConnection/startConnection.saga.ts
```diff
@@ -49,17 +61,20 @@ export function* startConnectionSaga(
...
-  // Handle suspending current connection
-  yield* takeEvery(initActions.suspendWebsocketConnection, cancelRootTaskSaga, task)
-  console.log('Frontend is ready. Starting backend...')
+
   // @ts-ignore - Why is this broken?
   yield* apply(socket, socket.emit, [SocketActionTypes.START])
+
+  // Handle suspending current connection
+  const suspendAction = yield* take(initActions.suspendWebsocketConnection)
+  yield* call(cancelRootTaskSaga, task, suspendAction)
```
leblowl: This only needs to happen once (`takeEvery` is not necessary and causes issues)

packages/mobile/src/store/nativeServices/events/nativeServicesCallbacks.ts
```diff
@@ -9,10 +9,18 @@ import { navigationActions } from '../../navigation/navigation.slice'
...
+    const channel = yield* call(deviceEvents)
+    while (true) {
+      const action = yield* take(channel)
+      yield put(action)
+    }
+  } finally {
+    console.log('nativeServicesCallbacksSaga stopping')
+    if (yield cancelled()) {
+      console.log('nativeServicesCallbacksSaga cancelled')
+    }
```
leblowl, line 23: Just adding additional logging around all of these long-running sagas (anything that runs until cancelled... so anything with a `takeEvery` in it or other `while (true)` loop)

packages/mobile/src/store/nativeServices/leaveCommunity/leaveCommunity.saga.ts
```diff
@@ -9,11 +9,8 @@ import { ScreenNames } from '../../../../src/const/ScreenNames.enum'
 
 export function* leaveCommunitySaga(): Generator {
   console.log('Leaving community')
-
   // Restart backend
   yield* putResolve(app.actions.closeServices())
-
-  yield takeLeading(initActions.canceledRootTask.type, clearReduxStore)
```
leblowl, line 16: We never want to run `takeLeading` or `takeEvery` more than once because, for example, in this case it will result in `clearReduxStore` being run multiple times. I moved this into the `rootSaga` instead.

packages/mobile/src/store/root.saga.ts
```diff
@@ -1,19 +1,29 @@
...
-    // Below line is reponsible for displaying notifications about messages from channels other than currently viewing one
-    takeEvery(publicChannels.actions.markUnreadChannel.type, showNotificationSaga),
-  ])
+  console.log('rootSaga starting')
+  try {
+    yield all([
+      fork(setupCryptoSaga),
+      fork(initMasterSaga),
+      fork(navigationMasterSaga),
+      fork(nativeServicesMasterSaga),
```
leblowl, line 51: Running these multiple times without cancelling and restarting `rootSaga` results in multiple sagas being run for each event. I didn't see any need to cancel and restart `rootSaga` like we do with `useIO`, so I just used `fork` here.

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -223,47 +223,51 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
     }
   }
 
-  public async closeAllServices(options: { saveTor: boolean } = { saveTor: false }) {
+  public async closeAllServices(
+    options: { saveTor: boolean; purgeLocalDb: boolean } = { saveTor: false, purgeLocalDb: false }
```
leblowl: Feels a little clunky to add another parameter here since `closeAllServices` is doing a little more than just closing things now, but it seems like the simplest change without a larger refactor. Also if `closeAllServices` is purging `local-db` then maybe it should also purge data on the filesystem.

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -223,47 +223,51 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
...
+  ) {
+    this.logger('Closing services')
+
+    await this.closeSocket()
```
leblowl, line 229: Close the data server first

islathehut (approved):
This looks great!

packages/backend/src/nest/connections-manager/connections-manager.service.ts
adrastaea: [question] Why do we need to add this parameter? Maybe I'm reading into the name `localDbService` too much, but it seems like the data that would be cleared by `this.localDbService.purge()` would be the same data that would be cleared by `this.purgeData()` called in `leaveCommunity()` since that `purgeData()` seems to holistically remove all the local db data from the filesystem. Adding a side effect to `closeAllServices()` seems like it would be a bit of a code smell.

packages/backend/src/nest/socket/socket.service.ts
```diff
@@ -27,13 +27,17 @@ import { CONFIG_OPTIONS, SERVER_IO_PROVIDER } from '../const'
 import { ConfigOptions, ServerIoProviderTypes } from '../types'
 import { suspendableSocketEvents } from './suspendable.events'
 import Logger from '../common/logger'
+import { sleep } from '../common/sleep'
```
adrastaea: [nitpick] unused import

packages/backend/src/nest/connections-manager/connections-manager.service.ts
leblowl: I'm not really sure, if we are sure that LevelDB would be totally cleared by `purgeData` then I don't think `this.localDbServer.purge()` is required - I just rearranged what was already there.

leblowl: If we close and then re-open the same LevelDB, is there data in memory that sticks around? Or does removing the LevelDB files clear everything?

leblowl: https://groups.google.com/g/leveldb/c/MXIxcn8akd4?pli=1

leblowl: From the link above "DestroyDB operation will be
faster for moderate to large databases (its implementation works
pretty much as Krzysztof Kowalczyk mentioned: it deletes all the
database files in the directory)." and then taking a look at the source code itself, looks like that is what's going on: https://github.com/google/leveldb/blob/068d5ee1a3ac40dabd00d211d5013af44be55bea/db/db_impl.cc#L1542

packages/backend/src/nest/socket/socket.service.ts
```diff
@@ -195,25 +201,72 @@ export class SocketService extends EventEmitter implements OnModuleInit {
...
+    const sockets = new Set<net.Socket>()
+
+    this.serverIoProvider.server.on('connection', conn => {
+      sockets.add(conn)
+      conn.on('close', () => {
+        sockets.delete(conn)
+      })
+    })
+
+    return () => sockets.forEach(s => s.destroy())
```
adrastaea: [design] This feels very unintuitive to me. `attachListeners()` doubling as `closeSockets()` feels like a violation of the single responsibility principle. Why can't we have a separate `closeSockets()` function? This would make the code easier to read and understand.

packages/backend/src/nest/socket/socket.service.ts
```diff
@@ -195,25 +201,72 @@ export class SocketService extends EventEmitter implements OnModuleInit {
...
+  public getConnections = (): Promise<number> => {
+    return new Promise(resolve => {
+      this.serverIoProvider.server.getConnections((err, count) => {
+        if (err) throw new Error(err.message)
+        resolve(count)
+      })
+    })
   }
 
   public listen = async (port = this.configOptions.socketIOPort): Promise<void> => {
```
adrastaea: [refactor] `port` is unused. We should either remove the parameter or use it in the function instead of hardcoding the port number.

packages/backend/src/nest/connections-manager/connections-manager.service.ts
leblowl: I just tested this and it appears to work.

packages/backend/src/nest/socket/socket.service.ts
```diff
@@ -195,25 +201,72 @@ export class SocketService extends EventEmitter implements OnModuleInit {
...
     })
+
+    // Ensure the underlying connections get closed. See:
+    // https://github.com/socketio/socket.io/issues/1602
+    //
+    // I also tried `this.serverIoProvider.io.disconnectSockets(true)`
+    // which didn't work for me.
+    const sockets = new Set<net.Socket>()
+
+    this.serverIoProvider.server.on('connection', conn => {
```
adrastaea, line 206: [bug] This seems like a confusion of the server and the socket io.

adrastaea: And in my testing never seems to be called

packages/backend/src/nest/socket/socket.service.ts
```diff
@@ -195,25 +201,72 @@ export class SocketService extends EventEmitter implements OnModuleInit {
...
+    return new Promise(resolve => {
+      this.logger(`Closing data server on port ${this.configOptions.socketIOPort}`)
+
+      if (!this.listening) {
+        this.logger('Data server is not running.')
+        resolve()
+        return
+      }
+
+      this.serverIoProvider.io.close(err => {
```
adrastaea, line 267: [question] Does this suffer from the same issue mentioned in the cleanup function stuck in `attachListeners` where you need to loop through each socket and disconnect? Also, does this remove all of the listeners that were attached in the `attachListeners` function? Could this cause a memory leak and duplication of actions if the `SocketService` is started back up after being `close` and the listeners are attached again?

packages/backend/src/nest/socket/socket.service.ts
```diff
@@ -195,25 +201,72 @@ export class SocketService extends EventEmitter implements OnModuleInit {
...
-    return await new Promise(resolve => {
-      if (this.serverIoProvider.server.listening) resolve()
+    this.logger(`Opening data server on port ${this.configOptions.socketIOPort}`)
+
+    // Sometimes socket.io closes the HTTP server but doesn't close
+    // all underlying connections. So it doesn't appear that
+    // `this.serverIoProvider.server.listening` is sufficient.
+    if (this.listening) {
+      const numConnections = await this.getConnections()
+      this.logger('Failed to listen. Connections still open:', numConnections)
```
adrastaea, line 245: [question] Wouldn't it be better to just clean up the existing connections and proceed with listening?

packages/backend/src/nest/socket/socket.service.ts
```diff
@@ -195,25 +201,72 @@ export class SocketService extends EventEmitter implements OnModuleInit {
...
+      return
+    }
+
+    return new Promise(resolve => {
       this.serverIoProvider.server.listen(this.configOptions.socketIOPort, '127.0.0.1', () => {
         this.logger(`Data server running on port ${this.configOptions.socketIOPort}`)
+        this.listening = true
```
adrastaea: [design] This seems like reinventing the wheel. I feel like this just introduces chances for desync between the state of the http server and the state of `SocketService.listening`. I'd rather see us use the built in property and properly handle the cases related to no longer listening, but not having properly cleaned up connections.

packages/backend/src/nest/socket/socket.service.ts
```diff
@@ -195,25 +201,72 @@ export class SocketService extends EventEmitter implements OnModuleInit {
...
+      }
+
+      this.serverIoProvider.io.close(err => {
         if (err) throw new Error(err.message)
+        this.logger('Data server closed')
+        this.listening = false
         resolve()
       })
+      this.logger('Disconnecting sockets')
+      this.closeSockets()
```
adrastaea, line 273: [question] Shouldn't we also detach listeners when closing sockets? I mentioned it before, but I'd rather see a dedicated `closeSockets()` and `detachListeners()` method.

adrastaea (changes requested):
I'm not entirely sure about the changes to the sagas, but I have some questions, some nitpicks, and some design reservations about the changes to the SocketService.

packages/backend/src/nest/socket/socket.service.ts
leblowl: I'm not sure. I thought about that, but there might be some trickiness with timing due to the listen and close being async. I'll have to look into it more, so just went with the simplest option for now.

packages/backend/src/nest/socket/socket.service.ts
leblowl: Yea, this doesn't appear to close all connections properly. I don't think we need to remove the listeners ever. So those just get added once in the constructor.

packages/backend/src/nest/socket/socket.service.ts
leblowl: I don't think we need to detach any listeners. They are just attached once and persist between opening and closing of the server.

packages/backend/src/nest/socket/socket.service.ts
leblowl: This gets called for me. There are two different sockets, the socket.io Socket (https://socket.io/docs/v4/server-api/#socket) and the Node `net.Socket`. I'm really just following the advice here: https://github.com/socketio/socket.io/issues/1602 and it worked for me in cleaning up the hanging connections I was seeing, so I just went with that.

packages/backend/src/nest/socket/socket.service.ts
```diff
@@ -71,7 +76,7 @@ export class SocketService extends EventEmitter implements OnModuleInit {
     this.logger('init: Frontend connected')
   }
 
-  private readonly attachListeners = (): void => {
+  private readonly attachListeners = () => {
```
adrastaea, line 77: [nitpick] lost the return type

packages/backend/src/nest/socket/socket.service.ts
```diff
@@ -44,12 +46,15 @@ export class SocketService extends EventEmitter implements OnModuleInit {
     this.readyness = new Promise<void>(resolve => {
       this.resolveReadyness = resolve
     })
+
+    this.sockets = new Set<net.Socket>()
+
+    this.attachListeners()
```
adrastaea, line 52: [FYI] attachListeners() moved from `onModuleInit()` to the `constructor`. I'm not sure what side effects this would have. Might cause `attachListeners` to trigger before `init`?

packages/backend/src/nest/socket/socket.service.ts
```diff
@@ -195,25 +200,77 @@ export class SocketService extends EventEmitter implements OnModuleInit {
...
+        return
+      }
+
+      this.serverIoProvider.io.close(err => {
         if (err) throw new Error(err.message)
+        this.logger('Data server closed')
         resolve()
       })
+
+      this.serverIoProvider.io.disconnectSockets(true)
```
adrastaea: [nitpick] I would put this in `closeSockets`, but mostly inconsequential.

adrastaea (approved):
Just nitpicks that don't necessarily need addressed. I am concerned about the move of `attachListeners()` from `onModuleInit` to the `constructor`, but I'm approving on the assumption that you have proven this doesn't cause any issues.

packages/backend/src/nest/socket/socket.service.ts
leblowl: I think not specifying void is generally the standard TS convention

packages/backend/src/nest/socket/socket.service.ts
leblowl: Yea it does trigger before `init` and I think that's fine. I think it makes sense to put everything necessary to setup an object in the constructor and only using `onModuleInit` when necessary and it doesn't seem like `attachListeners` needs to be in `onModuleInit`.

PR #2518, opened 2024-05-10
Chore(2523): Improve logging, primarily on desktop

islathehut:
* Adds a new `QuietLogger` class with improved formatting and colorization
* Use new logger in state-manager
* Use new logger in backend
* Use new logger in desktop (partial)

### Pull Request Checklist


islathehut:
Resolves #2523

packages/logger/src/index.ts
```diff
@@ -14,8 +46,92 @@ export const consoleLogger =
...
+
+export const electronLogger =
+  (packageName: string, parallelConsoleLog: boolean = false) =>
+  (module: string): QuietLogger => {
+    const name = `${packageName}:${module}`
+    nodeConsoleLogger.info(`Initializing logger ${name}`)
+    return new QuietLogger(name, parallelConsoleLog)
+  }
+
 export const logger = (packageName: string): ((arg: string) => Logger) => {
```
adrastaea: [question] Is this just left in for compatibility reasons? Why is this still the default export instead of a QuietLogger?

packages/logger/src/index.ts
```diff
@@ -14,8 +46,92 @@ export const consoleLogger =
...
   return consoleLogger(packageName)
 }
 
+export const createQuietLogger = (
```
adrastaea, line 264: [style] Can we get a doc string on this and others? I'm having trouble at a glance understanding which loggers exist for which purposes, and where the preferred entrant point are. Like `logger` is a `Logger` type and seems to be used in place of `QuietLogger` in some places. Then this `createQuietLogger` seems to wrap the `electronLogger`, but neither precludes someone from just importing the `QuietLogger` class directly and using it. I'd like to see either a narrower implementation with one point of entry, or more documentation explaining what each of these are used for and the standard way to import and use a logger in our code base.

adrastaea (commented):
I would prefer to see the rest of our `console.log` statements brought into the new paradigm and for clarity on my comments before we merge just so that we don't end up accidentally walking down the multiple logging standards path longer than we'd like.

packages/logger/src/index.ts
islathehut: Yea just for compatibility reasons.  I think some things are still using the old logger but I'll double check, I didn't want to bloat this PR with basically every file being modified at once lol.  I'll go back in and fix up all the other logs in a separate PR.  My main concern was getting backend and state manager logs fixed up.

packages/logger/src/index.ts
islathehut: Yea I can clean things up/add doc strings.  Eventually we'll only have the QuietLogger code but see my other comment about why I left in the old logger.

adrastaea (approved):

packages/common/src/polyfill.ts
```diff
@@ -0,0 +1,20 @@
+import { shouldPolyfill as shouldPolyfillCanon } from '@formatjs/intl-getcanonicallocales/should-polyfill'
```
islathehut, line 7: I'm not sure if this is the best place for this polyfill.  I wanted it to be in mobile since its specific to mobile but the mobile code wasn't being run soon enough, it seems, to run before the logging started.

packages/logger/src/index.ts
islathehut: For posterity I just went ahead and cleaned up all references.

packages/logger/src/index.ts
islathehut: For posterity I got rid of the old logger entirely.

packages/common/src/polyfill.ts
islathehut: I also added some doc strings to the this file for clarity.

PR #2525, opened 2024-05-13
Fix queue processing

islathehut:
### Pull Request Checklist

### DIFF
--- a/packages/backend/src/nest/libp2p/process-in-chunks.service.ts
+++ b/packages/backend/src/nest/libp2p/process-in-chunks.service.ts
@@ -36,7 +36,6 @@ export class ProcessInChunksService<T> extends EventEmitter {
   public updateQueue(items: T[]) {
     this.logger(`Adding ${items.length} items to the task queue`)
     items.forEach(item => this.addToTaskQueue(item))
-    this.logger(`Queue has ${this.taskQueue.length()}`)
   }
 
   private async addToTaskQueue(task: ProcessTask<T>): Promise<void>
--- a/packages/backend/src/nest/libp2p/process-in-chunks.service.ts
+++ b/packages/backend/src/nest/libp2p/process-in-chunks.service.ts
@@ -1,8 +1,8 @@
 import { EventEmitter } from 'events'
-import fastq from 'fastq'
-import type { queue, done } from 'fastq'
+import fastq, { queueAsPromised } from 'fastq'
 
 import Logger from '../common/logger'
+import { randomUUID } from 'crypto'
 
 const DEFAULT_CHUNK_SIZE = 10
 export const DEFAULT_NUM_TRIES = 2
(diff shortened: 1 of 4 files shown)


packages/backend/src/nest/libp2p/process-in-chunks.service.ts
```diff
@@ -27,59 +28,92 @@ export class ProcessInChunksService<T> extends EventEmitter {
...
-        this.data.delete(item)
-        try {
-          this.taskQueue.push({ data: item, tries: 0 } as ProcessTask<T>)
-        } catch (e) {
-          this.logger.error(`Error occurred while adding new task for item ${item} to the queue`, e)
-          this.data.add(item)
-        }
+  private async addToTaskQueue(task: ProcessTask<T>): Promise<void>
+  private async addToTaskQueue(item: T): Promise<void>
+  private async addToTaskQueue(itemOrTask: T | ProcessTask<T>): Promise<void> {
```
leblowl, line 43: Do you know the purpose of adding multiple method signatures in this case? Doesn't the final signature cover the two above it?

packages/backend/src/nest/libp2p/process-in-chunks.service.ts
```diff
@@ -27,59 +28,92 @@ export class ProcessInChunksService<T> extends EventEmitter {
...
   }
 
-  public updateData(items: T[]) {
-    this.logger(`Updating data with ${items.length} items`)
-    this.taskQueue.pause()
-    items.forEach(item => this.data.add(item))
-    this.addToTaskQueue()
+  public updateQueue(items: T[]) {
+    this.logger(`Adding ${items.length} items to the task queue`)
+    items.forEach(item => this.addToTaskQueue(item))
```
leblowl, line 38: It might be nice to log taskQueue.length() each time we add to the task queue to surface any issues with the queue getting backed up

packages/backend/src/nest/libp2p/process-in-chunks.service.ts
```diff
@@ -27,59 +28,92 @@ export class ProcessInChunksService<T> extends EventEmitter {
...
+      this.deadLetterQueue.push(task)
+      return
+    }
+
+    this.logger(`Adding task ${task.taskId} with data ${task.data} to the task queue`)
+    try {
+      const success = await this.taskQueue.push(task)
+      if (!success) {
+        this.logger(`Will try to re-attempt task ${task.taskId} with data ${task.data}`)
+        await this.taskQueue.push({ ...task, tries: task.tries + 1 })
```
leblowl: Do we want to only retry once?

leblowl (approved):
Looks good, thanks!

packages/backend/src/nest/libp2p/process-in-chunks.service.ts
islathehut: It's just javascript magic, that way you should get a different signature when using autocomplete depending on the type of your input.

packages/backend/src/nest/libp2p/process-in-chunks.service.ts
islathehut: Good idea, it wouldn't give an accurate picture if things are moving swiftly but would cover us if there's a bottleneck.

packages/backend/src/nest/libp2p/process-in-chunks.service.ts
islathehut: Ideally we would retry differently depending on the error but figuring out all of the errors that are possible with libp2p that amount to "we couldn't connect because we haven't been able to yet" or "this peer is offline and we won't connect right now no matter what" to filter those out and only retry multiple times when its something that would keep us from adding the peer to the peerstore and auto-dialing.

I don't think it's worth doing that in this PR but we should be doing it.

packages/backend/src/nest/libp2p/process-in-chunks.service.ts
islathehut: Done

PR #2542, opened 2024-05-17
Improve connection speed/reliability and add more logging around connections

islathehut:
This should make connecting to new peers, especially when opening the app or resuming the app (iOS), noticeably faster with fewer errors when connecting to peers that are online.

* Wait to dial peers until we know tor is fully bootstrapped
* Deduplicate tasks in processInChunksService to avoid clutter
* Reduce the number of parallel dials on libp2p
* Update some timeouts/ping intervals and use `doPX` on libp2p connection manager (PX allows peers to share peer information with each other)
* Add more logging internally and on libp2p packages

### Pull Request Checklist

### DIFF
--- a/packages/backend/src/nest/libp2p/libp2p.service.ts
+++ b/packages/backend/src/nest/libp2p/libp2p.service.ts
@@ -41,13 +41,22 @@ export class Libp2pService extends EventEmitter {
     super()
   }
 
-  private dialPeer = async (peerAddress: string) => {
-    if (this.dialedPeers.has(peerAddress)) {
+  private dialPeer = async (peerAddress: string): Promise<boolean> => {
+    const ma = multiaddr(peerAddress)
+    const peerId = peerIdFromString(ma.getPeerId()!)
+
+    const libp2pHasPeer = await this.libp2pInstance?.peerStore.has(peerId as any)
+    const weHaveDialedPeer = this.dialedPeers.has(peerAddress)
+
+    if (weHaveDialedPeer || libp2pHasPeer) {
       this.logger(`Skipping dial of ${peerAddress} because its already been dialed`)
-      return
+      return true
     }
     this.dialedPeers.add(peerAddress)
-    await this.libp2pInstance?.dial(multiaddr(peerAddress))
+    const connection = await this.libp2pInstance?.dial(multiaddr(peerAddress))
+
+    if (connection) return true
+    return false
   }
 
   public getCurrentPeerInfo = (): Libp2pPeerInfo => {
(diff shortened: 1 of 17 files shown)


packages/backend/src/nest/libp2p/libp2p.service.ts
```diff
@@ -153,21 +162,22 @@ export class Libp2pService extends EventEmitter {
...
       libp2p = await createLibp2p({
         start: false,
         connectionManager: {
-          minConnections: 3, // TODO: increase?
+          minConnections: 5, // TODO: increase?
           maxConnections: 20, // TODO: increase?
-          dialTimeout: 120_000,
-          maxParallelDials: 10,
+          dialTimeout: 120000,
+          maxParallelDials,
```
holmesworcester, line 180: What does this set maxParallelDials to? We observed a while back that too many dials or too many connections could create a performance problem, perhaps only on mobile. I think the performance problem was with Tor if I remember correctly. Maybe that's been fixed in Tor and it's not a problem anymore but we should check to make sure.

packages/backend/src/nest/libp2p/process-in-chunks.service.ts
```diff
@@ -2,7 +2,7 @@ import { EventEmitter } from 'events'
 import fastq, { queueAsPromised } from 'fastq'
 
 import Logger from '../common/logger'
-import { randomUUID } from 'crypto'
+import CryptoJS from 'crypto-js'
```
holmesworcester, line 5: This says it's deprecated and not maintained. I think that's a barrier to us using it unless there's something I'm missing? https://www.npmjs.com/package/crypto-js

packages/backend/src/nest/libp2p/libp2p.service.ts
islathehut: This PR lowers the max parallel dials from 10 to 2 because it seems to cause issues with connections.

packages/backend/src/nest/libp2p/process-in-chunks.service.ts
islathehut: It was already in the project and I'm only using it to generate a consistent shared ID via hashing, no security threat or anything.

holmesworcester: And in the other place it's not used for anything security related either?

islathehut: That I don't know, I haven't checked the usages

Issue #2302, opened 2024-02-17
Language is not supported (Swedish)


matstage:
Hi, 

I tried to create a channel using the Swedish word for meeting (Möte). When creating the channel, I get the error message: "Name cannot begin with whitespaces or dashes"!

Replacing the ö with an o creates the channel successfully.

![Quiet_Language](https://github.com/TryQuiet/quiet/assets/949035/1f069d48-5366-4464-8fec-0c98ef7c4efd)

Version tested on are 2.1.0 and 2.1.1!

holmesworcester:
Right now channel names are validated just as strictly as usernames. We don't need to do this, I don't think, so the solution would be to allow special characters in channel names.

leblowl:
Also the error message is misleading here. @richardtorres314, tagging you in case you find this interesting.

richardtorres314:
@holmesworcester Will the same apply to community names (allowing of special characters)? What are we classifying as special characters? Lastly, we set the username to lowercase; will the same apply both community and channel names? I'll have a PR open this week for this issue. Thank you!

holmesworcester:
@richardtorres314 apologies for the slow reply!

> Will the same apply to community names (allowing of special characters)?

Yes. There will be no attempt at a uniqueness guarantee for community names so it makes no sense AFAICT to restrict the choice of community name.

> What are we classifying as special characters? 

For usernames, anything *not* a-z, 0-9.

> Lastly, we set the username to lowercase; will the same apply both community and channel names?

The reason for enforcing all lowercase for usernames is to limit impersonation, so we do not need to do this for community and channel names since those can be duplicates.

> I'll have a PR open this week for this issue. Thank you!

Thank **you**!

PR #2584, opened 2024-07-31
fix: channel and community naming #2302

richardtorres314:
### Pull Request Checklist


packages/desktop/src/rtl-tests/channel.add.test.tsx
```diff
@@ -90,7 +90,7 @@ describe('Add new channel', () => {
     const alice = await factory.create<ReturnType<typeof identity.actions.addNewIdentity>['payload']>('Identity', {
       nickname: 'alice',
     })
-    const channelName = { input: 'my-Super Channel ', output: 'my-super-channel-' }
+    const channelName = { input: 'my-Super Channel ', output: 'my-Super Channel ' }
```
adrastaea, line 93: I don't think that allowing spaces in channel names makes sense for a couple reasons. It's impossible to tell how many spaces are in a channel name when referencing it. It looks like a bug in the UI when a channel name starts with white spaces. Also, the industry standard is to not allow capitals or spaces in channel names.

packages/desktop/src/renderer/forms/fields/communityFields.ts
```diff
@@ -15,10 +15,6 @@ export const communityNameField = (name = 'name'): FieldData => {
         value: 20,
         message: CommunityNameErrors.NameTooLong,
       },
-      pattern: {
```
adrastaea, line 18: I think it's worth still disallowing leading and trailing whitespace for the sake of aesthetics. There are also options for regexing to include diacritical marks if we want to continue to disallow special characters, but support international alphabets.

adrastaea (changes requested):
I have stylistic qualms with this change. Mostly related to allowing whitespaces which I think make for a bad experience telling apart channels, typing references to them, and  displaying them in the UI. I think fully pulling out pattern matching for channels and communities is the wrong move. I found [this article](https://brettrawlins.com/blog/regular-expression-for-international-names/) for supporting international characters which might be worth testing, but I also don't know if this is something we want to support outside of the community name for the sake of users easily referencing channels. I think that in certain resolutions it may be hard to tell certain diacritical marks apart which would make it difficult to search or reference them. @holmesworcester could overrule me there though

PR #2601, opened 2024-09-11
Update Publishing Instructions

adrastaea:
Adds more detail to publishing instructions to clear up some confusion about process.

### DIFF
--- a/PUBLISHING.md
+++ b/PUBLISHING.md
@@ -43,11 +43,10 @@
 1. QA tests the production release.
 1. Production release is published.
 
-## Checklist for Alpha Release
+## Checklist Before Alpha Release
 
 - [ ] Release branch is created from `develop` branch with the production version number, e.g. `2.1.0`.
 - [ ] Reviewed the base `CHANGELOG.md` file and ensured that it is up to date with all changes included in the release since the last production release. Package level `CHANGELOG.md` files are automatically updated during the release process.
-- [ ] Approved the updated `CHANGELOG.md` file with the @holmesworcester.
 - [ ] Reviewed the [Quiet Planning Board](https://github.com/orgs/TryQuiet/projects/3) and ensured all issues contained in the release candidate are in the `Ready for QA` column.
 
 ## Preparing a Release Candidate (Alpha)
@@ -60,7 +59,7 @@ Alpha releases are pre-release versions of the release which are delivered to QA
 1. Promote the alpha release on the [Google Play Console](https://play.google.com/console/) to a closed testing track. Contact @holmesworcester if you need access to the organization.
 1. Notify QA that the alpha release is ready for testing.
 
-## Checklist for Production Release
+## Checklist Before Production Release
 
 - [ ] Build is working correctly, passes automated tests and self-QA
 - [ ] Alpha was delivered for QA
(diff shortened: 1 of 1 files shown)


PUBLISHING.md
```diff
@@ -4,87 +4,99 @@
...
 
-## Pre-release checklist
 - [ ] Build is working correctly, passes automated tests and self-QA
-- [ ] Release candidate is delivered for QA
+- [ ] Alpha was delivered for QA
 - [ ] Sprint column is free from QA reported blocking issues
 - [ ] QA approved the release
-- [ ] All hotfixes to previous releases have been merged into the release (and develop) branch
-- [ ] CHANGELOG.md is up to date
+- [ ] All hotfixes for issues discovered in alpha releases have been merged into the release (and develop) branch
```
holmesworcester, line 68: I don't think this line is equivalent to the one that it replaces. In some cases we'll make full releases (not alpha releases) that include a hotfix to a previous release. We then need to make sure these hotfixes are brought over to the new release because it's possible someone forgot to add them to both the hotfix branch and develop at the time they were created. Does this make sense?

PUBLISHING.md
```diff
@@ -4,87 +4,99 @@
...
+1. Trigger a production release with `GH_TOKEN=<token> npm run publish --release=[patch|minor|major|<EXACT_VERSION>]` (e.g. to create a prerelease of a minor update `npm run publish --release=minor`) This should remove the `alpha.x` suffix, and create a release on the [Releases Page](https://github.com/TryQuiet/quiet/releases), and trigger Github Actions to deploy the release to the Google Play Store and App Store. If the semantic version is different from the alpha release, select no in the prompt to create new versions, and call the command again with the exact version.
+1. Manually update the release notes on the [Releases Page](https://github.com/TryQuiet/quiet/releases) with the changes included in the release. See [RELEASE_NOTES_GUIDE.md](RELEASE_NOTES_GUIDE.md) for guidance on writing release notes.
+1. Promote the release on the [Google Play Console](https://play.google.com/console/) to a production track.
+1. Promote the release on the [App Store Connect](https://appstoreconnect.apple.com/).
 
 ## Post-release checklist (production)
-- [ ] Download links are updated on a website and in README.md
+
+- [ ] Download links are updated on website and in README.md
```
holmesworcester: I think we can cut "and in README.md" because that download link now links to the downloads section on the website.

holmesworcester (changes requested):
I found one spot (I think) where a note intended as a guardrail wasn't carried over, so it's worth checking to make sure I didn't miss any others. But this looks good overall and thanks!

PUBLISHING.md
```diff
@@ -4,87 +4,99 @@
...
+1. If QA finds no issues, the release is approved.
+1. A production release is built.
+1. QA tests the production release.
+1. Production release is published.
+
+## Checklist for Alpha Release
+
+- [ ] Release branch is created from `develop` branch with the production version number, e.g. `2.1.0`.
+- [ ] Reviewed the base `CHANGELOG.md` file and ensured that it is up to date with all changes included in the release since the last production release. Package level `CHANGELOG.md` files are automatically updated during the release process.
+- [ ] Approved the updated `CHANGELOG.md` file with the @holmesworcester.
```
holmesworcester: We can cut this since it repeats later. 

The main idea here is that CHANGELOG is a public facing document that we want to be clear, readable, and meaningful, and that conforms more or less to the style described in publishing guidelines.

holmesworcester (changes requested):

PUBLISHING.md
adrastaea: I don't think that entirely makes sense as a rule because there may be cases where hotfixes applied to a previous release are solved in more robust way through a refactor in the develop branch, and in that case, it wouldn't make sense to cherry pick the hotfix applied to an old release. To me, it seems better to just focus on ensuring that at the time of a release (alpha or production) merges the appropriate commits from the release branch to `develop`. Otherwise, there becomes this question of how many previous releases back will we comb through every commit in that release branch to check if it was cherry picked to `develop` or not, and if that change should be applied to the current release or if it was fixed in another way.

PUBLISHING.md
adrastaea: I'm just going to remove the approval part from the alpha release section because I think its valuable for QA to have an updated changelog for the alpha too

holmesworcester (approved):

Issue #2631, opened 2024-10-08
Encrypt messages with team key


adrastaea:
## Description

As a first step before having tiered message encryption, we should encrypt every message with the team key. In storage.service:sendMessage, encrypt the message before adding it to the channel db.

## Acceptance Criteria

1. Encryption mechanism / libraries pass review by e.g. Alishah or some other expert
2. Message model includes key name and generation
3. Messages get encrypted before being added to channel db
4. ~~(maybe) Differentiate public and private channels, keep the publicChannelsRepo infrastructure as a way to send unencrypted messages, and add a privateChannelsRepo which gets encryption~~ Just use the team key
5. (maybe, up to implementer) Pull channels functions out of StorageService, create a Channels module, and move all the message sending, receiving, encrypting, and decrypting logic there

The key question on 4 is: can we do this in a very quick and minimal way that is also clean and likely to not introduce bugs? If so, we should do that. If not, we should do as much as necessary to make sure it's going to be clean and bug free.

holmesworcester:
Note: have a look at how background notifications are working on Android before tackling this.

kingalg:
Version: 4.0.0-alpha.8

Done

PR #2713, opened 2025-01-27
chore(2631): Refactor storage service to separate channel and message logic

islathehut:
### Pull Request Checklist


islathehut:
#2631

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -706,6 +707,7 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
 
     const community = {
       id: payload.id,
+      name: payload.name,
```
islathehut: @adrastaea This is all you need to do to fix the chain loading

packages/backend/src/nest/storage/channels/channel.store.ts
```diff
@@ -0,0 +1,288 @@
...
+   * @param message Message to add to the OrbitDB database
+   * @returns Hash of the new database entry
+   */
+  public async addEntry(message: ChannelMessage): Promise<string> {
+    this.logger.info('Adding message to database')
+    const encryptedMessage = await this.messagesService.onSend(message)
+    try {
+      return await this.getStore().add(encryptedMessage)
+    } catch (e) {
+      throw new CompoundError(`Could not append message (entry not allowed to write to the log)`, e)
```
adrastaea, line 217: This may be existing behavior, but when we hit this error, the message is still displayed in the timeline (grayed out as you might expect for an unsent message, but when you close and reopen the app, it is no longer grayed out which is misleading as to whether it was actually sent)

adrastaea (approved):
One nitpick of user facing behavior, but may be out of scope for this pr

packages/backend/src/nest/storage/channels/channel.store.ts
islathehut: Yea that's existing behavior we need to deal with.  We can't know if a message has been sent or not right now, as far as I know.

Issue #568, opened 2022-06-03
User should be able to right click to copy image (and text)


holmesworcester:
Right now there's no way to get an image out of Quiet. You can't save it, You can't copy it. 

Letting users right click to copy would be awesome. This looks like a reasonable way to achieve this in Electron, and it seems like it will enable copying text too: https://github.com/sindresorhus/electron-context-menu

See also https://stackoverflow.com/questions/63379503/electron-js-with-the-right-click-be-able-to-perform-a-copy-and-cut-if-a-sel

agiledev24:
This issue is solved with my last PR - https://github.com/TryQuiet/quiet/pull/2355
@holmesworcester

holmesworcester:
@agiledev24 thank you for your work on this! I just fixed the issue with tests and your PR is now awaiting review!

PR #2719, opened 2025-02-01
feat(503): Adds desktop context menu

holmesworcester:
Adds a context menu in Quiet desktop for copying text ([#503](https://github.com/TryQuiet/quiet/issues/503)) and saving images ([#568](https://github.com/TryQuiet/quiet/issues/568))

This adds a small tweak from an earlier PR (https://github.com/TryQuiet/quiet/pull/2355) by @agiledev24 to keep the new change from breaking unit tests. 

Thank you @agiledev24 for your contribution!

### Pull Request Checklist


holmesworcester:
@islathehut -- this fails dependency review, but the dependency seems to be used by a lot of electron-related packages, not the one added by this PR.  how do you suggest I proceed?

Also, it looked like it could be tricky to include the context-menu in the unit tests, and that the value could be limited, since this seems like a "set and forget" feature that will not interact much with other features--at least not until we start using the context menu ourselves for our own particular functions. 

It also seemed like this was not the best place to learn about mocks and tests, but maybe I'm wrong about that.

What do you think?  Want to make me write a test for this?

islathehut:
> @islathehut -- this fails dependency review, but the dependency seems to be used by a lot of electron-related packages, not the one added by this PR. how do you suggest I proceed?
> 
> Also, it looked like it could be tricky to include the context-menu in the unit tests, and that the value could be limited, since this seems like a "set and forget" feature that will not interact much with other features--at least not until we start using the context menu ourselves for our own particular functions.
> 
> It also seemed like this was not the best place to learn about mocks and tests, but maybe I'm wrong about that.
> 
> What do you think? Want to make me write a test for this?

@holmesworcester I wouldn't worry about the dependency review, especially if its on electron since we're not gonna be upgrading that right now.

For tests this may be a good candidate for e2e tests.

holmesworcester:
I looked into making an e2e test for this, but it isn't a great candidate, I don't think. Selenium could right click on an element, but then it could not see or manipulate the context menu, probably because it is in the Electron chrome and not in the page itself. 

There are some articles (e.g. https://elementalselenium.com/tips/63-right-click) that say you can manipulate the context menu with arrow keys and perhaps use keys for copy/paste, but this didn't work either.

I did learn about e2e tests, get the linux ones running locally, and update `e2e-tests/README.md` with much better instructions, though. I'm going to move this back to "waiting for review".

Issue #505, opened 2022-05-13
Sticky date markers to make conversations cleaner and more readable (desktop)


holmesworcester:
Nearly every chat app uses "sticky" date markers that float above the conversation view, aligned so that the date markers in the scrolling text pass underneath them. 

Open whatsapp or slack and look at the date markers for an example of how this works. 

This makes the date more visible, and it also means you don't need to redundantly display date information on each message. 

We should add sticky date markers. We should remove date from the timestamp on each message. We should also add a tooltip that appears when clicking a timestamp that shows the full date and time.

There's another subtlety here where we should probably use "Yesterday" instead of a date, and we may want to use days of week for dates within the year, and years for dates in previous years. WhatsApp I believe does days of the week for the past seven days and then switches to a mm/dd/yyyy date.

holmesworcester:
@jgaylor if you feel like we should tackle this now for cleanliness and readability going forward, here's the ticket. I'm a bit torn on whether this is a priority or not.

kingalg:
@holmesworcester I believe that we have dates markers established already. Do you want to change anything in this area or can we close it?

jgaylor:
Latest [Figma](https://www.figma.com/file/ImDxrLLEBMH9EL383FzfD3/Sticky-date-markers?type=design&node-id=13%3A37&mode=design&t=7Rnqm5YY3kyXLh16-1) & [Loom](https://www.loom.com/share/1e35c6541eb04882bbc9951a5fc03850) (from last week, I don't believe you had any feedback)

holmesworcester:
@trahantet thanks for taking an initial look at this! 

There ended up being quite a lot to it, and testing it required some changes to our tests, so it ended up not being a great first issue. Let me know if we can pick another for you!

Also, a note for QA: messages still show the date, unfortunately, because changing this requires some refactoring which I had a question about, so I created #2745.

kingalg:
Version: 5.1.0-alpha.0
Done.

PR #2731, opened 2025-02-06
feat: add sticky date markers #505

holmesworcester:
This addresses [#505](https://github.com/TryQuiet/quiet/issues/505) to show sticky, floating date markers. 

### Pull Request Checklist


adrastaea:
Screenshot 2025-02-12 at 3 34 25 PM
One style nitpick, after scrolling all the way up, the floating date marker does not line up with the static date marker which looks a little janky until the floating marker disappears. I think that aligning the margins would make it look more seamless

packages/desktop/package.json
```diff
@@ -147,7 +145,9 @@
     "react-markdown": "^6.0.3",
     "react-qr-code": "^2.0.11",
     "remark-gfm": "^1.0.0",
-    "socket.io-client": "^4.7.4"
+    "socket.io-client": "^4.7.4",
+    "cypress": "^12.17.4",
```
adrastaea: This should be a dev dependency since it's not required for prod operation

packages/desktop/src/renderer/components/Channel/Channel.cy.tsx
```diff
@@ -0,0 +1,181 @@
+import React from 'react'
+import CssBaseline from '@mui/material/CssBaseline'
+import { composeStories, setGlobalConfig } from '@storybook/testing-react'
+import { it, beforeEach, cy, Cypress, describe } from 'local-cypress'
+
+import * as stories from './Channel.stories'
+import { withTheme } from '../../storybook/decorators'
+import { mount } from 'cypress/react18'
+
+declare global {
```
adrastaea, line 10: This violates our eslint config. I think the more modern way to accomplish this is to create a type declaration file that typescript will pick up at `cypress/support/index.d.ts`. I found this [section in the Cypress docs](https://docs.cypress.io/app/tooling/IDE-integration#Triple-slash-directives) that talks about adding type definitions for custom commands. They also suggest defining custom commands in the `commands.ts` file.

packages/desktop/src/renderer/components/Channel/Channel.cy.tsx
```diff
@@ -0,0 +1,181 @@
...
+    })
+
+    it('should display the correct date text', () => {
+      cy.get(channelContent)
+      cy.get(messageInput).focus().type('{pageup}')
+      
+      cy.get(floatingDateSelector)
+        .should('be.visible')
+        .invoke('text')
+        .should('contain', '28 Oct')
```
adrastaea: Is there a way we can avoid hardcoding this, and derive the expected value from the test data?

packages/desktop/src/renderer/components/Channel/Channel.stories.tsx
```diff
@@ -1,48 +1,98 @@
...
+      pkcs10: 'fakePkcs10',
     },
-    userCertificate: 'userCertificate',
-    joinTimestamp: null,
   },
+  userCertificate: 'fakeCertificate',
+  joinTimestamp: null,
+}
+
+// Replace "ModalName.uploadedFileModal" etc. with the REAL enum/constant your store uses
```
adrastaea: Let's close up this TODO before merging. These are at `/desktop/src/renderer/sagas/modals/modals.types.ts`

packages/desktop/src/renderer/components/Channel/Channel.stories.tsx
```diff
@@ -602,3 +656,73 @@ const component: ComponentMeta<typeof ChannelComponent> = {
...
+          type: 'Modals/openModal',
+          payload: { name: ModalName.duplicatedUsernameModal },
+        })}
+        unregisteredUsernameModalHandleOpen={() => ({
+          type: 'Modals/openModal',
+          payload: { name: ModalName.unregisteredUsernameModal },
+        })}
+        pendingGeneralChannelRecreation={false}
+        removeFile={dummyRemoveFile}
+        {...args}
```
adrastaea, line 710: I don't think you're using the spread operator correctly here. I think you're trying to use the `args` partial you defined earlier, but you've ended up needing to hardcode a lot of the values included in that partial because you have put them next to args like
```
{...args}
otherVal='otherVal'
```
 instead of using your hardcoded values to add new keys or replace keys in args like 
 ```
 {
     ...args,
     otherVal: 'otherVal'
 }
 ```
 I think this would work better if you used the `Template` pattern used by the other stories.

packages/desktop/src/renderer/components/Channel/Channel.stories.tsx
```diff
@@ -60,23 +110,27 @@ const args: Partial<ChannelComponentProps & UploadFilesPreviewsProps> = {
...
   )
 }
 
-// States
-export const Normal = Template.bind({})
+export const Component = Template.bind({})
 export const Pending = Template.bind({})
 
-Normal.args = args
+Component.args = args
```
adrastaea: Normal seems more descriptive as a state descriptor than Component.

packages/desktop/src/renderer/components/Channel/Channel.stories.tsx
```diff
@@ -602,3 +656,73 @@ const component: ComponentMeta<typeof ChannelComponent> = {
...
 
 export default component
+
+export const InteractiveLocalState: ComponentStory<typeof ChannelComponent> = () => {
+  const [localMessages, setLocalMessages] = useState<{
+    count: number
+    groups: { [day: string]: DisplayableMessage[][] }
+  }>(mock_messages())
+
+  const handleSend = (message: string) => {
```
adrastaea, line 680: The new message isn't ordered properly when sent in this interactive story. Also, I don't have a real Quiet instance with old messages, but in this interactive UI, scrolling above the top briefly shifts the floating date marker from `Oct 26` to `Today`

https://github.com/user-attachments/assets/31127e73-7a19-4f4d-9aef-ab1bc103226d

packages/desktop/src/renderer/components/widgets/channels/ChannelMessages.tsx
```diff
@@ -96,72 +101,153 @@ export const ChannelMessagesComponent: React.FC<IChannelMessagesProps & FileActi
...
+            })}
+          </div>
+        ))}
       </List>
-    </Root>
+    </StyledRoot>
   )
 }
 
+export { FETCHING_MESSAGES as fetchingChannelMessagesText, DELETING_CHANNEL as deletingChannelMessage }
```
adrastaea: Renaming constants in order to export them feels like an anti-pattern. We should just commit to renaming them with the all caps pattern everywhere. If you use F2 to rename a variable in your IDE, it should rename it everywhere automatically, so you don't have to do hacks like this to fix import problems.

packages/desktop/src/shared/setupTests.ts
```diff
@@ -38,27 +38,36 @@ jest.mock('electron-store', () => {
...
+      isFocused: () => boolean
+      show: jest.Mock
+    }>
   }
+  [Symbol.iterator]: () => Generator<number, void, unknown>
+}
 
-  // @ts-expect-error - expression of type 'unique symbol' can't be used to index type
-  mock[Symbol.iterator] = function* () {
+// Then define your mock like this:
```
adrastaea:
```suggestion

```

packages/state-manager/src/sagas/publicChannels/publicChannels.selectors.test.ts
```diff
@@ -272,7 +272,7 @@ describe('publicChannelsSelectors', () => {
 
     // Get groups names
     const groupDay1 = formatMessageDisplayDay(formatMessageDisplayDate(msgs['7'].createdAt))
-    expect(groupDay1).toBe('Feb 05')
+    expect(groupDay1).toBe('Feb 5')
```
adrastaea, line 275: Another place we could be using references to the test data instead of hardcoding

adrastaea (changes requested):

packages/desktop/src/renderer/components/Channel/Channel.cy.tsx
holmesworcester: `*.cy.tsx` was in `.eslintignore`. Claude says the right way to do this is to have a separate tsconfig for Cypress, and the Cypress docs mention something similar: https://docs.cypress.io/app/tooling/typescript-support#Clashing-types-with-Jest. I'll try this and see if I can get it to work without errors.

holmesworcester: Resolved in 5753d7f7. linter works for Cypress files.

packages/desktop/src/renderer/components/Channel/Channel.cy.tsx
holmesworcester: There's an easy solution to this, but doing so makes me a bit uncomfortable, since if we derive the date format we might introduce a bug that changes the formatter in both production and in the test. Also, hardcoded test data makes it very clear to anyone looking at this what date format is expected.

(There's another issue with this test where the test should be "equals" and not "contains" (since if the date also displayed the year, incorrectly, this test would still pass). I fixed that in the other PR I'm working on for making dates match the design.)

Are you okay with leaving it?

packages/desktop/src/renderer/components/Channel/Channel.stories.tsx
holmesworcester: Resolved in cd9475f6

packages/desktop/src/renderer/components/Channel/Channel.stories.tsx
holmesworcester: Fixed ordering! I saw that and was being lazy because it didn't really affect what I was testing, but it's weird and definitely good to fix. 

On the "Today" flashing on transitions, I was not able to reproduce what you were seeing in Storybook, so I assume the issue was happening in the app too and I just wasn't seeing it there either. I found what I think was the culprit and fixed it. Let me know if you still see it!

packages/desktop/src/renderer/components/Channel/Channel.stories.tsx
holmesworcester: I think the reason we can't use the Template pattern here is because in this story for the interactivity we need component state (UseState) but Args in Storybook are static. (I don't totally understand this stuff yet.)

holmesworcester: Is this better though? Fewer hardcoded values now?

packages/state-manager/src/sagas/publicChannels/publicChannels.selectors.test.ts
holmesworcester: This is easy to fix but I have the same question/doubt as before.

packages/desktop/src/renderer/components/Channel/Channel.cy.tsx
adrastaea: I'm okay with leaving it. I see your point of making it obvious to anyone reading the test what the value should be, which is more obvious with the string than with a const.

tsconfig.json
```diff
@@ -2,6 +2,7 @@
     "compilerOptions": {
       "target": "ES2020",
       "baseUrl": ".",
+      "outDir": "./dist",
```
adrastaea: The outDir defined by the packages is `./lib` not `./dist`. It's probably fine to include an outDir in the base config file to avoid the linter problem, but we should change this value to match the standard value that the packages tsconfig overrides this value with

packages/desktop/src/renderer/components/widgets/channels/ChannelMessages.tsx
```diff
@@ -1,22 +1,35 @@
...
-
 import SpinnerLoader from '../../ui/Spinner/SpinnerLoader'
 
-import { DownloadStatus, MessagesDailyGroups, MessageSendingStatus } from '@quiet/state-manager'
-
+import {
+  DownloadStatus,
+  MessagesDailyGroups,
+  MessageSendingStatus,
+  FileMetadata,
```
adrastaea: Worth noting that this is the deprecated type. `@quiet/types:files.ts` is the actual FileMetadata type that is commonly used. You should check if any of these types imported from `@quiet/state-manager` are defined the common types package and use those instead. This is fixed in my feature branch, but if any of your feature depends on those old types it may be worth fixing now too.

package.json
```diff
@@ -28,7 +28,7 @@
     "husky": "^9.0.11",
     "lerna": "^6.6.2",
     "pnpm": "^9.12.1",
-    "typescript": "^4.9.5"
+    "typescript": "^4.9.3"
```
adrastaea: Is this downgrade intentional?

adrastaea (changes requested):

adrastaea (approved):

packages/desktop/src/renderer/components/widgets/channels/ChannelMessages.tsx
holmesworcester: Updated these in a few places.

adrastaea (approved):
Looks good. Best to remove the extraneous type definition files, but dropping my approval in so you can merge.

packages/@types/3rd-party-auth/index.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '3rd-party-auth' {
```
adrastaea: remove

packages/@types/rtl-test-factory/index.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'rtl-test-factory' {
```
adrastaea: remove

PR #2733, opened 2025-02-06
feat(2631): Add e2e encryption/decryption to messages

### LINKED ISSUE #2631: Encrypt messages with team key

### LINKED ISSUE #2632: Decrypt messages with team key
## Description

We already store messages in the redux store after receiving them from OrbitDB, so until we prioritize work to move from the redux store to a better local storage solution we should just decrypt during the replicate and replicate.progress events in storage.service:subscribeToChannel and pass the decrypted message to the redux store. We can maintain a queue of messages that failed to decrypt in LevelDB until we have a better solution.

## Acceptance Criteria

1. Messages are decrypted in backend when replicated from OrbitDB and passed to Redux
2. Messages which fail to decrypt are added to a queue and checked when the lockboxes a user can access change
6. < 20% increase in message delivery time (from sending until user sees message)
7. < 10% increase in time to view recent messages (joining community) 
8. Android notifications display with message contents, as they do now (message logic and decryption must be in backend)
9. All versions work well enough to do a successful demo

islathehut:
### Pull Request Checklist

### DIFF
--- a/packages/backend/src/nest/storage/channels/channel.store.ts
+++ b/packages/backend/src/nest/storage/channels/channel.store.ts
@@ -104,6 +104,10 @@ export class ChannelStore extends EventStoreBase<EncryptedMessage, ConsumedChann
       this.logger.info(`${this.channelData.id} database updated`, entry.hash, entry.payload.value?.channelId)
 
       const message = await this.messagesService.onConsume(entry.payload.value!)
+      if (message == null) {
+        this.logger.error(`Couldn't consume message ${entry.payload.value!.id}`)
+        return
+      }
 
       this.emit(StorageEvents.MESSAGES_STORED, {
         messages: [message],
@@ -243,6 +247,9 @@ export class ChannelStore extends EventStoreBase<EncryptedMessage, ConsumedChann
         // NOTE: we skipped the verification process when reading many messages in the previous version
         // so I'm skipping it here - is that really the correct behavior?
         const decryptedMessage = await this.messagesService.onConsume(x.value)
+        if (decryptedMessage == null) {
+          continue
+        }
         messages.push(decryptedMessage)
       }
     }
--- a/packages/backend/src/nest/storage/channels/channel.store.ts
+++ b/packages/backend/src/nest/storage/channels/channel.store.ts
@@ -231,7 +231,7 @@ export class ChannelStore extends EventStoreBase<EncryptedMessage, ConsumedChann
   }
 
   /**
-   * Read a list of entries on the OrbitDB event store
+   * Read a list of entries on the OrbitDB event store and decrypt
    *
    * @param ids Optional list of message IDs to filter by
    * @returns All matching entries on the event store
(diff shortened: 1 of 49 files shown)


islathehut:
#2631 
#2632

packages/backend/src/nest/storage/channels/channel.store.ts
```diff
@@ -95,10 +100,14 @@ export class ChannelStore extends EventStoreBase<ChannelMessage> {
...
     this._subscribing = true
 
-    this.getStore().events.on('update', async (entry: LogEntry<ChannelMessage>) => {
+    this.getStore().events.on('update', async (entry: LogEntry<EncryptedMessage>) => {
       this.logger.info(`${this.channelData.id} database updated`, entry.hash, entry.payload.value?.channelId)
 
       const message = await this.messagesService.onConsume(entry.payload.value!)
+      if (message == null) {
+        this.logger.error(`Couldn't consume message ${entry.payload.value!.id}`)
+        return
```
adrastaea, line 109: I think this could lead to messages being dropped from the local redux store permanently if the peer receives a message encrypted with a version of the team key that they haven't yet added to their sigchain. Do we need some method of caching messages that we fail to decrypt, and then retrying their decryption when a new team key is added to the sigchain maybe?

adrastaea: Unless there's some other place in the code I haven't looked at yet where we pull all the messages in the store on some regular basis that is.

packages/backend/src/nest/storage/channels/channel.store.ts
```diff
@@ -224,18 +236,21 @@ export class ChannelStore extends EventStoreBase<ChannelMessage> {
...
-    const messages: ChannelMessage[] = []
+    const messages: ConsumedChannelMessage[] = []
 
     for await (const x of this.getStore().iterator()) {
       if (ids == null || ids?.includes(x.value.id)) {
         // NOTE: we skipped the verification process when reading many messages in the previous version
         // so I'm skipping it here - is that really the correct behavior?
-        const processedMessage = await this.messagesService.onConsume(x.value, false)
-        messages.push(processedMessage)
+        const decryptedMessage = await this.messagesService.onConsume(x.value)
```
adrastaea, line 247: Is the comment above still relevant? It seems like we're doing some verification by proving we can decrypt the message, or is the verification you're talking about something else?

adrastaea: Also, is there maybe some utility in adding a parameter to the getEntries method to skip decryption and get the raw encrypted entries? Not sure if we'll need that, but could be a nice to have.

adrastaea (approved):
Looks good overall. Approving, but there's some comments that would be nice to have clarified before you merge.

packages/backend/src/nest/storage/channels/channel.store.ts
islathehut: This is handled already by the state manager.  When we get a new message (and possibly elsewhere) it sends all message IDs up to the state manager and if anything is missing from redux it fetches it.

packages/backend/src/nest/storage/channels/channel.store.ts
islathehut: Yea that's not relevant anymore.

Not currently but it would be possible to add.

PR #2747, opened 2025-02-13
fix(2708): Better error handling in orbitdb to mitigate cascading crashes

### LINKED ISSUE #2708: App crashes from libp2p errors
This mainly happens when creating channels and when joining large communities but can happen in other scenarios.  The main errors seen stream reset errors on the muxer and want aborted/promise rejection errors on orbitdb/helia operations.

The full fix will require updates to libp2p/yamux/orbitdb but we can workaround it in our code to avoid the crashes.

islathehut:
### Pull Request Checklist


packages/backend/src/nest/libp2p/libp2p.datastore.ts
```diff
@@ -63,24 +63,28 @@ export class Libp2pDatastore {
...
+    return this._deleteKeysByPrefixLevel(prefix === Libp2pDatastorePrefix.ALL ? undefined : prefix)
   }
 
   public async close() {
     if (!this.inMemory) {
-      await (this.datastore as LevelDatastore).db.close()
-      await (this.datastore as LevelDatastore).close()
+      try {
+        await (this.datastore as LevelDatastore).close()
+        await (this.datastore as LevelDatastore).db.close()
```
adrastaea, line 75: Why would we call `LevelDatastore.db.close()` when all the method `LevelDatastore.close()` does is call `this.db.close()`?

packages/backend/src/nest/libp2p/libp2p.datastore.ts
```diff
@@ -91,7 +95,7 @@ export class Libp2pDatastore {
     return deletedKeys
   }
 
-  private async _deleteKeysByPrefixLevel(prefix: string): Promise<string[]> {
+  private async _deleteKeysByPrefixLevel(prefix: string | undefined): Promise<string[]> {
```
adrastaea: Should we just make the parameter type `KeyQuery` since we're coercing it into that type anyway?

packages/backend/src/nest/storage/orbitDb/keyValueIndexedValidated.ts
```diff
@@ -40,7 +40,9 @@ import {
 import { type Helia } from 'helia'
 import { toString as uint8ArrayToString } from 'uint8arrays/to-string'
 import { createLogger } from '../../common/logger'
+import { abortableAsyncIterable } from '../../common/utils'
 import { KeyValueWithStorage } from './keyValueWithStorage'
+import { OrbitDbService } from './orbitDb.service'
```
adrastaea:
```suggestion

```

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -146,6 +146,7 @@ export class StorageService extends EventEmitter {
   }
 
   public async stop() {
+    await this.channelsService.closeFileManager()
```
adrastaea: Maybe we should have a unified `close()` method if both of these need to be called to close the channelsService

package.json
```diff
@@ -15,7 +15,9 @@
     "lint-staged": "lerna run lint-staged",
     "build:auth": "cd ./3rd-party/auth && pnpm install && pnpm build",
     "build:noise": "cd ./3rd-party/js-libp2p-noise && npm i && npm run build",
-    "bootstrap": "npm run build:auth && npm run build:noise && lerna bootstrap",
+    "build:orbitdb": "cd ./3rd-party/orbitdb && npm i && npm run build:dist && npm run build:debug",
```
adrastaea, line 18: I'm not familiar with orbitdbs builds, but it looks like we only have an npm script for building a debug build of orbitdb. Is that safe for production, or should we have a different process for the CD pipeline?

packages/backend/package.json
```diff
@@ -13,7 +13,9 @@
     "webpack": "webpack --env mode=development && cp ./lib/bundle.cjs ../backend-bundle/bundle.cjs",
     "webpack:prod": "webpack --env mode=production && cp ./lib/bundle.cjs ../backend-bundle/bundle.cjs",
     "postinstall": "npm run applyPatches",
-    "applyPatches": "patch -f -p0 < ./patch/electron-fetch.patch || true && patch -f -p0 --forward --binary < ./patch/parse-duration.patch || true && patch -f -p0 --forward --binary < ./patch/parse-duration-esm.patch || true && patch -f -p0 < ./patch/ipfs-pubsub-peer-monitor.patch || true && patch -f -p0 < ./patch/itws-sink.patch || true && patch -f -p0 < ./patch/mplex.patch || true && patch -f -p0 < ./patch/upgrader.patch || true",
+    "applyPatches": "npm run applyPatches:it && npm run applyPatches:libp2p",
```
adrastaea, line 16: Nice to get rid of some patches. What changed to justify that?

adrastaea (changes requested):

packages/backend/package.json
islathehut: Most of them were invalid, pointing at files that didn't exist.

package.json
islathehut: This mimics what they do in production.  The `build` command runs both but the reason I split it out like this is because it also runs `jsdoc` and for some reason that was breaking windows builds in CI.  If you look in node_modules for a non-submodule branch the same files are in the orbitdb `dist` directory so we're good.

packages/backend/src/nest/storage/storage.service.ts
islathehut: I'm not sure why I kept them split.  I might have thought they would need to be run at different times like we do with the libp2p datastore but I'm not sure.

packages/backend/src/nest/libp2p/libp2p.datastore.ts
islathehut: I don't remember the issue but I had seen something about making sure the level db was actually open and closed because it could sometimes get stuck so I figured we would be cautious and verbose.

adrastaea (approved):

PR #2768, opened 2025-03-07
fix(2770): Don't remove from connected peers in hangUp and smarter redial delays

### LINKED ISSUE #2770: Crashes when hanging up and redialing peers after auth disconnects
# Version

4.0.2

# Relevant Discussion

https://zbay.slack.com/archives/G01EHE6JHCM/p1741378217376519

# Description

When connecting to a peer that isn't in our sig chain we hang up and clear the user from the libp2p peer store and our own data structures that track who we have dialed/connected to.  We won't try to redial that peer again but the other peer will try to redial us after a delay.  Two issues present themselves:

1. We also clear peers from the data structures that track who we have dialed/connected to when we receive a disconnect event from libp2p.  This can cause a race condition where we hang up before processing that event and end up hitting a null exception.
2. The redial delay isn't exclusive to an individual peer and we can end up redialing peers much more rapidly than intended.  In my testing this can cause crashes by, seemingly, overwhelming libp2p with bad connections.

# Reproducing

1. Create a community
2. Join with a peer
3. Close out of the non-owner peer
4. Join with another peer when only the owner is online
5. Once the new peer has synced close the owner
6. Reopen the first non-owner peer
7. Wait a little bit for the dial/redial process to occur

In my experience the first issue (on hang up) is less likely to occur than the second (redial overload).  I can pretty reliably see the latter when letting the redial process to go on for a minute or so.

islathehut:
### Pull Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [4.0.3]
+
+### Fixes
+
+* Fixes issue where the app may crash when trying to redial a peer that doesn't recognize your user as being in the sigchain ([#2770](https://github.com/TryQuiet/quiet/issues/2770))
+* Fixes issue where the app may crash when hanging up on a peer we don't have in our sigchain ([#2770](https://github.com/TryQuiet/quiet/issues/2770))
+
 ## [4.0.0]
 
 ### New features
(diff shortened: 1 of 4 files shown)


islathehut:
#2770

PR #2772, opened 2025-03-12
chore(2771): Logging to files

### LINKED ISSUE #2771: Log to file
Right now logs are exclusively written to the console and it's a pain to get access them because you have to be consuming the logs proactively.  We should log to the filesystem for easy retrieval.

islathehut:
### Pull Request Checklist


packages/node-common/src/logger.ts
```diff
@@ -0,0 +1,79 @@
+// @ts-ignore
```
islathehut, line 1: This is kind of dumb but I had to do it because we can't use anything that uses the `fs` package in ios or android.  We have to split it up amongst the desktop/node environments and mobile.

packages/mobile/patch/rnfl/AndroidManifest.xml.patch
```diff
@@ -0,0 +1,17 @@
+--- node_modules/react-native-file-logger/android/src/main/AndroidManifest.xml	2025-03-11 23:02:55
```
islathehut, line 1: These two patches implement functionality that it's in review on `react-native-file-logger` but haven't been released yet.

packages/mobile/android/app/src/main/cpp/own-native-lib.cpp
```diff
@@ -141,17 +141,33 @@ int start_redirecting_stdout_stderr()
...
+    jclass cls_StringBuilder = env->FindClass("java/lang/StringBuilder");
+    jmethodID ctr_StringBuilder = env->GetMethodID(cls_StringBuilder, "<init>", "(I)V");
+    jobject stringBuilder = env->NewObject(cls_StringBuilder, ctr_StringBuilder, 100);
+    jmethodID mid_StringBuilder_append = env->GetMethodID(cls_StringBuilder, "append", "(Ljava/lang/String;)Ljava/lang/StringBuilder;");
+    for (auto str: { dataPath, logs_js }) {
+        env->CallObjectMethod(stringBuilder, mid_StringBuilder_append, str);
+    }
+    jmethodID mid_StringBuilder_toString = env->GetMethodID(cls_StringBuilder, "toString", "()Ljava/lang/String;");
+    auto js_log_path = (jstring) env->CallObjectMethod(stringBuilder, mid_StringBuilder_toString);
+    const char* log_path = env->GetStringUTFChars(js_log_path, nullptr);
```
islathehut, line 164: I mentioned before I'm not great with C++, this is what came up when I searched how to concatenate two `jstring` variables.  I have to imagine there's a simpler way but this works.

packages/logger/src/index.ts
```diff
@@ -245,12 +248,10 @@ export class QuietLogger {
    * @param formattedLogStrings Array of formatted log strings
    */
   private printLog(level: LogLevel, ...formattedLogStrings: string[]): void {
-    // we have to do this conversion because console doesn't have a trace method
-    const printLevel: LogLevel = level === LogLevel.TRACE ? LogLevel.LOG : level
-
-    // @ts-ignore
-    nodeConsoleLogger[printLevel](...formattedLogStrings)
+    this.internalLogMethod(level, formattedLogStrings.join(' '))
```
islathehut, line 251: This allows us to easily swap in different log mechanisms depending on the envrionment.

packages/identity/src/logger.ts
```diff
@@ -1,3 +1,10 @@
-import { createQuietLogger } from '@quiet/logger'
+import { createQuietLogger, DEFAULT_INTERNAL_LOG_METHOD, __nodeConsoleLogger } from '@quiet/logger'
 
-export const createLogger = createQuietLogger('identity')
+export const createLogger = (() => {
```
islathehut, line 3: Everywhere that I'm doing this funky try/catch is meant to avoid the `fs` import errors on mobile.  The only place we load this package is in desktop environments and the backend so any of these common packages won't break in a react native environment.

packages/desktop/src/main/main.ts
```diff
@@ -32,16 +32,11 @@ const webcrypto = new Crypto()
 
 global.crypto = webcrypto
 
-let dataDir = DESKTOP_DATA_DIR
 let mainWindow: BrowserWindow | null
 let splash: BrowserWindow | null
 let invitationUrl: string | null
 
-if (isDev || process.env.DATA_DIR) {
```
islathehut, line 40: This logic just got moved to `loadMainEnvs` so we could ensure the log env variables are set as early as possible.

PR #2774, opened 2025-03-17
release: 4.0.3

islathehut:
### Pull Request Checklist


packages/node-common/src/logger.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Generate a function that creates a module-level logger with a name like `packageName:moduleName`.  This is the main
+ * entry point for logging in Quiet.
+ *
+ * @param packageName Name of the package we are logging in
+ * @param parallelConsoleLog If true we will also log to the native console (e.g. browser console)
+ * @returns A function that can be used to generate a module-level logger
+ */
+export const createWinstonQuietLogger = (
```
adrastaea, line 62: If I'm understanding right, we created a new package in the monorepo called `node-common` which just includes this logger module? Why didn't we just make this part of the existing `@quiet/logger` package?

packages/node-common/src/logger.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  return winston.createLogger({
+    level: 'silly', // this is just because we are doing the log level checking via debug
+    transports: winstonTransports,
+  })
+}
+
+/**
+ * Generate a function that creates a module-level logger with a name like `packageName:moduleName`.  This is the main
+ * entry point for logging in Quiet.
```
adrastaea: Now, both `createQuietLogger` and this function  contain the note `This is the main entry point for logging in Quiet.`

packages/node-common/package.json
```diff
@@ -0,0 +1,75 @@
+{
+  "name": "@quiet/node-common",
+  "version": "2.0.2-alpha.1",
```
adrastaea: This doesn't seem correct..

packages/integration-tests/src/logger.ts
```diff
@@ -1,3 +1,3 @@
-import { createQuietLogger } from '@quiet/logger'
+import { createWinstonQuietLogger } from '@quiet/node-common'
 
-export const createLogger = createQuietLogger('state-manager:test')
+export const createLogger = createWinstonQuietLogger('state-manager:test')
```
adrastaea, line 3: I feel like this is reversed from what it should be. `createQuietLogger` should be the entry, and if they have the flag set to log to a file, then createQuietLogger should instantiate Winston under the hood to provide that functionality.

adrastaea (changes requested):

packages/node-common/src/logger.ts
islathehut: That's how I had it originally but unfortunately that breaks mobile.  Importing winston alone breaks mobile because it imports `fs` and importing a logger that has the file functionality, even if it isn't used in mobile, breaks mobile.  Splitting it out is the only option.

I want this to be the dumping ground for common functions that break mobile.  One such case is the dummy file generator function that we have in `backend` and `e2e`; we tried putting that in `common` but it broke mobile for the same reason.

packages/node-common/package.json
islathehut: Ah yea, I started with a copy of the common package when I created this.  None of our package versions make sense besides desktop and mobile.

packages/node-common/src/logger.ts
islathehut: I can update the comments.

adrastaea (approved):

Issue #188, opened 2022-01-28
User should be able to see that grouped messages are different messages (increase padding)


holmesworcester:
Problem: right now, grouped messages are hard to distinguish from single messages.

Slack inserts larger line breaks or spacing between grouped messages than the linebreaks in individual messages. 

We should too.

holmesworcester:
Related to #1403

PR #2779, opened 2025-03-26
Feat 118 message spacing

jreamy:
This addresses https://github.com/TryQuiet/quiet/issues/188 to insert larger line breaks between grouped messages.

### Pull Request Checklist


jreamy:
Two few questions @holmesworcester:
 1. Do you think this is a good approach? I thought about just inserting a <br/> or other divider between each `<NestedMessageContent>`... thoughts on which approach you would prefer?
 2. How much spacing do you want here? I started with 7 pixels, but it could be more... in [#1403](https://github.com/TryQuiet/quiet/issues/1403) you said 1.5 lines which would be ~10 px margin

7px margin:
Screenshot 2025-03-26 at 12 48 32 PM

10px margin:
Screenshot 2025-03-26 at 1 09 50 PM

holmesworcester:
That approach with 10px looks good to me!

PR #2780, opened 2025-03-30
Fix/2766-wrong-timezone-on-mobile

### LINKED ISSUE #2766: Timestamps displaying in wrong timezone on Android
A user reports that—while timestamps in other apps reflect the device timezone—timestamps on messages in Quiet are in GMT or some other timezone. 

OS:
LineageOS 18.1 Feb 27, 2025 version

Installed from APK.

holmesworcester:
Fixes #2766 - the Android app was not correctly displaying the local timezone. 

### Pull Request Checklist

- [X] I have run e2e tests for mobile


holmesworcester:
I moved the mobile date formatting code into the front-end components--since both the intended behavior and the timezone detection mechanism are different on mobile--to avoid future confusion.

Does this seem reasonable? 

I currently don't have a way to update snapshots (I think this requires a Mac for running iOS) though this PR should not have changed them. Thoughts?

holmesworcester:
@adrastaea this is ready for review.

I'm fairly sure the detox tests are failing in CI because there is currently no runner set up (our mac mini). I think we shouldn't worry about it as part of this issue, but we should set that up. 

They're visual regression tests I think, so updating them locally might not be what we want, or I might need a Mac to do it.

packages/mobile/src/components/ChannelList/ChannelList.component.tsx
```diff
@@ -1,20 +1,40 @@
...
   let communityName = '...'
   if (community?.name) {
     communityName = capitalizeFirstLetter(community.name)
   }
   return (
     <View style={{ flex: 1 }} testID={'channel-list-component'}>
       <Appbar title={capitalizeFirstLetter(communityName)} position={'flex-start'} contextMenu={communityContextMenu} />
+
+      {/* Debug button - only for testing */}
+      {onDebugPress && (
```
adrastaea: I'm struggling to find where you have disabled the debug button in production. Seems here like you're choosing not to render the TouchableOpacity button when `onDebugPress` is not defined, but I can't find where you have selectively not defined `onDebugPress` when in prod

packages/mobile/src/screens/ChannelList/ChannelList.screen.tsx
```diff
@@ -78,5 +99,20 @@ export const ChannelListScreen: FC = () => {
...
+      })
+    )
+  }, [dispatch])
+
+  return (
+    <ChannelListComponent
+      community={community}
+      tiles={tiles}
+      communityContextMenu={communityContextMenu}
+      onDebugPress={handleDebugPress}
```
adrastaea: If I'm understanding the mechanism right, we need an environment switch here that doesn't include onDebugPress here to disable the debug button

packages/state-manager/src/utils/functions/dates/formatMessageDisplayDate.ts
```diff
@@ -1,11 +1,15 @@
...
+  const formattedOffset = `UTC${tzOffsetHours >= 0 ? '+' : ''}${tzOffsetHours}`
+
   const LC = process.env.LC_ALL || 'en_US.UTF-8'
   const locale = LC.split('_')[0]
-  const messageDate = DateTime.fromSeconds(createdAt).setLocale(locale)
-  const now = DateTime.now().setLocale(locale)
-
+  const messageDate = DateTime.fromSeconds(createdAt).setZone(formattedOffset).setLocale(locale)
+  const now = DateTime.now().setZone(formattedOffset).setLocale(locale)
+  const check = messageDate.hasSame(now, 'year') && messageDate.hasSame(now, 'day')
```
adrastaea: Value declared but never used

packages/mobile/src/components/LocaleDebug/LocaleDebug.component.tsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
+import { Text, ScrollView, StyleSheet } from 'react-native'
+import { DateTime } from 'luxon'
+
+export const LocaleDebug = () => {
```
adrastaea:
```suggestion
export const LocaleDebug: React.FC = () => {
```

Fixes an eslint warning for missing explicit return type

packages/mobile/src/screens/ChannelList/ChannelList.screen.tsx
```diff
@@ -54,6 +52,29 @@ export const ChannelListScreen: FC = () => {
     [dispatch]
   )
 
+  const formatTileDate = (createdAt: number): string => {
```
adrastaea, line 55: I'm not a fan of putting this here instead of in the mobile utils folder that we have been using for reusable functions. You mentioned that you did this to prevent confusion, but I think with this we risk having multiple date formatting functions being used across the mobile stack if we don't have a unified reusable function to use on mobile. What was wrong with it being in the utils folder?

adrastaea (changes requested):

packages/mobile/src/components/ChannelList/ChannelList.component.tsx
holmesworcester: This was cruft from me adding an elaborate debug screen to see why I wasn't getting the right timezone. Will fix!

packages/mobile/src/components/LocaleDebug/LocaleDebug.component.tsx
holmesworcester: This shouldn't even be here, my mistake. Will remove.

packages/mobile/src/screens/ChannelList/ChannelList.screen.tsx
holmesworcester: I can move it back to a utils function and I'm not attached to this. Here was my reasoning:

1. Luxon docs strongly suggested to keep the date formatting as close as possible to the rendering, and otherwise pass around unix epoch time, to avoid confusion. 

2. Formatted times are less reusable than they seem because they're very tailored to the context in which they're being used. The way we display date in the floating date marker might be different  from how we display it in a tooltip, or on a user profie, or on the channel list. 

3. If we try to centralize date formatting in one place, it becomes a tangle when you need to change it in a specific place or debug something.

4. The main thing I wanted to do was remove formatMessageDisplayDate, since there were nearly-identical versions of it being used in desktop and mobile, which tripped me up and seemed like it could easily trip up someone else.

Thoughts?

packages/mobile/src/screens/ChannelList/ChannelList.screen.tsx
holmesworcester: this was cruft. removed.

packages/state-manager/src/utils/functions/dates/formatMessageDisplayDate.ts
holmesworcester: removed

Issue #2570, opened 2024-06-14
Back button navigation issue in user profile screens


holmesworcester:
Right now there are two related issues with the user profile screens. 

Steps to reproduce:
1. Click user profile button in bottom left
2. Tap "edit profile"

Expected: there should be a back button in the top left of the "edit profile" panel to take you back to the "profile" panel.
Actual: there is a close button, not a back button.

Expected: if I click away from "edit profile" and click on the username tab in the bottom left again, I should see the user profile screens. Instead, if "edit profile" was open, I arrive directly at the "edit profile" panel and must close it to get to the "profile" panel. 

Screencast of issue:

https://github.com/TryQuiet/quiet/assets/213678/334c68fa-f5d7-4160-bf20-8f7144e85829

Designs for reference:

https://www.loom.com/share/94718e9418334addb45aaa9e507a5570
https://www.figma.com/design/ceybIXGmg8akl9qgBoUrwe/Desktop-settings?node-id=13-37&t=u0q6sQvILPkWLyoc-4

IvanTsxx:
Is anyone working on this?

DubemObi:
Hello, I'll like to work on this, please can you assign this task to me

holmesworcester:
@ivan2214 @DubemObi either of you is welcome to work on it! 

@ivan2214 since you asked first, would you like to pick it up? 

@DubemObi would you like to pick another "good first issue" from the list?

kingalg:
Version: 4.1.0-alpha.0

Fixed.

PR #2784, opened 2025-04-08
Fix(2570): Close and back functions in the User Profile

kingalg:
Pull Request Checklist

Fixes issue: https://github.com/TryQuiet/quiet/issues/2570

Back button navigation in user profile:

- Close button changed to back button in Edit Profile view,

- When the user closes the side menu by clicking in the area outside of the menu and opens the menu again, the view will be User Profile.

- [x]  I have linked this PR to a related GitHub issue.

- [x]  I have added a description of the change (and Github issue number, if any)

### DIFF
--- a/packages/desktop/src/renderer/components/ContextMenu/ContextMenu.component.tsx
+++ b/packages/desktop/src/renderer/components/ContextMenu/ContextMenu.component.tsx
@@ -1,6 +1,7 @@
 import React, { FC, useEffect, useRef } from 'react'
 import { Grid, List, Typography, useTheme } from '@mui/material'
 import CloseIcon from '@mui/icons-material/Close'
+import ArrowBackIcon from '@mui/icons-material/ArrowBack'
 import {
   ContextMenuProps,
   ContextMenuHintProps,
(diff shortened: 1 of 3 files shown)


packages/desktop/src/renderer/components/ContextMenu/ContextMenu.component.tsx
holmesworcester, line 1: I know this was already like this and feel free to ignore, but it's bugging me that this is called "ContextMenu", since a context menu has a specific meaning. SidebarMenu? SettingsMenu?

holmesworcester:
I'm seeing a visual glitch when I enter the "edit profile" screen where the whole screen flashes white for a moment. This was a pre-existing issue so we can make another issue for it if it's annoying to fix now. 

[Screen recording 2025-04-10 2.20.20 AM.webm](https://github.com/user-attachments/assets/e2ad9149-1ed4-4bc6-9cc8-5ea3d915b3a9)

holmesworcester (approved):
Looks great! 

I left a separate comment about a "white flash" display glitch on opening the edit profile menu, but this does not seem to be related to current work (i can reproduce it on develop.)

packages/desktop/src/renderer/components/ContextMenu/ContextMenu.component.tsx
kingalg: Thanks for review!

1. I know about the glitch. I talked about this with Bartek, and it looks like it's because of how this side menu is made/structured. In general, it should be refactored. Most issues in this area are caused by how it was made, but that's much bigger work requiring writing it almost from scratch, and what we have now is mostly working as intended after fixes, so it's not a top priority. 

2. About the naming, we already use some of those names (or their combinations) in different places. I agree that it is confusing and makes things harder to find. But if we were to change it here, we should carefully look at naming in all places to avoid duplication/inconsistency. 

3. FYI, we have the same close/back button icon issue in Community Settings. Now that I know how to fix it, it should be a quick fix if you want me to do it.

PR #2788, opened 2025-04-08
Chore/remove deprecated id systems

### LINKED ISSUE #2762: Identity Should Be Based on Sigchain
## Acceptance Criteria
- Removed Cert/CSR systems
- Identities based on LFA chain

adrastaea:
### Pull Request Checklist


packages/backend/src/nest/libp2p/libp2p.auth.ts
```diff
@@ -307,48 +310,34 @@ export class Libp2pAuth {
...
+            device: (sigChain.context as Auth.InviteeContext).device,
+            team,
+            user,
+          } as Auth.MemberContext
+        }
       }
       this.joinStatus = JoinStatus.JOINED
       this.unblockConnections(this.bufferedConnections)
       this.emit(Libp2pEvents.AUTH_JOINED)
-      await this.sigChainService.saveChain(team.teamName)
```
islathehut, line 338: So it seems like the only places we save the chain to disk now are on joining/changing teams, creating invites, and stopping the app cleanly, right?  I think we still need to at least periodically save the chain to disk because if the app crashes or a user forces the app to close and we don't go through the proper stop flow we won't save any intermediate changes, correct?

packages/backend/src/nest/libp2p/libp2p.service.ts
```diff
@@ -502,16 +551,23 @@ export class Libp2pService extends EventEmitter {
...
+    })
 
-    this.logger.info(
-      `Libp2p Multiaddrs:`,
-      this.libp2pInstance.getMultiaddrs().map(addr => addr.toString())
-    )
+    await this.libp2pInstance.start()
+    await this.addPeersToDialQueue()
+    this.logger.info('Queued peers for initial dialing')
+    this.logger.info(`Dialing peers and starting libp2p`)
```
islathehut: Could you move this above the instance start?

packages/backend/src/nest/local-db/local-db.service.ts
```diff
@@ -98,21 +95,90 @@ export class LocalDbService {
...
-      return filterAndSortPeers(peers, stats, localPeerAddress, includeLocalPeerAddress)
+  /**
+   * Update the local db entry for the given peers with the given stats
+   * @param stats
+   */
+  public async updatePeerStats(stats: Record<string, NetworkStats>) {
+    this.logger.debug('Updating peer stats', stats)
+    for (const addr of Object.keys(stats)) {
+      if (!isMultiaddr(multiaddr(addr))) {
+        this.logger.error('Invalid multiaddr', addr)
```
islathehut, line 121: If I'm reading this right if an address is invalid we log and still write hte stats for that invalid address to the DB?  Should we skip any invalid addresses?

packages/backend/src/nest/storage/channels/messages/messages.service.ts
```diff
@@ -90,4 +88,15 @@ export class MessagesService extends EventEmitter {
...
+
+  public validateMessage(message: ChannelMessage, encryptedMessage: EncryptedMessage): boolean {
+    if (message.id !== encryptedMessage.id) {
+      this.logger.info(`Message ID mismatch`, message.id, encryptedMessage.id)
+      return false
+    }
+    if (message.type === MessageType.Info) {
+      return true
+    }
+    return true
```
islathehut, line 100: Why are we checking this if we return true either way?

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -95,9 +83,9 @@ export class StorageService extends EventEmitter {
       return
     }
 
-    await this.communityMetadataStore.startSync()
-    await this.certificatesStore.startSync()
-    await this.certificatesRequestsStore.startSync()
+    // await this.communityMetadataStore.startSync()
+    // await this.certificatesStore.startSync()
+    // await this.certificatesRequestsStore.startSync()
```
islathehut: Delete

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -130,9 +118,9 @@ export class StorageService extends EventEmitter {
     // picked up by the CertificatesStore. Perhaps we can initialize stores
     // first and then load data/send events.
     this.logger.info('2/3')
-    await this.certificatesStore.init()
-    await this.certificatesRequestsStore.init()
-    await this.communityMetadataStore.init()
+    // await this.certificatesStore.init()
+    // await this.certificatesRequestsStore.init()
+    // await this.communityMetadataStore.init()
```
islathehut: Delete this plus the comment above about the certificates store.

packages/backend/src/nest/storage/storage.service.ts
```diff
@@ -181,163 +151,70 @@ export class StorageService extends EventEmitter {
...
     this.peerId = null
 
     await this.channelsService.clean()
 
-    this.certificatesRequestsStore.clean()
-    this.certificatesStore.clean()
-    this.communityMetadataStore.clean()
+    // this.certificatesRequestsStore.clean()
+    // this.certificatesStore.clean()
+    // this.communityMetadataStore.clean()
```
islathehut, line 206: Delete

packages/backend/src/nest/storage/storage.types.ts
```diff
@@ -12,14 +10,12 @@ export enum StorageEvents {
   UPLOAD_PROGRESS = 'uploadProgress',
   DOWNLOAD_PROGRESS = 'downloadProgress',
   MESSAGE_MEDIA_UPDATED = 'messageMediaUpdated',
-  CHECK_FOR_MISSING_FILES = 'checkForMissingFiles',
   // Misc
   SEND_PUSH_NOTIFICATION = 'sendPushNotification',
   // Users
   CSRS_STORED = 'csrsStored',
   CERTIFICATES_STORED = 'certificatesStored',
```
islathehut: Delete these plus any other related events.

packages/common/src/invitationLink/invitationLink.validator.ts
```diff
@@ -29,7 +29,8 @@ import base64url from 'base64url'
 const logger = createLogger('invite:validator')
 
 const ONION_ADDRESS_REGEX = /^[a-z0-9]{56}$/g
-const PEER_ID_REGEX = /^[a-zA-Z0-9]{52}$/g
+// Accept both 46‑char (CIDv0, "Qm…") and 52‑char (ed25519, "12D3…") base58 peer IDs
+const PEER_ID_REGEX = /^(?:[A-Za-z0-9]{46}|[A-Za-z0-9]{52})$/
```
islathehut, line 33: When are we using 46 character IDs?

packages/desktop/src/renderer/components/widgets/possibleImpersonationAttackModal/PossibleImpersonationAttackModal.container.tsx
```diff
@@ -10,7 +10,8 @@ const PossibleImpersonationAttackModalContainer = () => {
   const possibleImpersonationAttackModal = useModal(ModalName.possibleImpersonationAttackModal)
 
   const community = useSelector(communities.selectors.currentCommunity)
-  const duplicateCerts = useSelector(users.selectors.duplicateCerts)
+  // const duplicateUsers = useSelector(users.selectors.duplicateUsers)
+  const duplicateUsers = false
```
islathehut, line 14: Was this intentional?

packages/desktop/src/renderer/components/widgets/usernameTakenModal/UsernameTakenModal.container.tsx
```diff
@@ -32,20 +32,24 @@ const UsernameTakenModalContainer = () => {
...
-      if (registeredUsernames.has(nickname)) {
-        setRegistrationError(`${nickname} is already taken`)
-      }
+      // if (
+      //   Object.values(registeredUsernames)
+      //     .map(u => u.username)
+      //     .includes(nickname)
+      // ) {
+      //   setRegistrationError(`${nickname} is already taken`)
+      // }
```
islathehut, line 47: Same here, was this intentional?

packages/desktop/src/renderer/sagas/index.saga.ts
```diff
@@ -5,7 +5,7 @@ import { startConnectionSaga } from './socket/socket.saga'
 import { socketActions } from './socket/socket.slice'
 
 export default function* root(): Generator {
-  const dataPort = new URLSearchParams(window.location.search).get('dataPort') || ''
+  const dataPort = new URLSearchParams(window.location.search).get('dataPort') || '4677'
```
islathehut, line 8: Why 4677?

islathehut (changes requested):
First round

packages/backend/src/nest/libp2p/libp2p.auth.ts
adrastaea: No, if you look in `sigchain.service.ts` you will see that now we save the chain when the `updated` event is emitted which happens when a block is added to the chain. We were doing the same using the re-emitted event on the `Connection` object, but I noticed a ton of redundant saves being scattered around the application so I centralized it to the sigchain service itself. We shouldn't need to put `saveChain` anywhere else now.

packages/backend/src/nest/libp2p/libp2p.service.ts
adrastaea: Just the logs? Sure.

packages/backend/src/nest/local-db/local-db.service.ts
adrastaea: Right, I'll add a statement to delete the data at that key.

packages/backend/src/nest/storage/channels/messages/messages.service.ts
adrastaea: I considered moving all of the message form verification from the saga, but ended up just leaving it as a stub. I'll either delete it or add a TODO to remind us that we may want to move all the validation from the saga.

packages/common/src/invitationLink/invitationLink.validator.ts
adrastaea: I'm not sure, but it was breaking some tests

packages/desktop/src/renderer/components/widgets/possibleImpersonationAttackModal/PossibleImpersonationAttackModal.container.tsx
adrastaea: Yes, since we're not blocking duplicate usernames now, I wanted to just cleanly disable without completely ripping out all of this code in case we wanted to reuse it for later attack warnings

packages/desktop/src/renderer/sagas/index.saga.ts
adrastaea: It's a default we've used elsewhere in tests. This change just supresses warnings in renderer tests because we don't have a real electron window to get a dataport from.

islathehut (approved):

Issue #1018, opened 2022-11-10
Compress images to save space


holmesworcester:
A user was upset that we were downloading lots of data on a metered connection. Even if users have unlimited data plans or use wifi, local data storage will be an important limit for us. 

In my test of Signal it compressed a photo 13:1 (down to 119kb from 1.6MB) and quality was acceptable. So we should compress images at least 10:1 before sending. 

Ideally the user would not perceive this delay when sending an image. We should ensure: 

1. if this process is interrupted in any way it is restarted so the image can be sent 
2. we should ensure that if the user switches away from the app after sending an image it will still be sent.
3. this should live in the cross platform code (backend or state-manager) and not in frontend code

holmesworcester:
Note that we should merge https://github.com/TryQuiet/quiet/pull/2868 first, which will make this PR more readable.

kingalg:
Version: 5.1.0-alpha.0
System: all

My 1.6mb compressed to 179kb, not 119kb as suggested in the issue description, but I would still call it a successful implementation. 
<img width="574" alt="Image" src="https://github.com/user-attachments/assets/5d5e69c3-2461-4260-9104-58f52e0e8f3e" />

holmesworcester:
It's an inexact process at least with this library so that's fine.

PR #2789, opened 2025-04-15
Feat/1018 compress images

holmesworcester:
This PR compresses JPEG images on desktop and mobile, removes EXIF when compressing them [1], and lifts the size limit to 15MB for user profile photo JPEG images, so that users can comfortably upload most photos without any issue.

Addresses: #1018 

[1] though not for all images, because re-compressing could cause size increases or quality damage to small images, and this library doesn't support EXIF removal without re-compressing)

### Pull Request Checklist


islathehut:
The number of files changed seems high and it looks like there are changes from 5.0.0 being shown as changes here.  I think something might be off with the merge from develop.

holmesworcester:
> The number of files changed seems high and it looks like there are changes from 5.0.0 being shown as changes here.  I think something might be off with the merge from develop.

The high number of changed files is mostly because I did a big renaming in the course of these changes.

I could undo those commits or cherry pick them out into a separate PR and that actually might make things easier for me.

I did the renaming because all file attachments were named "uploaded" even if they were recently downloaded, which seemed confusing and crufty as we go into QSS work where we actually will be uploading things. So I changed the names to "attached" or "attachment" since that matches reality better. Should I just give up on this for now? 

It's possible something else got messed up here too.

holmesworcester:
Closing this in favor of a newer PR rebased on the renaming PR.

Issue #2833, opened 2025-05-06
Specifies timezone for mobile snapshot tests


holmesworcester:
We weren't specifying TZ=UTC on our mobile tests, so snapshots would fail depending on the timezone they were run in.

PR #2834, opened 2025-05-06
Fix/2833-specifies-timezone-in-mobile-snapshot-tests

holmesworcester:
https://github.com/TryQuiet/quiet/issues/2833

### Pull Request Checklist


packages/mobile/src/components/Chat/__tests__/__snapshots__/Chat.test.tsx.snap
```diff
@@ -651,7 +651,7 @@ exports[`Chat component renders component 1`] = `
                           }
                           verticalTextAlign="center"
                         >
-                          12:46
+                          Jan 1, 1970, 12:00 AM
```
adrastaea: I don't think that we should be displaying a null time instead of the correct timestamp from the data source. I think there may be a deeper problem here.

adrastaea (changes requested):

packages/mobile/src/components/Chat/__tests__/__snapshots__/Chat.test.tsx.snap
holmesworcester: You're right! Sorry for missing this. I'll look into it.

holmesworcester: I'm confused as to how these passed at all since the last release because mocks didn't have actual timestamps and we weren't specifying a timezone. 

But now mocks have actual timestamps. And the timezone env variable was already working correctly.

adrastaea (approved):

PR #2863, opened 2025-05-19
fix: Fixes for crashes and connection hanging on 5.0.0

islathehut:
### Pull Request Checklist


packages/backend/src/nest/common/timed-queue.ts
```diff
@@ -56,7 +56,8 @@ export class TimedQueue {
       throw new Error(`Backoff factor must be a positive number greater than or equal to 1`)
     }
 
-    this.queue = fastq.promise(this._processQueue.bind(this), options.concurrency ?? DEFAULT_CONCURRENCY)
+    this._processQueue = this._processQueue.bind(this)
```
islathehut, line 59: All of the places where I updated the bind to be on a separate line are because I found while working on the QSS branch that doing it inline with the assignment as a function parameter didn't work properly but doing it like this did.

packages/backend/src/nest/libp2p/libp2p.service.ts
```diff
@@ -118,10 +119,16 @@ export class Libp2pService extends EventEmitter {
             innerEvent.payload.message === UNKNOWN_THIS_PEER) ||
           (innerEvent.type === 'LOCAL_ERROR' && innerEvent.payload.type === 'TIMEOUT')
 
-        const remotePeerId = args[0].connection?.remotePeerId?.toString()
+        const remotePeerId = args[0].connection?.remotePeerId?.toString() ?? args[0].connection?.remotePeer?.toString()
```
islathehut, line 122: This was weird, I don't know if something changed but the field `remotePeerId` didn't appear on the connection object and was instead present as `remotePeer`.  I kept both in just in case there's something transient about that behavior.

packages/logger/src/index.ts
```diff
@@ -473,7 +473,7 @@ export class QuietLogger {
   private _getLogSetting(): LogSetting {
     if (this._canTrace()) {
       return LogSetting.TRACE
-    } else if (debug.enabled(this.name)) {
+    } else if (debug.enabled('*') || debug.enabled(this.name)) {
```
islathehut, line 476: These changes allow setting `DEBUG=*` or `DEBUG=*:trace` to quickly set log levels globally.  I got tired of rewriting the debug string with trace everytime I wanted to test.

packages/node-common/src/logger.ts
```diff
@@ -19,13 +19,17 @@ let _winstonLogger: Logger | undefined = undefined
 const initWinstonLogger = (): Logger | undefined => {
   if (_winstonLogger) return _winstonLogger
   // Add a non-colorized format for file logs
-  const fileFormat = format.combine(
+  const formats: winston.Logform.Format[] = [
```
islathehut, line 22: `COLORIZE` is now defaulting to false everywhere so the behavior should be the same under normal circumstances but can now be set for folks like me that like the ansi coloring for quickly finding log lines when skimming files.

packages/backend/src/nest/libp2p/libp2p.auth.ts
```diff
@@ -94,11 +94,15 @@ export class Libp2pAuth {
   private async unblockConnections(conns: { peerId: PeerId; connection: Connection }[]) {
     if (this.joinStatus === JoinStatus.NOT_STARTED && this.sigChainService.activeChainTeamName != null) {
       this.logger.info(`Unblocking ${conns.length} connections now that we have an active chain`)
-      this.joinStatus = JoinStatus.PENDING
+      this.joinStatus = this.sigChainService.getActiveChain()!.team == null ? JoinStatus.JOINED : JoinStatus.PENDING
```
adrastaea: I think this might be backwards? If the `team` is null, shouldn't we be pending a join not joined?

`condition ? valueIfTrue : valueIfFalse` -> `team is null ? JoinStatus.PENDING : JoinStatus.JOINED`

packages/backend/src/nest/libp2p/libp2p.service.ts
adrastaea: I think it's always been called `remotePeer` in the `Connection` object. There are other places where we interpret the string content of an event as `remotePeerId` which might be what you were thinking of.

adrastaea: Aaaand that's on me for mixing those up initially 🤦🏻‍♀️

packages/node-common/src/logger.ts
adrastaea: Can we remove this change to the default colorizing? I want it on by default for the console log because the colorization works there, but the ANSI codes in the log files don't render in a lot of text editors and may break some algorithmic parsing which is why I turned off the colorization in the log files. Can we separate out the console and file colorization into different environment variables?

adrastaea (changes requested):
Some changes needed I think. Also, when I ran the bootstrap, I ended up with a backend lock file different from what you committed here, so you may need to run bootstrap and then commit that again.

packages/backend/src/nest/libp2p/libp2p.auth.ts
islathehut: Oh shit, yea, that's my bad.

packages/node-common/src/logger.ts
islathehut: Done.

islathehut:
Just running bootstrap didn't update the package lock so I did a clean and bootstrap, hopefully that gets it in sync.

adrastaea (approved):

Issue #1740, opened 2023-08-30
Releases should always include the latest Tor binaries


holmesworcester:
Right now we have to manually update Tor binaries in a release.

Our build process should automatically pull or build the latest Tor binaries, so that we're never releasing with an out of date version of Tor.

Or we should have a manual process for this. 

Let's ask Tor folks about the best way to do this.

holmesworcester:
As a first step let's upgrade Tor on `develop` *after* each release so that checking the upgrade is folded into our dev and QA process.

Also we have a separate issue for Android

Issue #2875, opened 2025-05-28
Automate build of tor android binaries


islathehut:
# Overview

We now have a process for building tor binaries for android thanks to [this PR](https://github.com/TryQuiet/quiet/pull/2866) but for now this is a process we have to manually run.  For consistency and ease we should add a github action for performing this build.  

# Requirements

* Process laid out in the doc added as part of the above PR is automated
* Automation is easy for devs to run
* Easy to extract binary from the automation run
* Automation must run in Debian 12 (side note: test ubuntu)

# Notes

* Github actions is the most likely place for the automation to live
* If Debian is indeed a requirement as was noted in the PR then the binary build itself will probably need to be run in something like AWS codebuild since Debian isn't an available option for github runners
* If ubuntu is acceptable we can run the build process directly in github actions

bitmold:
HI, I've used a variant of this script in an Ubuntu LTS. Haven't tested recently, but it *should* run fine provided it's modern. Been talking to @holmesworcester about this and will take a stab at this later this week

PR #2866, opened 2025-05-21
Upgrdae Tor to 0.4.8.16 and tor's linked openssl to 3.5.0

bitmold:
@holmesworcester  Pertaining to the conversation on #2861 

libtor.so is build with the code in this repository https://github.com/bitmold/quiet_libtor_android

PR #2876, opened 2025-05-29
Update desktop (Mac, Windows, Linux) Tor binaries from 0.4.7.13 to 0.4.8.16 and improve upgrade process

holmesworcester:
Addresses https://github.com/TryQuiet/quiet/issues/1740

- Updates Tor binaries across all desktop platforms (Linux, macOS x64/ARM64, Windows) from version 0.4.7.13 to 0.4.8.16
- Adds new OpenSSL 3.0.16 libraries for Linux (libssl.so.3, libcrypto.so.3) 
- Documents the Tor upgrade process in desktop README
- Adds post-release checklist item to ensure Tor is updated after each release
- Adds automated Tor update script for future maintenance


bitmold:
```
➜  a scripts git:(feat/tor-upgrade-process → upstream {1} ✓) ./update-tor-binaries-desktop.sh  --check-only
[INFO] Checking current Tor version...
[INFO] Current Tor version: 0.4.8.16
[INFO] Checking for latest Tor Browser version...
[INFO] Latest Tor Browser version: 14.5.3
[INFO] Available Tor Browser version: 14.5.3
[INFO] Note: Tor binary version may differ from Tor Browser version
```

The last info logged is confusing because the `tor` binary version and the Tor Browser version will *always* differ, they're versioned completely separately, are on separate release schedules, and so for one thing their respective version numbers shouldn't shouldn't ever be compared + you can't infer say the version of tor from the version number of Tor Browser. Sometimes you'll see new builds of Tor Browser with changes (say for example to the firefox base) that doesn't actually even include a new release of tor. 

For that reason, I think the last log should be removed/edited so as to not confuse someone who may be running the script that isn't familiar with the exciting intricacies of tor + tor browser versions.

bitmold:
```
WORKFLOW:
    1. Run --check-only to see what would change
    2. Run --download to fetch new binaries
    3. Review changes, run tests manually if desired
    4. Run --install to replace current binaries
```

This is misleading. After running steps 1 and 2 the binaries are extracted to a temp directory eg `/tmp/tmp.32jKxCMJUH/extracted/linux`. At this point if someone were to move to step 3 there would be no changes to review, since none of the binaries have been added to Quiet so a user could mistakenly think the new binaries are ready to ship and proceed to `--install` when in reality they're just using the old binaries and nothing has changed.

bitmold:
As noted above, since  `--download` extracts to a temporary directory if a user were to follow the help instructions and run the following:

```
./update-tor-binaries-desktop.sh --download
./update-tor-binaries-desktop.sh --install
```

they get this error:
```
➜  a scripts git:(feat/tor-upgrade-process → upstream {1} U:1 ✗) ./update-tor-binaries-desktop.sh --install
[INFO] Installing extracted Tor binaries...
[ERROR] No extracted binaries found. Run --download first.
```

This is because each invocation of the script invokes 
`readonly TEMP_DIR="$(mktemp -d)"` so it's looking in a different temp directory between downloading and installing.

bitmold:
There's really no reason why you'd ever want to just download the binaries and *not* install them.

You can get rid of the issue above by just removing the `--download` and `--install` options altogether since downloading is useless without also installing them. The script only makes since to be run with `--full`.

Also, when you run `./update-tor-binaries-desktop.sh --full` you get this warning:
```
[WARNING] Running full automated update. Use with caution!
```

I think this is unnecessary since there's nothing really at stake. If the user were unsatisfied with the binaries (they break things or they simply want to go back to the old binaries they had before the script) they could just run `git reset --hard` to reset the repo to the original state.

bitmold:
@holmesworcester @islathehut this gave me an idea kind of regarding #2866 and #2875

Originally I had made #2866 because the version of guardian project's tor-android library was very out of date and upgrading to a version that was still like 2 years old broke quiet. So instead, I compiled `libtor.so` myself with certain compile flag optimizations for Quiet. 

It just occurred to me that since Tor Browser **also** targets Android this script could be extended to just grab their `libtor.so` from releases of Tor Browser on Android. 

Here's the page that the script queries for Tor Browser releases and currently uses for Linux, Mac and Windows: 
https://dist.torproject.org/torbrowser/14.5.3/

You'll see that it includes APKs for multiple architectures (`aarch64`, `arm7`, `x86` and `x86_64`) 

Currently quiet only targets `aarch64` aka `arm64-v8a`. But there you go, that's a perfectly good `libtor.so` for Quiet you just have to extract it which is trivial to do. A lot of people not intimate with Android don't know this, but .apk files are actually just .zip files with a different file extension - it's not some compiled binary blob like how a `.exe` is on windows.

Here's a crude example, for whatever reason in Tor Browser they name the file `libTor.so` instead of the more conventional `libtor.so`
```bash
curl -O https://dist.torproject.org/torbrowser/14.5.3/tor-browser-android-aarch64-14.5.3.apk
unzip tor-browser-android-aarch64-14.5.3.apk -d apk
cp apk/lib/arm64-v8a/libTor.so libtor.so
rm -rf apk
```

Viola, there's your latest Android APK with a `libtor.so` build by The Tor Project, and since libtor.so is built to include openssl, libevent, zlib and zstd you just need this one file into the Quiet Android package.


To use it, you'll still need to keep most of the work I had opened in: #2866, particularly: 
- changes I made to `BackendWorker.kt` to load libtor.so
- changes I made to `build.gradle` to remove the tor-android and jtorcontrol dependencies 
- all the header files I had git rm'd.
- libtor.so needs to be placed in the very specific folder I had used in that PR for the java/android runtime to allow it to be passed into a call to `System.loadLibrary("tor")` in Kotlin.

To me the easiest thing would be to merge that PR, delete the documentation I had included for building `libtor.so`, and then update this script to just overwrite the `libtor.so` that's found in my branch under `packages/mobile/android/app/src/main/jniLibs/arm64-v8a/`  with one retrieved by this script. 

@holmesworcester happy to implement these changes if interested

bitmold:
It's beyond the scope of this PR right now, but one thing that would absolutely be crucial to have for the day when Quiet is ready to be subjected to a serious security audit is to crytpographically verify the signatures of each binary. As of right now you're just hoping that the binaries served by this address haven't been tampered with. For each binary on that page, there's a `.asc` file which is a start.....

holmesworcester:
This is a great simplification, thanks for the suggestion! I think I should take it from here, since it's good for someone internal to the project to understand everything and run the scripts ourselves.

The remaining question is iOS. I'll look into that too but if you have suggestions it would be helpful!

Also thanks for your suggestions on the script. I'll simplify it so it just does what we want with no cruft.

I'll look into verifying the binaries too.

bitmold:
It occurred to me that when Tor Browser builds `tor` for Android they have to specify a minimum android API level with the Android NDK's `ndk-build` cmd that is needed to use libtor.so on Android. 

One reason why distributing binaries this way is very uncommon is probably because If they were to make a release of Tor Browser for Android with a minimum API level greater than Quiet's then suddenly all versions of Quiet running on an Android beneath that level would break. It seems like Tor Browser tries to keep their min API levels as low as possible and that Quiet actually has a higher min API level than Tor Browser does so this isn't likely to be an issue now but could possibly become one in the future, essentially forcing Quiet to keep in sync with whatever Tor Browser's min APi level is

holmesworcester:
Confirmed that this is working on Windows too.

3rd-party/tor/darwin/arm64/libevent-2.1.7.dylib
islathehut, line 1: No more arm mac binaries?

islathehut (approved):

PR #2877, opened 2025-06-02
feat(2759): Add QSS sigchain syncing to Quiet

### LINKED ISSUE #2759: QSS - Create a community with sigchain
Adding the functionality within QSS for adding a community with a sigchain to QSS and having QSS join as a server

### LINKED ISSUE #2760: Connect Quiet to QSS for community creation and joining
* Connect a quiet peer to QSS as an owner and run the create/join flow on QSS
* Connect a quiet peer to QSS as a non-owner and use it to join the sigchain
* Adding QSS-enabled invite links

islathehut:
- Adds QSS client logic
- Adds server service to sigchain
- Connect to QSS on joining/creating/signing in to community
- Updates LFA join logic to support partial join via QSS
- Adds new e2e test with QSS enabled
- Adds QSS docker initialization

### Pull Request Checklist


packages/backend/src/nest/auth/services/members/server.service.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    expect(servers.length).toBe(1)
+  })
+
+  it('should find the server we added and be valid', () => {
+    const thisServer = adminSigChain.server.getServer(server.host)
+    expect(thisServer).toBeDefined()
+    expect(thisServer?.host).toBe(server.host)
+    expect(thisServer?.keys).toEqual(server.keys)
+  })
+})
```
holmesworcester, line 34: Is this a good place to have a test to confirm that the server is not a member?

packages/backend/src/nest/auth/sigchain.service.ts
```diff
@@ -75,31 +77,44 @@ export class SigChainService extends EventEmitter {
...
+   * @param filter Filter query with either team name or team ID
    * @returns The chain for the team
    * @throws Error if the chain doesn't exist
    */
-  getChain(teamName: string): SigChain {
-    if (!this.chains.has(teamName)) {
+  getChain(filter: GetChainFilter): SigChain {
+    // reject filters with both team name and ID
+    if (filter.teamName != null && filter.teamId != null) {
+      throw new Error('Must provide one of `teamName` or `teamId` in filter query, not both!')
```
holmesworcester: why do we use both a teamName and teamId and not just teamId?

holmesworcester: Also, I'm curious how QSS supports multiple teams. Is there one instance of the sigchain and auth scheme that figures out what team the client is a member of, or will there be many, each one with its own endpoint?

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -394,11 +421,18 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
     }
   }
 
+  private async createCommunityOnQss(community: Community, sigchain: SigChain): Promise<void> {
```
holmesworcester, line 384: Should we force ourselves to decide whether we are using the word Team or Community and use it consistently?

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -420,11 +454,17 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
       peerList: [localAddress],
       psk: Libp2pService.generateLibp2pPSK().psk,
```
holmesworcester, line 415: does QSS need to know the PSK?  every time we submit to or fetch from QSS it will be gated by auth, right?

packages/backend/src/nest/libp2p/libp2p.auth.ts
```diff
@@ -306,18 +329,25 @@ export class Libp2pAuth {
...
         const user = this.sigChainService.user
         if (team) {
           authConnection.emit('sync', { team, user })
+          if (
+            authConnection._context.peer != null &&
+            !(authConnection._context.peer as Member).roles.includes(RoleName.MEMBER)
+          ) {
+            this.sigChainService.roles.addMember((authConnection._context.peer as Member).userId, RoleName.MEMBER)
+          }
+          await this.handleJoinViaQSS()
```
holmesworcester, line 348: so this is in the case where a peer encounters another peer on the p2p network who is a pending member who (half) joined via QSS?

holmesworcester (approved):

packages/backend/src/nest/auth/sigchain.service.ts
```diff
@@ -130,14 +145,21 @@ export class SigChainService extends EventEmitter {
...
+  addChain(chain: SigChain, setActive: boolean, teamName?: string, teamId?: string): boolean {
     teamName = teamName || chain.team!.teamName
     if (this.chains.has(teamName)) {
       throw new Error(`Chain for team ${teamName} already exists`)
     }
     this.chains.set(teamName, chain)
+    // if we have a team ID on either the team or passed into the method call we can add it to the
+    // id->name mapping
+    if (teamId != null || chain.team != null) {
+      this.teamIdsToNames.set((teamId ?? chain.team?.id)!, teamName)
```
adrastaea: This seems like it will create a possibility for disagreement between the teamName property of the `Team` object and this mapping. Why do we need this teamIdsToNames map?

packages/backend/src/nest/qss/qss.service.ts
```diff
@@ -0,0 +1,278 @@
...
+      host,
+      keys: generateKeysResponse.payload.payload.keys,
+    }
+
+    // add this QSS server/cluster to our chain using the keys we generated earlier
+    this.logger.info(`Got a valid keys response from QSS, adding it to the chain`, lfaServer)
+    sigChain.server.addServer(lfaServer)
+
+    const serializedSigChain: Uint8Array = sigChain.save()
+    const serializedKeyring: Uint8Array = uint8arrays.fromString(JSON.stringify(sigChain.team.teamKeyring()), 'utf8')
```
adrastaea, line 185: I believe the teamKeyring includes secrets. Are w sure this is what we want to be sending to QSS?

packages/backend/src/nest/auth/sigchain.service.ts
islathehut: Quiet tells QSS which team it's signing into.  QSS keeps records of all of the teams and maps a user's request to sync on a given chain to the correct chain on its end.

RE: team name and ID.  We should probably switch over entirely to mapping chains by ID and not name but that's a relatively big refactor.  For the purposes of this PR I needed to be able to filter on team ID since that's what QSS maps teams to (it doesn't care what the name is and obviously we can't guarantee any uniqueness between team names on QSS) and modifying the existing behavior to allow that without having to update the many instances we do the mapping by name rather than doing that refactor now seemed prudent.

packages/backend/src/nest/connections-manager/connections-manager.service.ts
islathehut: I don't think so (and in my current implementation it doesn't use it).  It is gated by chain authentication.  If you want to add/receive any data related to a community you have to authenticate with the chain.

packages/backend/src/nest/libp2p/libp2p.auth.ts
islathehut: Correct

packages/backend/src/nest/qss/qss.service.ts
islathehut: QSS needs the secrets to be able to join.  Servers don't join using an invite in LFA, an admin just adds the server metadata with the public keys that the server generates and then sends the complete chain and complete keyring.

packages/backend/src/nest/auth/sigchain.service.ts
islathehut: This is because unfortunately we don't have team ID on the team until we've synced with another user so when a new user joins we have to manually associate the ID with the name using the parameters passed in on the invite.

Maybe this could be handled better, though.

packages/backend/src/nest/connections-manager/connections-manager.service.ts
islathehut: In my mind Community refers to the overall set of people, data, functions, etc. So it encompasses the people, auth relationships, channels, messages, etc..  Team, on the other hand, would be the people and auth relationships in a community. Not sure if that matches y'alls thinking.

packages/backend/src/nest/auth/sigchain.service.ts
islathehut: I moved the logic into the QSS side to get rid of the mapping.

package.json
```diff
@@ -8,18 +8,23 @@
...
     "distAndRunE2ETests:mac:local": "lerna run --scope @quiet/desktop distMac:local && npm run e2e:mac:local",
+    "e2e:mac:local": "npm run start:qss && lerna run --scope e2e-tests test:localBinary --",
     "distAndRunE2ETests:linux:local": "npm run e2e:linux:build && npm run e2e:linux:run",
-    "e2e:mac:local": "lerna run --scope e2e-tests test:localBinary --",
-    "e2e:linux:build": "lerna run --scope @quiet/backend webpack:prod && lerna run --scope @quiet/desktop distUbuntu && lerna run --scope e2e-tests linux:copy",
+    "e2e:linux:build": "npm run start:qss && lerna run --scope @quiet/backend webpack:prod && lerna run --scope @quiet/desktop distUbuntu && lerna run --scope e2e-tests linux:copy",
     "e2e:linux:run": "lerna run --scope e2e-tests test --",
     "prepare": "husky",
     "lint-staged": "lerna run lint-staged",
+    "pull:submodules": "git submodule update --init --recursive --remote",
```
holmesworcester, line 17: ❤

packages/backend/src/nest/auth/sigchain.service.spec.ts
```diff
@@ -48,14 +48,14 @@ describe('SigChainManager', () => {
   it('should add a new chain and it be active if set to be', async () => {
```
holmesworcester, line 48: "and make it active"?

packages/backend/src/nest/connections-manager/connections-manager.service.ts
holmesworcester: Interesting. I think there might be a less ambiguous word than "Team" for that, if we're just talking about auth, especially if there's only one team per community. Like we should either say community or something really specific like "Security Perimeter" or "Perimeter" or something. We don't have to deal with it now but I bet this will be a source of confusion for newcomers at some point.

holmesworcester: What about "network"?

packages/backend/src/nest/qss/qss.service.ts
holmesworcester: Will QSS be able to add itself to be able to read messages in the general channel, e.g.? 

What will these secrets allow it to do, and not allow it to do?

holmesworcester (approved):

packages/backend/src/nest/qss/qss.service.ts
islathehut: No, servers can't have roles and can't get the member key.  They just get the team key and their own personal keys.  The only thing a server should be allowed to do with the chain is add new users to it but can't assign any roles.

adrastaea (approved):

PR #2879, opened 2025-06-03
Inject remote entries to OrbitDB manually

### LINKED ISSUE #2807: Consume sync data from QSS and write to respective OrbitDB stores
* Add logic to QSS service in Quiet to consume, process, and pass sync data to respective OrbitDB stores
* Explicitly we need to be able to handle the case where as a peer we can only receive a partial DAG. Will OrbitDB allow that natively?
* Add unit tests
* Tests should confirm data is not duplicated, data is successfully added to store and queryable, data should be deserializable from the store, data respective to each store type is appropriately distributed, and store update events that are normally triggered by the pubsub sync are also triggered by this method.

adrastaea:
### Pull Request Checklist


packages/backend/src/nest/common/utils.ts
```diff
@@ -257,6 +258,26 @@ export async function getLocalLibp2pInstanceParams(): Promise<Libp2pNodeParams>
   }
 }
 
+/**
```
islathehut, line 256: Woahhh I didn't know this existed!

packages/backend/src/nest/libp2p/libp2p.service.ts
```diff
@@ -96,14 +96,12 @@ export class Libp2pService extends EventEmitter {
...
         this.logger.warn('Redialing all known peers due to a server IO reconnect')
-        await this.hangUpPeers()
-        this.ensureDialQueueInterval()
       })
     })
   }
 
   public emit(event: string | symbol, ...args: any[]): boolean {
-    this.logger.info(`Emitting event: ${event.toString()}`, args)
+    // this.logger.info(`Emitting event: ${event.toString()}`, args)
```
islathehut: Should we just delete or make this a trace log?

packages/backend/src/nest/storage/orbitDb/orbitDb.service.ts
```diff
@@ -72,10 +79,93 @@ export class OrbitDbService {
...
+
+  private async joinHeads(address: string, heads: LogEntry[]): Promise<void> {
+    if (this.orbitDbInstance == null) {
+      throw new Error('OrbitDB instance is not initialized. Call create() first.')
+    }
+    const store = this.stores[address]
+    if (!store) {
+      this.logger.warn(`No store found for address ${address}, skipping join`)
+      // Keep heads in pendingHeads for later
+      this.pendingHeads.set(address, heads)
```
islathehut: We should repopulate this later if the app closes before writing these out to OrbitDB, right?  I'm trying to figure out if we need to write these to disk or not.

packages/backend/src/nest/libp2p/libp2p.service.ts
adrastaea: Ah yes, this snuck in just because it was kind of noisy when I was testing. Changing to a trace

packages/backend/src/nest/storage/orbitDb/orbitDb.service.ts
adrastaea: Hmm yeah we probably should. This exists largely as a mechanism for caching heads for messages while the channelsService creates their corresponding stores, so it should be fast in the case where the client has all the entries and will actually be able to join them to the log index. I think letting them drop in the case where they would never be able to join without being able to fully traverse the graph is fine. If the user happens to close out while the joins are processing, we should rehydrate. I'll add a write to disk to handle that.

packages/backend/src/nest/storage/channels/channels.service.spec.ts
```diff
@@ -132,6 +133,117 @@ describe('ChannelsService', () => {
...
+      })
+      expect(block2).toBeDefined()
+      expect(block2?.byteLength).toBeGreaterThan(0)
+
+      // reopen channel 1 and verify it is empty
+      logger.info(`Reopening channel ${channel1.id}`)
+      await channelsService.subscribeToChannel(channel1)
+    })
+
     // skipping because we don't have a strong way to prevent a user from deleting a channel yet
```
holmesworcester, line 246: I'm curious: why not?

packages/backend/src/nest/storage/channels/channels.service.spec.ts
```diff
@@ -253,4 +365,77 @@ describe('ChannelsService', () => {
...
+        payload: { value: encryptedMessage },
+        clock: { time: Date.now() },
+        hash: 'testhash',
+        bytes: Buffer.from('testbytes'),
+      }
+      // Mock joinEntry to track calls
+      const joinEntrySpy = jest.spyOn(store, 'joinEntry').mockResolvedValue(undefined)
+
+      // Call ingestEntries
+      await channelsService.ingestEntries([logEntry as any])
```
holmesworcester: I'll probably learn this later as I continue reading, but I'm curious why ingestEntries is part of channelsService and not more global. Like where will the ingestEntries function live for ingesting user profile photo updates, say, or some other thing that isn't in a channel?

packages/backend/src/nest/storage/channels/channels.service.ts
```diff
@@ -636,6 +688,14 @@ export class ChannelsService extends EventEmitter {
   public async clean(): Promise<void> {
     this.peerId = null
 
+    for (const [channelId, channel] of this.publicChannelsRepos.entries()) {
+      try {
+        this.logger.info(`Cleaning ${channelId} DB`)
```
holmesworcester, line 656: what does cleaning mean and why do we?

packages/backend/src/nest/storage/orbitDb/orbitDb.service.ts
```diff
@@ -72,10 +78,58 @@ export class OrbitDbService {
...
+    if (this.orbitDbInstance == null) {
+      throw new Error('OrbitDB instance is not initialized. Call create() first.')
+    }
+    const store = await this.orbitDbInstance.open<T>(address, options)
+    this.stores[address] = store
+    this.logger.info(`Opened OrbitDB store at address: ${address}`)
+    return store
+  }
+
+  public async ingestEntries(entries: LogEntry[]): Promise<void> {
```
holmesworcester, line 189: Ah okay, so this is where we make ingestEntries more general and not specific to channels or messages?

packages/backend/src/nest/libp2p/integration-tests/orbitdb-message-fanout.spec.ts
```diff
@@ -0,0 +1,896 @@
...
+  logger.info(
+    `Heads for ${store.id}:`,
+    heads.map((h: any) => ({ ...h, bytes: undefined }))
+  )
+  return heads
+}
+
+/**
+ * Performs a deep, order-insensitive equality check between two arrays.
+ * If an item has a 'hash' property, uses that as the comparison key; otherwise, uses JSON stringification.
```
holmesworcester: when would it not have a 'hash' property?

packages/backend/src/nest/libp2p/integration-tests/orbitdb-message-fanout.spec.ts
```diff
@@ -0,0 +1,896 @@
...
+
+/**
+ * Checks if all ChannelStores have the union of all entry hashes seen by any store.
+ * Optionally checks that each store has the expected number of entries.
+ *
+ * @param stores - Array of ChannelStore instances to check.
+ * @param expectedCount - Optional expected number of entries per store.
+ * @returns True if all stores have all entries, false otherwise.
+ */
+async function entriesAreFullyUnioned(stores: ChannelStore[], expectedCount?: number): Promise<boolean> {
```
holmesworcester: i don't get this yet. hmm.

holmesworcester: oh, is this simulating a network of several peers? and we're checking to make sure everyone has everything?

packages/backend/src/nest/libp2p/integration-tests/orbitdb-message-fanout.spec.ts
```diff
@@ -0,0 +1,896 @@
...
+      for (const libp2pService of modules.map(module => module.get(Libp2pService))) {
+        libp2pService.once(Libp2pEvents.PEER_DISCONNECTED, () => {
+          allDisconnected()
+        })
+      }
+      modules[0].get(Libp2pService).hangUpPeers()
+    })
+  })
+
+  // it('adds an initial entry to the mock store for all peers', async () => {
```
holmesworcester: what's all this commented-out code for?

holmesworcester (approved):
Looks good! None of my comments require attention. Just vibing and understanding what's going on.

islathehut:
@adrastaea There are a lot of errors on e2e tests, have you investigated those?  I see the tests fail on linux for the one client test and in my local testing the multiple clients test fails with similar errors.

EDIT: I did some digging and it seems like there's something going on in the order of operations when leaving the community.

packages/backend/src/nest/storage/channels/channels.service.spec.ts
adrastaea: Because we need to hook the permissions granted by LFA into the AccessController for the OrbitDB to prevent modification by users which do not have permission to delete the given entry. We can do some parsing above the DB level, but we need to reject entries that do not fit the access pattern to be consistent.

packages/backend/src/nest/storage/channels/channels.service.ts
adrastaea: Cleaning is deleting a db. We would do it when leaving a community.

packages/backend/src/nest/storage/orbitDb/orbitDb.service.ts
adrastaea: Yes.

packages/backend/src/nest/libp2p/integration-tests/orbitdb-message-fanout.spec.ts
adrastaea: tbh that comment was inserted by the AI agent. idk.

packages/backend/src/nest/libp2p/integration-tests/orbitdb-message-fanout.spec.ts
adrastaea: Yep

packages/backend/src/nest/libp2p/integration-tests/orbitdb-message-fanout.spec.ts
adrastaea: Oh that's leftover from my POC around behavior about injecting entries. I'll delete it because it's not relevant to testing production code.

packages/backend/src/nest/libp2p/integration-tests/orbitdb-message-fanout.spec.ts
holmesworcester: Okay, so the main thing is it's comparing two arrays in a test. That seems fine.

packages/backend/src/nest/storage/channels/channels.service.spec.ts
holmesworcester: Can you make an issue for this that describes what's missing, and give it the security label?

adrastaea: @holmesworcester https://github.com/TryQuiet/quiet/issues/2893

holmesworcester (approved):

PR #2915, opened 2025-07-02
Enforce UserProfile photo size limits

### LINKED ISSUE #2892: Joining users can't sync messages in existing channels
Version: 5.1.2
System: Probably all, as it has been spotted on macOS and Android. 

Issue: User on MacOS joined an already existing community (small but not new community with several channels). Once they joined, the welcome message on the general channel and the list of channels were appropriately loaded. None of the messages on any channel loaded. Logs from this user are attached below. 
An Android user joined the same community and had the same issue with syncing all messages beside those which were send by MacOS user. Those two users could communicate on general channel without issues. What's important, as mobile users can see preeviews of messages on channel list I can confirm that Android user recveived all messages from other users. They were just not displayed in channels.
The Owner confirmes that messages send by both Android and MacOS users are visible (screenshot below).

Logs from MacOS:

[log_2025-06-18.log](https://github.com/user-attachments/files/20821076/log_2025-06-18.log)

[error_2025-06-18.log](https://github.com/user-attachments/files/20821081/error_2025-06-18.log)

Screenshot from the Owner:

Image

Picture of general channel on MacOS (kinga2) and Android (kinga-android) users devices:

Image

adrastaea:
### Pull Request Checklist


packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -634,6 +643,7 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
     }
     const localAddress = createLibp2pAddress(onionAddress, peerIdData.peerId.toString())
 
+    const psk = 'v4UWpNX+SvCWs3eH3jatO/suEGFCkrHgUai24LtzyVc='
```
holmesworcester: testing cruft?

holmesworcester: If this got missed in review, are there security implications? Would communities end up using a hard coded PSK in some cases?

If this is even remotely possible (I think it is) we should always use env variables or debug panels for adding things like this and never modify the code itself, even if it feels like a temporary change we're not planning to ever submit.

packages/backend/src/nest/storage/userProfile/userProfile.store.spec.ts
```diff
@@ -142,7 +142,7 @@ describe('UserProfileStore/validateUserProfile', () => {
     // Changed the first byte from 137 to 136
     const pngByteArray = new Uint8Array([136, 80, 78, 71, 13, 10, 26, 10, 0, 0, 0, 13, 73, 72, 68, 82])
     const userProfile = await getUserProfile({ pngByteArray })
-    expect(await UserProfileStore.validateUserProfile(userProfile)).toBeFalsy()
+    expect((await UserProfileStore.validateUserProfile(userProfile)).success).toBeFalsy()
   })
 
   test('returns false if photo is larger than 200KB', async () => {
```
holmesworcester, line 150: there are some cases where the photo could be larger, because we are not strict with size limits when compressing user profile photos, IIRC. the sizes are targets but might fail sometimes. 

I think we should set the limit very high here, like equal to the limit on auto-download images, and create an issue for handling this the way we handle other images (CID => ipfs).

packages/backend/src/nest/storage/userProfile/userProfile.store.spec.ts
```diff
@@ -164,37 +164,37 @@ describe('UserProfileStore/validateUserProfile', () => {
...
   test('returns false if photo URL prefix is unexpected (invalid content-type)', async () => {
     const pngArray = [137, 80, 78, 71, 13, 10, 26, 10, 0, 0, 0, 13, 73, 72, 68, 82]
     // Invalid content-type
     const pngBase64 = 'data:text/html,' + Buffer.from(pngArray).toString('base64')
     const userProfile = await getUserProfile({ photoUrl: pngBase64 })
-    expect(await UserProfileStore.validateUserProfile(userProfile)).toBeFalsy()
+    expect((await UserProfileStore.validateUserProfile(userProfile)).success).toBeFalsy()
   })
 
   test('returns true if photo URL prefix is expected', async () => {
```
holmesworcester, line 196: Do we have a test to confirm that Alice cannot create updates for Bob's profile, that is that updates to Bob fail if not signed by Bob? Is this baked into write access control on the db?

(Also, how will we do this when we allow multiple devices?) 

I didn't see one in this file and it seems like it would be good to have.

packages/backend/src/nest/storage/userProfile/userProfile.store.ts
```diff
@@ -82,6 +130,12 @@ export class UserProfileStore extends EncryptedKeyValueStoreBase<EncryptedAndSig
     this.deferredProfiles = []
```
holmesworcester, line 88: Just a note but general, first-class support for "deferred" things like this that arrive out of order would be a great thing to have in a stack. Like, anything that doesn't validate gets checked again when new evidence arrives.

packages/backend/src/nest/storage/userProfile/userProfile.store.ts
```diff
@@ -95,52 +149,35 @@ export class UserProfileStore extends EncryptedKeyValueStoreBase<EncryptedAndSig
...
-      else if (encrypted.contents && typeof encrypted.contents === 'object' && !Buffer.isBuffer(encrypted.contents)) {
-        logger.debug('Converting object with numeric keys to Buffer')
-        const nums = Object.keys(encrypted.contents)
-          .filter(key => /^\d+$/.test(key))
-          .map(key => (encrypted.contents as any)[key] as number)
-        encrypted.contents = Buffer.from(nums)
-      }
-      logger.debug('Decrypting payload:', encrypted)
-      const decryptedPayload = this.auth.crypto.decryptAndVerify<UserProfile>(encrypted, payload.signature)
+      // Normaliz
```
holmesworcester: hm?

packages/backend/src/nest/storage/userProfile/userProfile.utils.ts
```diff
@@ -74,40 +75,47 @@ export const validatePhoto = (photoString: string, pubKey: string): boolean => {
...
-      return false
+      logger.error('Expected valid PNG or JPEG for user profile photo', pubKey)
+      return { success: false, error: 'Invalid photo format. Must be PNG or JPEG' }
     }
 
     // Apply different size limits based on image format:
-    // - For JPEG: 15MB limit (we compress these)
-    // - For PNG/GIF: 200KB limit (we don't compress these)
-    const MAX_SIZE_JPEG = 15 * 1024 * 1024 // 15MB in bytes
+    const MAX_SIZE_JPEG = 200 * 1024 // 200KB in bytes
```
holmesworcester, line 86: We should have a test where a user submits an 8MB image that is then compressed down to 200-300kb. This should work.

(Previously, when images were limited to 200k on the submission side, it was really annoying because it's hard to pick an already-compressed photo of yourself. So I added image compression a few releases ago.)

holmesworcester (changes requested):

packages/backend/src/nest/storage/userProfile/userProfile.utils.ts
adrastaea: Image compression has never actually been added to user profiles afaik though. It only affects files sent through messages and shared using the ipfsfilemanager. Unless that code change to the userprofile store got lost in a merge. I didn't see your image compression code touch the user profile store when I looked back at your prs though

holmesworcester (approved):

islathehut (approved):

PR #2918, opened 2025-07-04
add userprofile sidebar

### LINKED ISSUE #2920: Add a user profile list to sidebar
Context: 

We decided it would be helpful for a list of users to show in the sidebar, even if we don't have DMs yet. 

This feature was done opportunistically, so our criteria for validation is that it is subjectively not confusing. E.g. there should not be hover state if clicking on the usernames does not do anything. 

Opening the user profile view was considered as an option here.

adrastaea:
### Pull Request Checklist


packages/backend/src/nest/libp2p/libp2p.service.ts
```diff
@@ -155,12 +155,15 @@ export class Libp2pService extends EventEmitter {
     peerAddress: string,
     options: DialPeerOptions = { throwOnError: false, redialOnError: true }
   ) => {
-    if (this.connectedPeers.has(peerAddress.split('/').pop()!)) {
-      this.logger.debug(`Already connected to peer address: ${peerAddress}`)
+    const peerId = peerAddress.split('/').pop()!
+    if (peerId === '12D3KooWSGLPwT8vSmD9hp2umX6XKNVCQrneGJJbeFLdKTQ3DMjE') {
+      this.logger.info('Handling kingas peer address', peerAddress)
```
holmesworcester: is "kingas" cruft from debugging?

packages/backend/src/nest/libp2p/libp2p.service.ts
```diff
@@ -208,6 +211,11 @@ export class Libp2pService extends EventEmitter {
       return
     }
 
+    this.logger.info('Adding peers to dial queue', sortedPeers)
+    sortedPeers.push(
+      '/dns4/agrtyei6odml3ylmbfykcohzhd7n4wyxgdlcd5eamlmwdo2ciczchhqd.onion/tcp/80/ws/p2p/12D3KooWSGLPwT8vSmD9hp2umX6XKNVCQrneGJJbeFLdKTQ3DMjE'
```
holmesworcester: is this multiaddr more cruft?

packages/backend/src/nest/libp2p/libp2p.service.ts
```diff
@@ -375,7 +383,7 @@ export class Libp2pService extends EventEmitter {
     if (params.instanceName != null) {
       this.logger = this.logger.extend(params.instanceName)
     }
-    this.logger.info(`Creating new libp2p instance`)
+    this.logger.info(`Creating new libp2p instance with psk`, params.psk)
```
holmesworcester: We should avoid logging secrets, especially things that give access like PSK, so that when users send us logs or when we post logs to issues we aren't making communities public inadvertently.

packages/desktop/src/renderer/components/Sidebar/ChannelsPanel/ChannelsPanel.tsx
```diff
@@ -2,24 +2,33 @@ import React from 'react'
...
   currentChannelId: string
   createChannelModal: ReturnType<typeof useModal>
+  isTorInitialized: boolean
 }
 
 const ChannelsPanel: React.FC<ChannelsPanelProps> = ({
   channels,
+  myUserProfile,
+  userProfiles,
+  connectedPeers,
```
holmesworcester: A note: when we add QSS support we'll need to change the online status display to be based on some kind of heartbeat "i'm online" message (an orbitdb entry that for now can last forever but eventually will need a TTL) because we won't always be connecting to user devices directly.

packages/desktop/src/renderer/components/Sidebar/ChannelsPanel/UserProfileListItem.tsx
```diff
@@ -0,0 +1,109 @@
...
+  [`&.${classes.root}`]: {
+    width: 220,
+    padding: `3px 16px 3px 16px`,
+    gap: 8,
+    opacity: 1,
+    display: 'flex',
+    backgroundColor: 'inherit',
+    alignItems: 'center',
+  },
+  [`&:hover`]: {
```
holmesworcester: The hover style here is a bit confusing because it makes it seem that you can actually *do* something, which you can't yet. 

Maybe we should open the user's profile, even though it's obviously very limited for now? 

If that's bad for some reason, let's remove the hover states completely. We could also include a tooltip like "DMs coming soon!" though that might be ugly.

holmesworcester (changes requested):

holmesworcester:
Also I'm seeing test failures on this branch. Are they real?

packages/desktop/src/renderer/components/Sidebar/ChannelsPanel/UserProfileListItem.tsx
adrastaea: Added an onClick event that shows the user profile

islathehut (approved):

PR #2919, opened 2025-07-04
Fix/handle abrupt closes

### LINKED ISSUE #2921: Handle closing with ctrl+c
Currently, when the desktop application is force killed with ctrl+c, we leave tor processes hanging, don't save the sigchain, and otherwise lose state. We should intercept the SIGINT, SIGTERM, and unhandledRejection events, try to do a graceful close, and timeout if that takes too long.

adrastaea:
Gracefully closes backend when closed with ctrl+c, or when unexpected exception occurs. Prevents us from leaving hanging Tor processes or other backend processes.

### Pull Request Checklist


packages/backend/src/nest/auth/sigchain.service.ts
```diff
@@ -133,7 +133,7 @@ export class SigChainService extends EventEmitter {
       })) as User[]
     this.socketService.emit(SocketEvents.USERS_UPDATED, { users })
     this.emit('updated')
-    this.saveChain(this.activeChainTeamName!)
+    // this.saveChain(this.activeChainTeamName!)
     this.emit('updated')
```
islathehut: Looks like there's a dupe of the event emit, too.

holmesworcester:
Are these test failures meaningful?

packages/backend/src/nest/libp2p/libp2p.auth.ts
```diff
@@ -379,7 +379,7 @@ export class Libp2pAuth {
         this.sigChainService.setActiveChain(team.teamName)
       }
       this.joinStatus = JoinStatus.JOINED
-      this.sigChainService.saveChain(team.teamName)
+      // this.sigChainService.saveChain(team.teamName)
```
holmesworcester: I couldn't tell why these were all commented out, and what function was doing `saveChain` in their place. What's going on here?

holmesworcester (changes requested):
Two things:

1. That question about deleting the sigchain service. 

2. can we add some kind of e2e test to confirm shutdown was clean? Does that require new test tooling because it exists outside the running app? or is this easy?

packages/backend/src/nest/libp2p/libp2p.auth.ts
adrastaea: Oh sorry that was part of my testing to see if the chain was being saved before the app closes. I meant to uncomment those before merging. Theoretically, with the better shutdown process we don't need to be saving the chain every time it changes, and can just hold it in memory and defer saving it to disk until the app closes. Should be a lot more efficient that way.

holmesworcester: If the app gets killed is it still fine to not save it?

adrastaea: If the app gets a SIGKILL, then it will just instantly close and not save anything which is not ideal if we haven't been saving the chain along the way. Although, the main effect is that the user would look like they were back in time until someone gossiped the missing blocks to them again. That's why we should keep saving on chain updates for now, but in the future, we could probably relax how aggressively we save the chain and rely on the fact that in most every case of crash, quit, or interrupt that the first thing we do is save the chain.

adrastaea:
> 
> 2. can we add some kind of e2e test to confirm shutdown was clean? Does that require new test tooling because it exists outside the running app? or is this easy?
> 

I have added tests in the oneClient e2e test that test the different shutdown scenarios and main.test.ts provides some unit test support too for desktop.

packages/backend/src/nest/auth/sigchain.service.ts
islathehut: Just remove this duplicate emit and its good to go

islathehut (approved):

Issue #2820, opened 2025-04-25
Security concerns (library choice, lack of FS/PCS, replay attacks)


eragon4rya:
Is an audit planned?

fwding for @JavaDerg

![Image](https://github.com/user-attachments/assets/6c0f509e-ca91-473d-8260-9c6a8d2d219e)
side channel, depending on how this is used, pretty exploitable one as well lol

the good:
- they use libsodium internaly

the huh:
- they use some random library for "Decentralized authentication and authorization for team collaboration, using a secure chain of cryptological signatures." to encrypt their messages

the bad, as far as i can tell:
- no protection agains replay attacks
- no forward or post compromise secrecy
- group chats use a shared symetric key (i believe???), sortof, as far as i can tell they have generations, unsure how this is implemented the lib they use is weird
- dms are asymetricly encrypted, but as far as i can tell just libsodiums default asym encrypt, meaning all the above issues

eragon4rya:
There are working standards out there that "break" a private key after its been used X times, or after some time

JavaDerg:
Hi!

I feel like I should add some context.
Someone, I don't remember, and it doesn't matter, posted a link to this messenger in a joint group chat.
As I have an interest in secure & private messaging, I took a quick look and did a 5-minute-"audit".
As the chat contains a lot of other cryptography enjoyers, I posted these contents as sorts of preliminary findings.

> ![Image](https://github.com/user-attachments/assets/6c0f509e-ca91-473d-8260-9c6a8d2d219e) side channel, depending on how this is used, pretty exploitable one as well lol
> 
> the good:
> 
>     * they use libsodium internaly
> 
> 
> the huh:
> 
>     * they use some random library for "Decentralized authentication and authorization for team collaboration, using a secure chain of cryptological signatures." to encrypt their messages
> 
> 
> the bad, as far as i can tell:
> 
>     * no protection agains replay attacks
> 
>     * no forward or post compromise secrecy
> 
>     * group chats use a shared symetric key (i believe???), sortof, as far as i can tell they have generations, unsure how this is implemented the lib they use is weird
> 
>     * dms are asymetricly encrypted, but as far as i can tell just libsodiums default asym encrypt, meaning all the above issues

For various reasons beyond this the chat application is of no interest to, I would assume, most of us there, I can only speak for my self and a few people.
[...]

PR #2927, opened 2025-07-09
security(2820): More secure frontend<->backend socket secret generation/verification

islathehut:
### Description

* Use libsodium to generate socketIOSecrets on desktop and android (ios has some complications but we were already using the built-in `SecRandomCopyBytes` method to generate the random bytes which should be much better than what we were using on other platforms)
* Use libsodium to validate socketIOSecrets on all platforms
* Update length of socketIOSecrets from 20 to 32
* Update string encoding to hex for compatibility purposes with the android libsodium bindings
* Migrate token verification logic from common to backend
* Use updated LFA that includes libsodium helper functions because WASM hates me

### Pull Request Checklist


islathehut:
#2820

packages/mobile/android/app/src/main/java/com/quietmobile/MainActivity.kt
```diff
@@ -43,7 +43,7 @@ class MainActivity : ReactActivity() {
         val intent = intent
         checkAgainstIntentUpdate(intent)
 
-        if (BuildConfig.SHOULD_RUN_BACKEND_WORKER === "true") {
+        if (true) {
```
holmesworcester: What's this change for?

holmesworcester (changes requested):
I have questions about:

- The lazysodium library on Android
- Where exactly the compare function comes from now in the backend. I couldn't follow where that changed.

Let's discuss on voice!

packages/mobile/android/app/src/main/java/com/quietmobile/MainActivity.kt
islathehut: Oh whoops, this was accidentally committed.  For some reason this check breaks locally and I forgot to remove it before committing.

islathehut:
> I have questions about:
> 
> * The lazysodium library on Android
> * Where exactly the compare function comes from now in the backend. I couldn't follow where that changed.
> 
> Let's discuss on voice!

RE: compare: this is the compare method that comes from libsodium

EDIT: here's the line where it was added https://github.com/TryQuiet/auth/commit/05da5a04da99dbc45273d4c8961cd0fbb1663897#diff-34cdd7b338b2f611a6a4db5dec13bd0e178b8b524dd3074f8104cc5ed1082d47R24

packages/backend/src/nest/app.module.ts
```diff
@@ -125,14 +125,15 @@ export class AppModule {
               }
 
               const token = authHeader && authHeader.split(' ')[1]
+              _ioLogger.warn('secret, token', options.socketIOSecret, token)
```
adrastaea: I don't think we want to log the secret

adrastaea (changes requested):

packages/backend/src/nest/app.module.ts
islathehut: Gah another accidental commit.  An argument for not staying up until 3 while working.

holmesworcester (approved):

islathehut:
This was merged in the hotfix PR and brought into develop via 6.0.1

PR #2930, opened 2025-07-11
Security/Share socketIOSecret More Securely

### LINKED ISSUE #2931: Remove socketIOSecret from process args

adrastaea:
This replaces our previous method of sharing the socketIOSecret with process args with an IPC handshake which will prevent other processes from querying the process list and seeing the secret in the args list. There's a nonce exchange included which just provides some peace of mind that you're sharing and accepting with the same process though IPC should be relatively secure by default. If we feel later that we want to encrypt the exchange, then the nonce exchange is a solid stand in to be replaced with the encrypted exchange later.

* Adds secret exchange handshake over ipc
* Fixes rn_bridge communication with Android process
* Make rn_bridge message parsing more robust
* Adds handling on iOS for messages from rn_bridge
* Removes places where we put socketIOSecret in logs, redux store, url, or args to avoid accidentally exposing it
* Replaces Basic header with Bearer to be more compliant with standards since this is a token exchange not an encoded username:password pair

### Pull Request Checklist


packages/backend/src/backendManager.ts
```diff
@@ -33,33 +34,47 @@ program
...
+interface SecretMessage {
+  type: 'set-socket-secret'
+  secret: string
+  nonce: string
+}
+
+let secretReceived = false
 
-export const runBackendDesktop = async () => {
+function isSecretMessage(msg: any): msg is SecretMessage {
```
holmesworcester, line 54: Are we using the most high-level primitives for this we can be to avoid the "rolling our own" problem? What's the most boring and solid way to do this? Can/should we be using some feature of libsodium for this instead of Node.js webcrypto?

packages/backend/src/backendManager.ts
```diff
@@ -119,46 +128,85 @@ export const runBackendMobile = async () => {
...
-    throw error
+  let ipcNonce: string | undefined = randomBytes(16).toString('hex')
+  process.on('message', async msg => {
+    if (secretReceived) return
+    if (!isSecretMessage(msg)) return
+
+    secretReceived = true
+    const secret = msg.secret
+    if (msg.nonce && typeof msg.nonce === 'string') {
+      if (msg.nonce !== ipcNonce) {
```
holmesworcester, line 256: is this a place where we should be doing constant time string compare?

packages/backend/src/nest/app.module.ts
```diff
@@ -123,8 +122,15 @@ export class AppModule {
                 res.end()
                 return
               }
-
-              const token = authHeader && authHeader.split(' ')[1]
+              // Require Bearer token
+              const match = /^Bearer (.+)$/.exec(authHeader)
+              if (!match) {
```
holmesworcester, line 127: is this another comparison function we need to be careful about?

packages/backend/src/rn-bridge.ts
```diff
@@ -1,259 +1,204 @@
...
-        }
-        parsed[split[0]] = split[1]
-      })
+interface MessageEnvelope {
+  event: string
+  payload: string | Record<string, string>
+}
 
-      logger.info('parsed', parsed)
+class MessageCodec {
```
holmesworcester, line 28: what is this for? 

is there a naming issue here? are these messages in the "you have new messages in Quiet" sense? or in a socket/IPC sense?

packages/desktop/src/main/main.test.ts
```diff
@@ -271,3 +272,32 @@ describe('Invitation code', () => {
...
+    expect(loadURL).toBeDefined()
+    // Find the call to loadURL and extract the URL
+    const loadURLCalls = loadURL.mock.calls
+    expect(loadURLCalls.length).toBeGreaterThan(0)
+    const urlArg = loadURLCalls[0][0]
+    // Should NOT contain socketIOSecret=...
+    expect(urlArg).not.toMatch(/socketIOSecret=[a-f0-9]{64}/i)
+  })
+
+  it('does not leak socketIOSecret in backend process fork arguments', () => {
```
holmesworcester, line 358: anywhere else we should check?

packages/mobile/android/app/src/main/java/com/quietmobile/Backend/BackendWorker.kt
```diff
@@ -54,44 +109,50 @@ class BackendWorker(private val context: Context, workerParams: WorkerParameters
...
+                        .setContentTitle(title)
+                        .setTicker("Quiet")
+                        .setContentText("Backend is running")
+                        .setSmallIcon(icon)
+                        // Add the cancel action to the notification which can
+                        // be used to cancel the worker
+                        // .addAction(android.R.drawable.ic_delete, "cancel", intent)
+                        .build()
 
         val id = ThreadLocalRandom.current().nextInt(0, 9000 + 1)
```
holmesworcester, line 140: what's this for?

holmesworcester (changes requested):
Two things:

1. Those string comparison questions. 
2. I think it might be helpful if there was some doc that explained how all this is intended to work, so that we have some reference to return to in the future. This could be in the repo or in the wiki. Probably in the repo is best. Thoughts? We could also include a lot more comments in the code itself.

packages/backend/src/backendManager.ts
islathehut: Related to this I wonder if we should be encrypting the ipc and local socket messages.  Skip the "secret" and just properly encrypt.  We could use a scheme like I had on QSS previously where we do a public key exchange to generate a shared secret for encryption so we get the ephemerality of the existing `socketIOSecret` but the communications are private.  Snooping on an unsecured local websocket isn't hard if you have access to the device.

Doing this would require either some research on encryption in IOS or switching to something other than libsodium since support on IOS is....bad.

packages/backend/src/nest/app.module.ts
islathehut: This just appears to validate the format of the header and not the actual token.

packages/backend/src/backendManager.ts
islathehut: In the current implementation we should be using the same methods for generating and comparing the nonce that we use for the secret.

packages/backend/src/rn-bridge.ts
adrastaea: This class holds decoding functions for ipc messages on the rn_bridge used by iOS and Android. They have some quirks so they need a custom decoder. This is basically all existing code, but I rearranged it so that linting works.

packages/mobile/android/app/src/main/java/com/quietmobile/Backend/BackendWorker.kt
adrastaea: These changes are mostly formatting done by the linter.

packages/backend/src/backendManager.ts
adrastaea: I don't think that a constant time string compare is really an issue here because it's a one time exchange that should resolve almost instantly and then ignore all further attempts to share a secret. There's no real way to brute force the nonce because the main app will finish the secret exchange before any attacker could brute force it. Also, the IPC channel should be private anyway afaik, so this is more of a gut check as is.

adrastaea: I support the idea of encrypting socket comms in a future feature though.

packages/desktop/src/main/main.test.ts
adrastaea: I don't think so

packages/backend/src/backendManager.ts
adrastaea: I think so. This is relatively simple. The important thing from what I understand is to pass it over IPC because that should be private at the OS level unless the attacker has root or admin privileges. The nonce is peace of mind and placeholder for future encrypted exchanges if we get more paranoid, and the `isSecretMessage` validator is just to make the linter not yell at me

packages/backend/src/backendManager.ts
islathehut: I can get behind that.  Since this is one time only its not really worth using libsodium for.

islathehut (approved):

PR #2944, opened 2025-07-22
chore(2912): Swap to QSS_ALLOWED flag and reconfigure how connections/events determine if qss is enabled

### LINKED ISSUE #2912: Clarify QSS flags
We need to change the QSS_Enabled feature flag to QSS_ALLOWED to signify whether to show QSS related features, and have a QSS_ENABLED flag which is flipped according to opt-in.

islathehut:
### Pull Request Checklist


packages/backend/src/nest/qss/qss.service.spec.ts
```diff
@@ -84,31 +85,38 @@ describe('QSSService', () => {
     if (mockedSendMessage != null) {
       mockedSendMessage.mockRestore()
     }
+    if (mockedAllowed != null) {
+      mockedAllowed.mockRestore()
+    }
   })
 
   describe('connect', () => {
```
holmesworcester, line 93: should we add "QSS_ALLOWED" to these tests too, so that we make sure it can't be enabled when not allowed?

holmesworcester (changes requested):

packages/backend/src/nest/qss/qss.service.spec.ts
islathehut: This should already be accounted for on all of the tests, e.g. `mockedAllowed = jest.spyOn(qssService, 'qssAllowed', 'get').mockReturnValue(false)`

islathehut: I added that getter specifically for easier mocking since we resolve environment variables when initializing the module which makes things a bit tricky from a testing standpoint.  All it does is return the value stored in QSS_ALLOWED.

holmesworcester (approved):

Issue #2423, opened 2024-04-10
App doesn't sync old messages 


kingalg:
**Version:** mobile@2.2.0-alpha.0
**System:** spotted one time on android

**How to reproduce** - I can't reproduce it, it happened to me one time; below you can see logs in the attachment

**What happened:** The Android app doesn't sync old messages and avatars but does sync new messages without any issues. 

[logs1004.txt](https://github.com/TryQuiet/quiet/files/14931039/logs1004.txt)

Issue #2413, opened 2024-04-05
Messages/channels not syncing reliably between users


islathehut:
This one is a little tricky to reproduce.  Connect multiple users and send messages from each user, disconnect users, etc. and you should see messages not sync between all users at all times.

holmesworcester:
Are you seeing situations where users A and B can both see messages from user C but not from each other? 

Or situations where A can see messages from B but not vice versa?

Or is it possible this is caused just by one peer not connecting or losing its connection?

PR #2954, opened 2025-08-06
Fix Race Conditions in Message Verification and User Profile Indexing

### LINKED ISSUE #2847: Messages may be received before UserProfile in Race Condition
Once a new user joins the chain, they start syncing all their local orbitDB databases at once. This means that they may share their intro message before their UserProfile, and cause their peers to invalidate their intro message because they cannot connect it with a username. We can either defer messages we can't get a profile for until a new profile is added and then try again, or less ideally just display the bare userID. This could also potentially happen with messages from a disconnected peer who are on the LFA signature chain, but while we are in the process of recomputing the team state with them in it we get an orbitDB block from them that we can't decrypt yet.

adrastaea:
Implements event based re-indexing of UserProfileStore and verification of previously unverifiable messages. Solves the issue of messages or user profiles received before the corresponding auth blocks not having a retry mechanism to add them into the state when those blocks are received.

### Pull Request Checklist


adrastaea:
Issues that this may resolve:
#1604, #2413, #2423, #2847

islathehut (approved):

PR #2966, opened 2025-08-13
Fix/mac autoupdater

### LINKED ISSUE #2965: Autoupdater is broken on mac

adrastaea:
Fixes an issue where triggering an update would stall out on Mac (and maybe others) because of the preventDefault guards that let us intercept default behavior and do cleanup or minimize instead of close in the case of macOS. 

* Adds an `updating` flag which is set before `app.quitAndInstall()` and lets other events be more permissive of default close behaviors when installing an update.

### Pull Request Checklist


packages/desktop/src/main/main.ts
```diff
@@ -560,6 +574,7 @@ app.on('ready', async () => {
 
   ipcMain.on('proceed-update', () => {
     logger.info('ipcMain: proceed-update')
+    updating = true
```
islathehut, line 577: Do we need to reset this to false anywhere?

adrastaea: I don't think so because the application should fully close and reinitialize. I can't actually test this without pushing out a new alpha though.

PR #3004, opened 2025-10-06
chore(2916): Handle QSS reconnects/retries and fixes for connections

islathehut:
### Pull Request Checklist


packages/backend/src/backendManager.ts
```diff
@@ -262,11 +263,7 @@ if (platform === 'desktop') {
     runBackendDesktop(secret).catch(async error => {
       logger.error('Error occurred while initializing backend', error)
       // Prevent stopping process before getting output
-      await new Promise<void>(resolve => {
-        setTimeout(() => {
-          resolve()
-        }, 10000)
-      })
+      await sleep(10_000)
```
adrastaea, line 266: Just for discussion, should we maybe only sleep like this in development? I don't think there's value in waiting in production

packages/backend/src/nest/qss/qss-auth-conn.ts
```diff
@@ -147,43 +178,27 @@ export class QSSAuthConnection extends EventEmitter {
...
       this._active = true
-      if (this.sigChainService.activeChainTeamName != null) {
+      if (this.sigChainService.activeChainTeamName != null && this._joinStatus !== JoinStatus.JOINED) {
         this.logger.debug(`Sending sync message because our chain is initialized`)
         const sigChain = this.sigChainService.getActiveChain()
-        const team = sigChain.team
+        const team = sigChain.team!
         const user = sigChain.user
-        this.authConnection.emit('sync', { team, user })
+        this._authConnection!.emit('sync', { team, user })
```
adrastaea: This should probably be an if statement not ! incase the Connection assigned to this._authConnection is replaced and we get a race condition. I also think that as written you could theoretically end up with one Connection state machine emitting connected causing a totally different Connection state machine to emit `sync`. I believe you could set this up so that the reference is not dynamic to fix that.

adrastaea (approved):

packages/backend/src/backendManager.ts
islathehut: We should generally avoid it but this should be the same functionality as before just standardized.

packages/backend/src/nest/qss/qss-auth-conn.ts
islathehut: Updated

PR #3012, opened 2025-10-22
fix: Better QSS handling in invite links, proper redialing on disconnects, resetting LFA join status when in intermediate state on disconnect

### LINKED ISSUE #3024: Quiet doesn't accept AWS links in invites
The invite link validator breaks when using QSS endpoints with port 80

### LINKED ISSUE #3025: Client fails to connect to peers when using AWS QSS
Dialing is degraded and LFA joining fails when AWS QSS is used.  This is due to a race condition that causes join status on the libp2p auth service to get into a bad state.

islathehut:
### Pull Request Checklist


packages/backend/src/nest/libp2p/libp2p.auth.ts
```diff
@@ -414,6 +414,25 @@ export class Libp2pAuth {
...
+      return
+    }
+
+    /**
+     * We need to manually reset the join status in the case where the status is stuck on an intermediate state like
+     * JOINING when disconnecting (for example this can happen when the user you are connecting to doesn't have your
+     * information in their chain yet resulting in an invalid device error)
+     */
+    const oldJoinStatus = this.joinStatus
+    if (this.sigChainService.team == null) {
```
adrastaea: I believe this condition will always be true and if it's not then it will throw an error because SigchainService.team returns `this.getActiveChain().team!` which asserts that it is true.

adrastaea (changes requested):
One change needed in onPeerDisconnected to avoid an uncaught exception if the user disconnects from a peer while their chain context is in the InviteeMemberContext state and does not have team defined in their sigchain, but otherwise fine

adrastaea (approved):

PR #3031, opened 2025-11-19
add mise-en-place support for simpler SDK dependency management

mcginty:
I previously used `just` and `.envrc` to make it easier to set environment variables and tasks across my macOS/Linux/Windows machines, and discovered a newer tool `mise` (https://mise.jdx.dev/) along the way that works quite well with the Quiet codebase.

It can install and load/unload pretty much all the dependencies Quiet uses in a way that can also be used by CI.

It is especially helpful at drastically simplifying the Android developer setup by automatically installing the correct Temurin JDK and Android SDK and setting the environment variables as expected whenever you enter the quiet workspace.

This would replace other recommended installation methods like SDKMAN/Volta/rbenv/etc and turn developer/CI setup into a one-liner (after installing mise):

```
# (install mise via package manager or `curl https://mise.run | sh`)
mise install
```

While nix is very powerful, it's an absolutely pain to install and maintain, and this feels a lot more accessible.

I haven't updated any `README`s yet until I get feedback one way or another on if this is desirable to merge :).

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,0 +1,119 @@
+min_version = "2024.9.5"
+
+[env]
+
+[hooks]
+
+[tools]
+# general dev/sdk tools & dependencies
+usage = "latest" # needed by mise for shell completions
+yq = "latest" # needed by mise for android-sdk
+git-lfs = "latest"
+
+# project-wide dependencies
+node = "18.20.4"
+npm = "10.7.0"
+"npm:lerna" = "6.6.2"
+"npm:typescript" = "4.9.5"
+"npm:pnpm" = "9.12.1"
+
+# android dependencies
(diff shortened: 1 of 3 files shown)


adrastaea (approved):

holmesworcester:
If there are no issues with this on Ubuntu or Mac and it simplifies things, let's do it.

mcginty:
I tested it on Linux (Arch-based, I'll make sure it works on Ubuntu before merging too) and macOS!

There's one small hiccup which I confirmed is a `mise-android-sdk` plugin bug where you might have to run `mise install` twice depending on if the OS has the correct version of `yq` installed already, and I contacted the dev about it for a fix, but it's not a blocker and should be resolved soon.

PR #3011, opened 2025-10-20
Intertwined Mobile Updates (Android 16KB Alignment, Necessary iOS changes, React Native 0.77)

bitmold:
Here is my branch after merging the latest `develop` into the project.

As of right now, I'm working out some issue where I can't successfully run `bundle exec pod install` after merging `develop`. So will update here but posting for the time being.

After you bootstrap and do everything to configure the project:

```bash
cd packages/mobile/android 
./gradlew clean assembleDebug
cd ../../../scripts
./check_elf_alignment.sh ../packages/mobile/android/app/build/outputs/apk/standard/debug/app-standard-debug.apk
```

We can see that everything is 16KB Aligned:

```
➜  scripts git:(react-native-upgrade77) ✗ ./check_elf_alignment.sh ../packages/mobile/android/app/build/outputs/apk/standard/debug/app-standard-debug.apk

Recursively analyzing ../packages/mobile/android/app/build/outputs/apk/standard/debug/app-standard-debug.apk

NOTICE: Zip alignment check requires build-tools version 35.0.0-rc3 or higher.
  You can install the latest build-tools by running the below command
  and updating your $PATH:

    sdkmanager "build-tools;35.0.0-rc3"

=== ELF alignment ===
/var/folders/51/8996x1_x5bj3khs7h33_s6y40000gn/T/app-standard-debug_out_XXXXX.a3vMbdnVBz/lib/arm64-v8a/libnode.so: \e[32mALIGNED\e[0m (2**14)
/var/folders/51/8996x1_x5bj3khs7h33_s6y40000gn/T/app-standard-debug_out_XXXXX.a3vMbdnVBz/lib/arm64-v8a/libgifimage.so: \e[32mALIGNED\e[0m (2**14)
/var/folders/51/8996x1_x5bj3khs7h33_s6y40000gn/T/app-standard-debug_out_XXXXX.a3vMbdnVBz/lib/arm64-v8a/libjsi.so: \e[32mALIGNED\e[0m (2**14)
/var/folders/51/8996x1_x5bj3khs7h33_s6y40000gn/T/app-standard-debug_out_XXXXX.a3vMbdnVBz/lib/arm64-v8a/libnative-filters.so: \e[32mALIGNED\e[0m (2**14)
/var/folders/51/8996x1_x5bj3khs7h33_s6y40000gn/T/app-standard-debug_out_XXXXX.a3vMbdnVBz/lib/arm64-v8a/libfbjni.so: \e[32mALIGNED\e[0m (2**14)
[...]

PR #3035, opened 2025-11-29
Remove Unused Leveldown Binaries and Config from Desktop

bitmold:
In Quiet, `leveldown` was replaced with a (now pretty outadted) version of `classic_level` a few years ago. 

This PR removes some unsued json+js, as well as unused `leveldown` binaries for mac + windows + linux , hopefully simplifying some of the confusion around Quiet's leveldb integration.

Am able to run `npm run start:desktop` and upload images just fine without these.

You'll notice that binaries for iOS and Android still exist on this PR, that's because similar work to get rid of `leveldown` on mobile is in my mobile PR #3011.

Haven't tested this locally on Windows like i have linux + mac..., but the Windows CI seems fine


bitmold:
cc @holmesworcester this PR doesn't do anything other than remove unused leveldown DB binaries (that are named *very similarly* to binaries Quiet actually uses and of course **needs** to function - _the `leveldown` bins being deleted here were in fact confusingly replaced with `classic-level` binaries, which sound older because of the *classic* but actually in fact are the newer ones used by the app...._.) ---  I spent a nontrivial amount of time trying to make sense of the different leveldown/classic-level binaries when working on #3011. once i realized i eventually needed to rebuild classic-level on android, i wasted time not knowing which binary i'd need to swap out for my modified one, etc.

once this is merged, other devs won't have to waste time trying to wrap their head around which bin is doing what like i did

bitmold:
Closing, just threw these 2 commits on this forgotten PR onto the related #3092

PR #3048, opened 2025-12-12
IPFS Profile Photos

### LINKED ISSUE #2926: User should be able to add large profile image
Right now our image compression service is built around the ipfs file manager service, but user profile photos do not use this. (They should anyway.)

Validation criteria:

1. Quiet uses IPFS file manager for user profile images.
2. Quiet compresses user profile images on all platforms, adds to IPFS, and includes file metadata in user profile.
3. User profile images must either be jpg or if png/gif be smaller than the max threshold for auto download. (Consider requiring jpg.) 
4. Validation is correct

mcginty:
### Migration

This update deprecates the previous profile photo method of setting a base64-encoded image URI within the profile metadata, in favor of using the IPFS attachment upload pipeline.

Old profile photos are still displayed as expected, but when a new profile photo is set, it will *only* use the new IPFS flow, so older clients will not see a photo. I can change this to do both at the same time, but that feels like it defeats the purpose a bit, and we're at the turn of a new major version here as well.

### Approach

This creates a new "virtual channel" for attachment photos so that we can reuse as much of the same image attachment processing pipeline as possible.

Some things obviously couldn't be handled the same, though, so you'll notice some conditional handling of profile photo attachments that stray from the typical assumption that an attachment is associated with a message in a viewable chat room.

Fixes #2926.

### DIFF
--- a/packages/state-manager/src/sagas/users/users.master.saga.ts
+++ b/packages/state-manager/src/sagas/users/users.master.saga.ts
@@ -3,6 +3,7 @@ import { all } from 'typed-redux-saga'
 import { type Socket } from '../../types'
 import { usersActions } from './users.slice'
 import { saveUserProfileSaga } from './userProfile/saveUserProfile.saga'
+import { downloadProfilePhotosSaga } from './userProfile/downloadProfilePhotos.saga'
 import { createLogger } from '../../utils/logger'
 
 const logger = createLogger('usersMasterSaga')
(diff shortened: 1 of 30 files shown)


packages/state-manager/src/sagas/users/users.master.saga.ts
```diff
@@ -3,14 +3,18 @@ import { all } from 'typed-redux-saga'
...
 import { createLogger } from '../../utils/logger'
 
 const logger = createLogger('usersMasterSaga')
 
 export function* usersMasterSaga(socket: Socket): Generator {
   logger.info('usersMasterSaga starting')
   try {
-    yield all([takeEvery(usersActions.saveUserProfile.type, saveUserProfileSaga, socket)])
+    yield all([
+      takeEvery(usersActions.saveUserProfile.type, saveUserProfileSaga, socket),
```
holmesworcester, line 15: Claude insists that this creates nested watchers:

>   This creates nested watchers - every updateUserProfiles dispatch spawns a new takeEvery watcher. Should either be fork(downloadProfilePhotosSaga) once, or the inner saga should not use takeEvery.

Is this a valid issue or a false positive?

adrastaea (approved):
I pushed a little update to the e2e test that fixed the failure. LGTM otherwise

mcginty:
You're a hero @adrastaea

packages/state-manager/src/sagas/users/users.master.saga.ts
mcginty: Ahhhh, after looking a bit into how this works, I think this is in fact an issue, which will cause new permanent child tasks to keep spawning with every run. Nice catch.

adrastaea:
One note, please update the changelog before pushing

PR #3061, opened 2026-01-14
Feat/2806 log entry pull

### LINKED ISSUE #2806: QSS: User should be able to catch up on OrbitDB data since last sign-in
* Add logic to QSS to stream latest sync data that the user missed since they last logged in
* Add tests

My gut says that we can, at least to start, just track the last timestamp we got a sync from QSS and pass that in on sign-in to QSS.  That keeps things simple without a big performance hit.

adrastaea:
Adds client side logic for pull ops on the orbitdb logs in qss.

* Starts an interval to pull new logs once auth connection is established
* Iterates over pages until caught up
* QSSAuthConnection emits connected and disconnected events
* Adds a serializer to deserialize packed log entries from qss

### Pull Request Checklist


packages/backend/src/nest/qss/qss.service.ts
```diff
@@ -481,6 +496,44 @@ export class QSSService extends EventEmitter implements OnModuleDestroy, OnModul
     return true
   }
 
+  private async _pullLatestLogEntriesForTeam(teamId: string): Promise<void> {
```
holmesworcester, line 510: Is it possible for `pullLatestLogEntries` to happen before all the latest team key is synced? 

Either because QSS hasn't synced the latest team key from the sender yet, or because we haven't synced it from QSS yet? Or some other reason?

And if that happens, would the message be undecryptable?

holmesworcester: Claude suggests this case:

 Timeline for New User Joining via QSS:
  ──────────────────────────────────────────────────────────────────────────────►

  T=0: User clicks invite link, connects to QSS

  T=1: LFA auth connection starts
       └─► 'connected' event fires
       └─► QSS_AUTH_CONNECTED emitted

  T=2: Pull interval starts (because QSS_AUTH_CONNECTED fired)
       └─► pullLatestLogEntries() called

  T=3: QSS returns encrypted LogEntries (encrypted with MEMBER key)

  T=4: User tries to decrypt...
       └─► FAILS! User doesn't have MEMBER key yet!
       └─► Entry silently dropped (returns null)

  T=5: LFA auth sync continues...
       └─► 'joined' event fires
       └─► _joinStatus = PENDING_MEMBER (no team/keys yet)

  T=6: More LFA sync messages...
       └─► User receives MEMBER role keys via lockbox

  T=7: User now has MEMBER keys
       └─► But messages from T=3-T=4 were ALREADY DROPPED!

holmesworcester (changes requested):
I had one specific question (see comment below) but also I have a general question about validation: we have a few different validation paths now, right? 

Is that going to cause problems for us? 

Is there anything that will validate on one path but not on the other, in a way that could create weirdness for users or attack surface?

The most important question I think is whether any invalid data get added to OrbitDB via QSS or if it goes through the same validation?

packages/backend/src/nest/qss/qss.service.ts
```diff
@@ -669,12 +742,127 @@ export class QSSService extends EventEmitter implements OnModuleDestroy, OnModul
...
+    const finalPullResponse: LogEntryPullResponseMessage = {
+      ts: DateTime.utc().toMillis(),
+      status: CommunityOperationStatus.SUCCESS,
+      payload: {
+        entries: [],
+        hasNextPage: false,
+      },
+    }
+    this.logger.info(`Completed pulling all log entries from QSS for team ${teamId}`)
+    return finalPullResponse
```
islathehut, line 873: Can we clear the interval at this point?  Once we pull the entries we've missed we should be getting new messages in realtime so we don't need to pull constantly.

packages/backend/src/nest/qss/qss.service.ts
```diff
@@ -669,12 +742,127 @@ export class QSSService extends EventEmitter implements OnModuleDestroy, OnModul
...
+      const decryptedEntries = await Promise.all(
+        deserializedEntries.map(entry => {
+          try {
+            const decrypted = this.sigChainService
+              .getChain({ teamId })
+              .crypto.decryptAndVerify<LogEntry>(entry.encrypted, entry.signature, false)
+            if (decrypted.isValid) {
+              return decrypted.contents
+            }
+            return null
```
islathehut: We should shove the failed messages into a dead letter queue to be processed later.  It seems like the way its written now we'll drop these messages entirely and because the last sync time is set at the end of the entry ingestion we won't retry pulling these messages in the future from QSS.  Like in Holmes' comment above in the future when the keys can be rotated we may have messages that we can't decrypt now but will be able to decrypt in the future so we want to be able to retry without having to connect to a peer.

islathehut (changes requested):

islathehut (approved):

PR #3063, opened 2026-01-15
Add extra secret salt to pull MEMBER role keys from QSS

### LINKED ISSUE #3056: Creating invite links for non-P2P authentication
# Overview

We need to include a second piece of information on invite links for identifying the keys for an invite on the link.  Invite links with this piece of data will allow users to pull the `MEMBER` role keys without having a peer connection.

# Required Work

- Update `createLongLivedUserInvite`
- Update `invitationLink.ts` with `salt`
- Update `connectionSlice`
- Update `connection.selectors.ts`

# Notes

* In the design I specify using a randomly generated salt for specifying the key scope but this is kind of arbitrary.  It just has to be a value that any joining user has so they can generate the keys.

# Relevant Docs

* https://www.notion.so/QSS-Authentication-Without-P2P-2da5bfa9e19c8023960bc30f8d56e85d?source=copy_link#2df5bfa9e19c80bf9304dcb5a8e5b78b

mcginty:
Closes #3056

Based on extending LFA's `InviteResult` to include an associated salt.


islathehut (approved):

islathehut:
Once unit tests are fixed this will be good to merge

PR #3090, opened 2026-01-22
16kb Aligned React Native

bitmold:
### Pull Request Checklist

This bumps react native to 0.77.3, which aligns many native binaries on Android. It contains necessary JavaScript, Android-specific, and iOS-specific changes. 

Making this as a draft right now...

This should be merged after the work of #3065 and #3067 has been tested and looks good. Just getting this ready right now. I also need to rebuild the project on iOS and test against other changes that have been merged from the past week, but I expect that to be fine. Will update this once I do though.

PR #3065, opened 2026-01-16
Preliminary Android/Gradle Changes from the big 16KB Alignment PR

bitmold:
This PR grabs a lot of the initial work from #3011 that pertained to the android project. Nothing in JavaScript is changed here from `develop`, so these changes shouldn't touch the iOS project and it should be identical to building the iOS project on `develop`.

Lots of the individual pieces that make up the mega 16kb alignment PR depend on this work, making this PR a foundation for a bunch of smaller incoming Android changes...

## Non-code changes:
- README updates from #3011
- Android build scripts from #3011, including Google's script to check ELF alignment. Included my documentation for the weirder scripts
- From #3011 Comment out android part of the top-level script to update tor binaries from Tor Browser releases - this breaks the 16KB Aligned binary I built for Quiet. Add some clarifying language for when it updates tor binaries on mac/linux/windows
- Remove unused Guardian Project (formerly for GP's tor-android) and Jitpack (Github repo as Gradle DEP) gradle repositories from top-level Gradle project on Android 
- **Significantly Cleaned up and Modernized `build.gradle`, `settings.gradle` and most importantly `app/build.gradle**.
  - The broken Groovy syntax Quiet used was very inconsistent. The syntax is now uniform and similar to the modern Kotlin DSL language, which is what Google wants Android apps to use and will make upgrading sometime way easier. Google encourages people to make this change before modernizing their gradle projects.
  - Fixed/Removed a lot of deprecated Gradle features which give warnings today, but ultimately will give errors/be unsupported when a future release of the Android Gradle Plugin Quiet will be forced to upgrade to is released. With these changes, a lot less gradle stuff (if any) will need to be changed to keep the app building whenever AGP 10 is released. In Android Studio the scrollbar no longer has massive yellow blocks of warnings.
  - Minor: Fixed Kotlin/Java syntax, warnings 
  - Bump to android gradle plugin 8.6.1 from 8.5.0, this is still pretty old, but is supported in modern Android Studios, when we upgrade to react native 0.77 we can opt for a much newer version...

### Code Changes to App
- Bumped various gradle dependencies for the Quiet android app to more recent versions 
[...]

PR #3092, opened 2026-01-23
16kb Aligned Classic Level Binary 

bitmold:
My build script and recompiled classic-level binary. Also all the work to clear up redundant + confusing unused duplicate classic-level + leveldown binaries on iOS and Android. Should be merged once the work contained in #3065 and #3067 is tested. 

This binary needs to be aligned for the play store, but Android's script for checking binary alignment in an APK doesn't take `.node` files into account. 

This also includes my documentation for building a nodejs module with C++ code (classic-level) against our nodejs-mobile headers, just in case this confusing thing ever needs to happen again someday. 

Need to re-test this branch on IOS. Also making this a draft for the time being, like #3090, but this work is 99.9% done.


bitmold:
@islathehut i just confirmed that this still works on iOS, however I'm going to keep it as a draft until #3090 is merged. But it's ready once that one is !

bitmold:
@islathehut @holmesworcester this is ready for merging. Once this is in 100% of android is 16KB Aligned, and you also have a bunch of unused and duplicate executables for every desktop + mobile platform removed from the repo. Just re-tested on Android and iOS

islathehut (approved):

PR #3114, opened 2026-02-25
Feat/3086 client push service

### LINKED ISSUE #3086: Create Push Notification Service
# Summary

Create a shared service in packages/backend that handles calling QPS to trigger push notifications for a community. This service will be used by all platforms.

# Background

The push trigger logic should live in the shared backend code so desktop, Android, and iOS all use the same implementation

# Acceptance Criteria
- [ ] PushNotificationService class created in packages/backend
- [ ] Service listens for ack from QSSService of upload
- [ ] In p2p-only mode, push is triggered when self sends a message
- [ ] HTTP client configured with appropriate timeout (5 seconds)
- [ ] Success response (200/204) handled correctly
- [ ] Error responses mapped to typed errors
- [ ] QPS base URL configurable via environment
- [ ] Unit tests with mocked HTTP

adrastaea:
### Pull Request Checklist


packages/backend/src/nest/qps/qps.service.ts
```diff
@@ -49,53 +57,142 @@ export class QPSService implements OnModuleInit {
...
+      `Triggering push notifications for team ${teamId} with ${ucans.length} UCAN(s) in ${batches.length} batch(es)`
+    )
+    for (const [index, batch] of batches.entries()) {
+      try {
+        const response = await this.qssClient.sendMessage<SendBatchPushResponse>(
+          WebsocketEvents.SEND_BATCH_PUSH,
+          {
+            ts: DateTime.utc().toMillis(),
+            status: CommunityOperationStatus.SENDING,
+            payload: { ucans: batch, title, body, data },
```
mcginty, line 139: I forget where we landed on QSS "team awareness" - do we need a `teamId` sent in for batches?

packages/backend/src/nest/qps/qps.service.ts
```diff
@@ -41,6 +48,7 @@ export class QPSService implements OnModuleInit {
     })
 
     this.qssClient.on(QSSEvents.QSS_CONNECTED, () => this._flushPendingToken())
+    this.qssClient.on(QSSEvents.QSS_LOG_SYNCED, (teamId: string) => void this.sendBatchPush(teamId))
```
mcginty, line 51: Just to confirm - `QSS_LOG_SYNCED` is triggered once per message?

packages/backend/src/nest/qss/qss.service.ts
```diff
@@ -449,9 +449,14 @@ export class QSSService extends EventEmitter implements OnModuleDestroy, OnModul
       }
     }
 
-    // we need to normalize the hostname for QSS when running locally before adding the server to the chain
+    // Normalize local-ish hostnames (loopback, LAN IPs) to 'localhost' so the
+    // client matches the QSS server's default QSS_HOSTNAME in the sigchain.
     let host = url.parse(this._qssEndpoint).hostname!
-    if (host === '127.0.0.1') {
+    if (
+      /^(127\.\d+\.\d+\.\d+|10\.\d+\.\d+\.\d+|192\.168\.\d+\.\d+|172\.(1[6-9]|2\d|3[01])\.\d+\.\d+|localhost)$/.test(
```
mcginty, line 456: Was this change caused by some issue? I don't know if I fully understand this hostname normalization.

packages/mobile/src/store/pushNotifications/pushNotifications.master.saga.ts
```diff
@@ -95,7 +96,13 @@ function* watchDeviceToken(): Generator {
...
     while (true) {
       const { token } = yield* take(channel)
-      logger.info('Received device token')
+      logger.info('Received device token, waiting for websocket connection')
+      while (true) {
+        const connected = yield* select(initSelectors.isWebsocketConnected)
+        if (connected) break
+        yield* delay(500)
+      }
+      logger.info('Sending device token to backend')
```
mcginty, line 105: Is the wait-for-websocket here to hold future device token events until a connection is made? Just wondering if this is the place to have this check, or if it should be later once the device token is trying to be "used".

mcginty (commented):
Mostly clarifying questions - sorry I let this fall through the cracks @adrastaea!

packages/backend/src/nest/qps/qps.service.ts
adrastaea: No, the teamId is encoded in the ucan facts field

packages/backend/src/nest/qps/qps.service.ts
adrastaea: Yes, and only when the client is the originator.

packages/backend/src/nest/qss/qss.service.ts
adrastaea: This was necessary to support testing on iOS with a dockerized local QSS instance that I exposed on LAN

packages/mobile/src/store/pushNotifications/pushNotifications.master.saga.ts
adrastaea: This is for the websocket with the local express server that the backend sets up. This wait loop prevents us from dropping device tokens that we get from the native service before the backend has set up.

packages/backend/src/nest/qps/qps.service.ts
adrastaea: Oh, and that change to the ucan format is in https://github.com/TryQuiet/quiet-storage-service/pull/23

packages/backend/src/nest/qss/qss.service.ts
mcginty: This is a nitpick, but is it a thing we can put behind a non-production environment variable or something just to ensure this workaround isn't run in production?

adrastaea: It is sort of already behind an environment variable because it's interpreting QSS_ENDPOINT which in production would be set to our production endpoint. I just pushed a commit that reinforced that with a NODE_ENV guard now though

mcginty: Nice! I think the explicitness helps here just for auditing/readability as well to reassure someone unfamiliar that this only happens in testing.

mcginty (approved):
Sweet lgtm

PR #3119, opened 2026-03-03
Electron32

bitmold:
Keeping this as a draft until #3118 and the electron 31 PRs have been integrated. 

Also just making the PR now to verify CI testing is as it should be.....

Bumps from electron 31 to 32,


bitmold:
@mcginty @holmesworcester @islathehut 

Now that #3138 has been merged, this electron 32 PR is ready for review. 

For some reason, on `develop` after bootstrapping the `package-lock.json` for Desktop is changed with tons of deletions. Perhaps this was some mistake around #3138? So the first commit is unrelated, it just gets the repo to a clean slate. Then the second commit is the small changes needed for electrion 32, and updating the web driver for e2etests....

Note: that the CI is complaining again about a dependency. **The only** complaint it has is with electron 32, it's the same issue it gave when we upgraded electron last time. It'll stop giving an error when we hit electron 35, so this PR brings us one step closer. Once this is merged, the CI won't complain about this (it's only when deps change that it raises an error)

bitmold:
Bah, somehow the e2e tests are breaking now... Fixing.,....

bitmold:
@islathehut @mcginty this is finally in a good state with everything passing except the dependency analyzer (and that's just cuz we haven't made it to v35 of electron yet)

islathehut (approved):

PR #3121, opened 2026-03-04
Update to tor 0.4.9.5 on Android, Add script for future android tor builds 

bitmold:


packages/mobile/.gitignore
```diff
@@ -83,3 +83,5 @@ buck-out/
 # Detox
 /e2e/artifacts
 /e2e/visual-regressions/local
+
+quiet_libtor_android
```
mcginty, line 87: Would it make sense to put this folder within `packages/mobile/android/build` which is already an ignored Android artifact directory?

packages/mobile/update-android-libtor.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
```
mcginty:
```suggestion
set -e
cd "$(dirname "$0")"
```

packages/mobile/update-android-libtor.sh
```diff
@@ -0,0 +1,43 @@
...
+# set to only build arm64-v8a matching quiet...
+./tor-droid-make.sh build
+
+ls -al external/lib/arm64-v8a
+cp -fv external/lib/arm64-v8a/libtor.so ../android/app/src/main/jniLibs/arm64-v8a/libtor.so
+cd ..
+git status
+cd android
+./gradlew clean assembleDebug
```
mcginty:
```suggestion
```

Seems a bit out of scope to build the whole Android app at the bottom of this script, but I understand having it to confirm things build.

bitmold:
This actually isnt ready for review yet but no it wouldn't make sense to put this there.

That directory would be wiped and rewiped through rebuilding the android app in gradle / android studio - which would be very unwelcomed when you're troubleshooting a new tor build.

I can find a diff spot for it, but most people wouldn't even have this on their machine its just an artifact for those who decide to locally build a new tor arm64 bin for android.

PR #3129, opened 2026-03-13
Closes #3123 Removes Superflous @peculiar/webcrypto dependency

### LINKED ISSUE #3123: Figure out @peculiar/webcrypto  dependency
Is there any reason to continue using @peculiar/webcrypto instead of the default JavaScript webcrypto library we have with nodejs 20.

If there is a reason, document it. If there isn't a reason, remove @peculiar/webcrypto from the project and replace it with the standard library.

bitmold:
Closes #3123 Removes Superflous @peculiar/webcrypto dependency. Removes hacky JavaScript that was needed to override global.crypto

    - this was probably needed for along time because iOS was using nodejs12 until some recent work, hence
    having no global.crypto flag.
    - iOS and Android need the --experimental-global-webcrypto flag until nodejs mobile is rebuilt

### DIFF
--- a/packages/mobile/android/app/src/main/java/com/quietmobile/Backend/BackendWorker.kt
+++ b/packages/mobile/android/app/src/main/java/com/quietmobile/Backend/BackendWorker.kt
@@ -228,6 +228,7 @@ class BackendWorker(private val context: Context, workerParams: WorkerParameters
 
         val command: MutableList<String> = ArrayList()
         command.add("node")
+        command.add("--experimental-global-webcrypto")
         command.add("--experimental-global-customevent")
         command.add(scriptPath)
         command.addAll(args)
--- a/packages/backend/jestSetup.ts
+++ b/packages/backend/jestSetup.ts
@@ -1,14 +1,10 @@
 import { setEngine, CryptoEngine } from 'pkijs'
-import { Crypto } from '@peculiar/webcrypto'
-
-const crypto = new Crypto()
-global.crypto = crypto
 
 setEngine(
   'newEngine',
   new CryptoEngine({
     name: 'newEngine',
     // @ts-ignore
-    crypto: crypto,
+    crypto: global.crypto,
   })
 )
--- a/packages/backend/package-lock.json
+++ b/packages/backend/package-lock.json
@@ -43,7 +43,6 @@
         "@nestjs/platform-express": "^10.2.10",
         "@orbitdb/core": "file:../../3rd-party/orbitdb",
         "@paralleldrive/cuid2": "^2.2.2",
-        "@peculiar/webcrypto": "1.5.0",
         "abortable-iterator": "^3.0.0",
         "async-mutex": "^0.5.0",
         "base-x": "^5.0.0",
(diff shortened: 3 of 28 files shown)


packages/mobile/android/app/src/main/java/com/quietmobile/Backend/BackendWorker.kt
```diff
@@ -228,6 +228,7 @@ class BackendWorker(private val context: Context, workerParams: WorkerParameters
 
         val command: MutableList<String> = ArrayList()
         command.add("node")
+        command.add("--experimental-global-webcrypto")
```
mcginty, line 231: Out of curiosity, did it fail without this flag? Looking at https://nodejs.org/api/webcrypto.html, I see that since v20.19.3 Ed25519 and X25519 are stable, but I wouldn't be surprised if there was still other experimental algorithms this was needed for.

bitmold: Yes, it failed without this flag.

As far as I understand, this flag, which used to be present on iOS/Android and Desktop, was doing *nothing* due to Quiet's usage of `@peculiar/webcrypto`.I'm just adding it back here.

There are four nodejs instances to keep track of with Quiet:

- the nodejs you run when you build the project (ie the version set by `nvm use`) 
- the nodejs version that the Desktop app uses at runtime, which is packaged with `electron` and changes based on which version of `electron` you are using 
- android `nodejs-mobile` that's used for the `backend` 
- and the same thing for iOS  

(Quiet packages these two nodejsmobile binaries on iOS and Android because react native apps on their own don't actually ship with nodejs at all, very much unlike how electron does on Desktop) 

My previous PR that was merged upgraded the first two nodejs's in that list from 18 to 20. However, on iOS and Android we are still using `nodejsmobile` that's set for version 18 - so there is webcrypto support, but needs it enabled with that flag. 

So, in a sense it was a mistake for me to have removed the flag on iOS and Android, but in another sense it never mattered since the flag wasn't doing anything in the first place! Since we never used anything enabled by the flag on mobile (since nothing broke) - I just assumed things were good. Another funny realization is that this flag was never needed on Desktop, since we only actually started to use the proper webcrypto API after upgrading to nodejs 20, where webcrypto's enabled by default. 

I'm happy to make iOS and Android builds of nodejsmobile based on node20, but until we do, those two apps need the flag, which is now actually used, in order to function. [I recently compiled a nodejs 18 build for iOS since it was still using node 16.](https://github.com/TryQuiet/quiet/commit/a1c035ca81634f0eec0f938d3afcfa39d78ab901#diff-852023f2bd5d53aa914f2ec0dabe6ef5ac7ba5a3f0c14d70c80c1ea1873e23e7R26). I also recently recompiled nodejs mobile for android (not changing the node version) since the previous compiled binary wasn't 16KB Aligned (a new restriction on binaries Google was enforcing). IMO, these should be two seperate PRs though.

adrastaea (approved):

mcginty (approved):
Given the clarifying comment, approving and we can address the node version disparity later.

PR #3130, opened 2026-03-13
fix soon to be broken github actions that need the node24 runner

bitmold:
Recently if you start using the GitHub CI you see the following warning message after completing a workflow:

> Warning: Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/cache@v3, actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11, actions/setup-python@v5, dorny/paths-filter@v3. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

I think this will break workflows in early June. The solution seems to be updating to the node24 releases of the github actions we use.

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -35,7 +35,7 @@ jobs:
       TEST_MODE: true
       IS_LOCAL: true
     steps:
-      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+      - uses: actions/checkout@v5
         with:
           submodules: recursive
 
@@ -99,7 +99,7 @@ jobs:
       startsWith(github.ref, 'refs/tags/@quiet/desktop')
 
     steps:
-      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: actions/checkout@v5 # v4.1.1
         with:
           submodules: 'recursive'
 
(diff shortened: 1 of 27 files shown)


.github/workflows/desktop-build.yml
```diff
@@ -227,7 +227,7 @@ jobs:
       S3_BUCKET: ${{ github.event.action == 'released' && 'quiet.7.x' || 'test.quiet' }}
 
     steps:
-      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: actions/checkout@v5 # v4.1.1
```
mcginty: lol the comment

mcginty (approved):

.github/workflows/desktop-build.yml
bitmold: Ah :upside_down_face:  whoops, I removed it. Not sure how I missed that in the diff.

PR #3125, opened 2026-03-10
iOS Push Notification Support

mcginty:
* Upgrades the iOS project to build in Xcode 26.3 and work on newer iOS emulators.
* Adds Firebase libraries as Pods
* Adds a Network Service Extension with the needed glue to process APN push notifications via Firebase

PR #3144, opened 2026-03-24
iOS CI Build Fixes

mcginty:
This is a PR of the necessary CI-related fixes for building iOS with notification support.

- Add the new `IOS_FIREBASE_KEY` secret env var
- Change the Xcode compatibility version from 12 to 16 (needed due to a CocoaPods/Xcodeproj issue)
- Remove unused entitlements that required Apple Dev Portal changes
- Add new provisioning profile for the NSE and glue it into the Xcode project


packages/mobile/ios/Quiet/Quiet.entitlements
```diff
@@ -4,7 +4,5 @@
 <dict>
 	<key>aps-environment</key>
 	<string>development</string>
-	<key>com.apple.security.application-groups</key>
```
islathehut, line 7: Does this have any side effects in the app?

packages/mobile/ios/Quiet.xcodeproj/project.pbxproj
```diff
@@ -5810,7 +5822,8 @@
 				CLANG_WARN_DOCUMENTATION_COMMENTS = YES;
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CODE_SIGN_ENTITLEMENTS = QuietNotificationServiceExtension/QuietNotificationServiceExtension.entitlements;
-				CODE_SIGN_STYLE = Automatic;
+				CODE_SIGN_IDENTITY = "Apple Distribution: A Quiet LLC (CTYKSWN9T4)";
+				CODE_SIGN_STYLE = Manual;
```
adrastaea, line 5826: Is this really necessary? We're going to have to change this back to automatic in Xcode during development each time if we do this.

packages/mobile/ios/Quiet/Quiet.entitlements
mcginty: With @adrastaea's change to use the iOS Keychain across the main Quiet app and the NSE, this will be necessary again as far as I understand, but we would need to specify an app group that doesn't currently exist, it's just empty here.

mcginty: I added it unnecessarily in my prior iOS notification PR: https://github.com/TryQuiet/quiet/pull/3125/changes#diff-4c6f53c1ca8f912391f9a35df3fec6c26a075dda679cb0026fc8c4078d918eefR7

packages/mobile/ios/Quiet.xcodeproj/project.pbxproj
mcginty: Note that this only affects the Release build, so should be unrelated to development! It seemed necessary to assign the correct code signing profile when I was trying to get it up and running in CI. I was running into signature issues.

packages/mobile/ios/Quiet/Quiet.entitlements
islathehut: Ahh gotcha

adrastaea (approved):

PR #3177, opened 2026-04-03
feat(3155): Add private channels with modifiable membership (no removals) to desktop

### LINKED ISSUE #3155: Users should be able to create a private channel
This is a placeholder issue for an initial sprint to "users can create and use private channels".

Figma: https://www.figma.com/design/PVQ1Kjf6Cq8ng1czuVtvR8/Create-channels-for-specific-roles?node-id=838-9409&t=gb61QyjnTR1QjJUG-0

The key is the switch/slider on the channel creation step to create a private channel. 

Image

The feature was imagined for either adding users or roles. It sounds like skipping roles for now and just focusing on users is the most ergonomic, so we can do that. Use this design and omit the "Roles" area.

Image

Note the use of a new sidebar area on desktop to help unify desktop and mobile designs. 

Acceptance criteria:

1. Users can create private channels and invite others
2. Channels correspond to encrypted groups in local-first auth and are secure. Non-invited members cannot commpel access
3. Channels are only visible to (and channel names are only decryptable by) invited members.

Nice to have:
1. Users can add more people to private channels
2. Feature matches design

islathehut:
### Pull Request Checklist


packages/backend/src/nest/auth/sigchain.service.ts
```diff
@@ -132,29 +135,20 @@ export class SigChainService extends EventEmitter {
...
-      .map(user => ({
-        userId: user.userId,
-        roles: user.roles,
-        isRegistered: true,
-        isDuplicated: false,
-      })) as User[]
-    this.socketService.emit(SocketEvents.USERS_UPDATED, { users })
-    this.emit('updated')
+  private handleChainUpdate = async () => {
+    this.emit(SigchainEvents.UPDATED)
```
islathehut: I had to move the `USERS_UPDATED` logic out of the sigchain service and into the connection manager because I had to pull channel data to map role names to channel ids/names.  To obfuscate membership and hide private channel names from non-members (and QSS) the role names are deterministically generated from channel IDs but aren't decodable without already knowing the channel ID itself.  To determine channel membership we have to pull the private channels from the channel metadata DB and match the two.  Since storage has the sigchain as a dependency I couldn't just pull it in as a dependency here so the logic had to be moved.

holmesworcester:
@islathehut I found a few superficial design issues on desktop and I think fixed them in this branch (off of your PR branch): https://github.com/TryQuiet/quiet/tree/fix/private-channel-icons

There might be similar issues on mobile. Can you address them on mobile while you're finishing that?

packages/backend/src/nest/auth/sigchain.service.ts
```diff
@@ -132,29 +135,20 @@ export class SigChainService extends EventEmitter {
...
   private attachSocketListeners(chain: SigChain): void {
     this.logger.info('Attaching socket listeners')
-    chain.on('updated', this.handleChainUpdate)
+    chain.on(SigchainEvents.UPDATED, () => this.handleChainUpdate(chain.team!.id, chain.team!.teamName))
   }
 
   private detachSocketListeners(chain: SigChain): void {
     this.logger.info('Detaching socket listeners')
-    chain.removeListener('updated', this.handleChainUpdate)
+    chain.removeListener(SigchainEvents.UPDATED, () => this.handleChainUpdate(chain.team!.id, chain.team!.teamName))
```
adrastaea: I don't think this will properly remove the listener because of the lambda

packages/backend/src/nest/storage/channels/messages/orbitdb/PrivateMessagesAccessController.ts
```diff
@@ -0,0 +1,51 @@
...
+      if (!crypto) throw new NoCryptoEngineError()
+
+      const writerIdentity = await identities.getIdentity(entry.identity)
+      if (!writerIdentity) {
+        return false
+      }
+
+      const { id } = writerIdentity
+      if (config.write.includes(id) || config.write.includes('*')) {
+        if (!identities.verifyIdentity(writerIdentity)) {
```
adrastaea: This is a promise, but it is not awaited. With this, we are evaluating the truthiness of the promise, so it is always true.

packages/backend/src/nest/storage/channels/messages/orbitdb/MessagesAccessController.ts
```diff
@@ -75,7 +26,7 @@ export const MessagesAccessController =
       }
 
       const { id } = writerIdentity
-      if (write.includes(id) || write.includes('*')) {
+      if (config.write.includes(id) || config.write.includes('*')) {
         if (!identities.verifyIdentity(writerIdentity)) {
```
adrastaea: Also, not awaited

packages/backend/src/nest/storage/channels/messages/orbitdb/PrivateMessagesAccessController.ts
```diff
@@ -0,0 +1,51 @@
...
+          return false
+        }
+      } else {
+        return false
+      }
+
+      const sigchain = config.sigchainService.getChain({ teamId: entry.payload.value!.teamId })
+      if (!sigchain.channels.memberInChannel(id, entry.payload.value!.channelId)) {
```
adrastaea: This trusts the entry's channel id parameter as necessarily true, so a member of channelA could write an entry to channelB that they are not a member of. We should pass the channelId in when insantiating the access controller so that we can ensure that both the entry is written to the correct channel, and that the user is a member of that channel.

packages/backend/src/nest/storage/channels/messages/orbitdb/PrivateMessagesAccessController.ts
```diff
@@ -0,0 +1,51 @@
...
+      const { id } = writerIdentity
+      if (config.write.includes(id) || config.write.includes('*')) {
+        if (!identities.verifyIdentity(writerIdentity)) {
+          return false
+        }
+      } else {
+        return false
+      }
+
+      const sigchain = config.sigchainService.getChain({ teamId: entry.payload.value!.teamId })
```
adrastaea: This too trusts that the channel is a channel in the team. This is somewhat in the future with multiteam apps, but we should pass in the expected teamId for the channel or add a todo

packages/backend/src/nest/storage/channels/messages/orbitdb/BaseMessageAccessController.ts
```diff
@@ -0,0 +1,114 @@
...
+        address = posixJoin('/', this.type, address)
+      }
+
+      const crypto = getCrypto()
+
+      return {
+        type: this.type,
+        address,
+        write,
+        canAppend: this.canAppend(config, identities) as any,
```
adrastaea:
```suggestion
        canAppend: this.canAppend({ ...config, write }, identities) as any,
```
Without this change, the peristed write value from the manifest is ignored in favor of the value from config passed in from the constructor. Might bite us in the dm case where we initialize an orbitdb with `write: [userA, userB]`. I think this is not an issue in the current implementation where every store gets `write: ['*']`, but probably worth fixing.

packages/backend/src/nest/storage/channels/channel.store.ts
```diff
@@ -272,7 +292,7 @@ export class ChannelStore extends EventStoreBase<EncryptedMessage, ConsumedChann
 
       if (ids == null || ids?.includes(x.value.id)) {
         const decryptedMessage = await this.messagesService.onConsume(x.value)
```
adrastaea, line 331: Should call privateMessagesService if channel is private? Maybe we should choose to alias a messagesService variable to the appropriate service at initialization instead of relying on an if else everywhere we want to call one of these methods.

packages/backend/src/nest/storage/channels/messages/orbitdb/PrivateMessagesAccessController.ts
adrastaea: We should probably get rid of these bangs and use a proper validation to avoid errors being thrown and just return false if it's malformed

packages/backend/src/nest/storage/channels/messages/private-channel-messages.service.ts
```diff
@@ -0,0 +1,131 @@
...
+  public validateMessage(message: ConsumedChannelMessage, encryptedMessage: EncryptedMessage): boolean {
+    if (message.id !== encryptedMessage.id) {
+      this.logger.warn(`Cannot validate msg ${message.id}: IDs do not match`)
+      return false
+    }
+    if (!isConsumedChannelMessage(message)) {
+      this.logger.warn(`Cannot validate msg ${message.id}: message shape is not valid`)
+      return false
+    }
+    if (!isEncryptedMessage(encryptedMessage)) {
```
adrastaea: I made another comment on the access controller about doing entry shape validation instead of using bangs, and I think moving this into the access controller would accomplish that.

packages/backend/src/nest/storage/channels/messages/orbitdb/PrivateMessagesAccessController.ts
adrastaea: It's also worth considering the case where we get a log synced to us before we know that the user is a member of the channel. We might want to consider stopping sync while negotiating syncs with the auth connection or buffering entries that fail for some time if they fail, and then trying again when we get sigchain updates. I would be fine putting that in as a note for future work though.

packages/backend/src/nest/storage/storage.module.ts
```diff
@@ -7,7 +7,7 @@ import { CommunityMetadataStore } from './communityMetadata/communityMetadata.st
 import { UserProfileStore } from './userProfile/userProfile.store'
 import { IpfsModule } from '../ipfs/ipfs.module'
 import { ChannelsService } from './channels/channels.service'
-import { MessagesService } from './channels/messages/messages.service'
+import { PublicChannelMessagesService } from './channels/messages/public-channel-messages.service'
```
adrastaea: unused imports

packages/backend/src/nest/validation/validators.ts
```diff
@@ -96,6 +96,9 @@ const channelSchema = joi.object({
   timestamp: joi.number().required(),
   id: joi.string(),
   address: joi.string(),
+  public: joi.boolean().required(),
```
adrastaea: This is going to make this not backward compatible. We should make this optional and if it's not present then assume it is public.

packages/backend/src/nest/storage/channels/channels.service.ts
```diff
@@ -127,9 +140,9 @@ export class ChannelsService extends EventEmitter {
    * subscribing to newly created channel stores.
    */
   public async createChannelsDb(): Promise<void> {
-    this.logger.info('Creating public-channels database')
+    this.logger.info('Creating channels database')
     this.channels = await this.orbitDbService.open<KeyValueIndexedValidatedType<EncryptedAndSignedPayload>>(
-      'public-channels',
+      'channels',
```
adrastaea: Changing this to channels will make this not backwards compatible. Should we put in a migration?

packages/backend/src/nest/storage/channels/messages/private-channel-messages.service.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Handle processing of message consumed from OrbitDB
+   *
+   * @param message Message consumed from OrbitDB
+   * @returns Processed message if decryptable, undefined if undecryptable and false if intentionally skip decryption
+   */
+  public async onConsume(message: EncryptedMessage): Promise<ConsumedChannelMessage | false | undefined> {
+    const chain = this.sigChainService.getChain({ teamId: message.teamId })
```
adrastaea: This is a good change from where `public-channel-messages.service.ts` still uses activeChain. This can also throw an unhandled error though, and we should wrap it.

packages/backend/src/nest/storage/channels/channels.service.ts
```diff
@@ -18,36 +18,46 @@ import {
...
-  // Map of message stores for each channel where the key is the channel ID
-  public publicChannelsRepos: Map<string, PublicChannelsRepo> = new Map()
+  // Map of message stores for each public channel where the key is the channel ID
+  public channelsRepos: Map<string, ChannelRepo> = new Map()
 
   // Channel metadata store
   public channels: KeyValueIndexedValidatedType<EncryptedAndSignedPayload> | undefined
 
+  // Is the service initialized
+  public initialized: boolean = false
```
adrastaea, line 76: We should use this to also prevent the file manager listeners from being applied twice if someone leaves a community and rejoins. Right now, calling `init` twice will duplicate those listeners. Or just clean up those listeners when closed

packages/backend/src/nest/storage/channels/channels.service.ts
```diff
@@ -56,7 +66,9 @@ export class ChannelsService extends EventEmitter {
     private readonly filesManager: IpfsFileManagerService,
     private readonly orbitDbService: OrbitDbService,
     private readonly moduleRef: ModuleRef,
-    private readonly sigchainService: SigChainService
+    private readonly sigchainService: SigChainService,
+    private readonly messagesAccessController: MessagesAccessController,
```
adrastaea: unused

packages/backend/src/nest/storage/channels/channels.service.ts
```diff
@@ -56,7 +66,9 @@ export class ChannelsService extends EventEmitter {
...
+    private readonly privateMessagesAccessController: PrivateMessagesAccessController
```
adrastaea: unused

packages/backend/src/nest/storage/channels/channels.service.ts
```diff
@@ -492,10 +575,45 @@ export class ChannelsService extends EventEmitter {
       store = await this.createChannelStore(channelData)
     }
     await store.deleteChannel()
-    this.publicChannelsRepos.delete(channelId)
+    this.channelsRepos.delete(channelId)
     return { channelId, deleted: true } as DeleteChannelResponse
   }
 
+  public async addMembersToPrivateChannel(payload: AddMembersChannelPayload): Promise<AddMembersChannelResponse> {
```
adrastaea, line 598: This should also check that the channel you are trying to add members to is marked as private

packages/desktop/src/rtl-tests/channel.add.test.tsx
```diff
@@ -176,6 +188,121 @@ describe('Add new channel', () => {
     // Check if sidebar item displays as selected
     const link = screen.getByTestId(`${channelName.output}-link`)
     expect(link).toHaveClass('ChannelsListItemselected')
+    const linkIcon = screen.getByTestId(`${channelName.output}-channel-link-icon-public`)
+    expect(linkIcon).toBeVisible()
+  })
+
+  it.only('Adds new private channel and opens it. Sends initial message', async () => {
```
adrastaea: Is this intended? Skips all other tests

packages/desktop/src/renderer/components/Channel/AddMembersChannel/AddMembersChannelComponent.tsx
```diff
@@ -0,0 +1,342 @@
...
+  open,
+  handleClose,
+  channelName,
+  channelId,
+  possibleMembers,
+  allUsers,
+  addMembersToChannel,
+}) => {
+  const theme = useTheme()
+  const [selectedMembers, setSelectedMembers] = useState<AutoCompleteOption[]>([])
```
adrastaea, line 161: selectedMembers isn't reset when the modal closes

packages/desktop/src/renderer/components/Channel/CreateChannel/CreateChannel.tsx
```diff
@@ -9,6 +9,7 @@ import { ModalName } from '../../../sagas/modals/modals.types'
 import { flushSync } from 'react-dom'
 import { generateChannelId } from '@quiet/common'
 import { createLogger } from '../../../logger'
+import { is } from 'ramda'
```
adrastaea: unused

adrastaea (changes requested):

packages/desktop/src/renderer/components/Channel/AddMembersChannel/AddMembersChannelComponent.tsx
adrastaea: Example of effect: open add member dialog, select a user, close modal, reopen and hit add. Previously selected member will be added without being shown in the selection box

packages/backend/src/nest/storage/channels/channel.store.ts
islathehut: Yea this is an issue I discovered while doing DMs.  I'll port over my change.

packages/desktop/src/rtl-tests/channel.add.test.tsx
islathehut: No that was accidentally left in

packages/backend/src/nest/storage/channels/channels.service.ts
islathehut: Reverted

packages/backend/src/nest/validation/validators.ts
islathehut: Made it optional

packages/backend/src/nest/storage/channels/messages/private-channel-messages.service.ts
islathehut: Added handling for missing chains and updated the public message service to use the same logic.

packages/desktop/src/renderer/components/Channel/AddMembersChannel/AddMembersChannelComponent.tsx
islathehut: Now clearing when I clear the options array.

packages/backend/src/nest/storage/channels/messages/orbitdb/PrivateMessagesAccessController.ts
islathehut: Updated to validate the channel ID matches and uses the channel ID from the metadata record.

I think your second point is a weak point for all of our orbit DB reads.  If you get an OrbitDB record before you have a user in your chain the OrbitDB identity check will fail since that validates against the chain.

packages/backend/src/nest/storage/channels/messages/private-channel-messages.service.ts
islathehut: Moved it.

packages/backend/src/nest/storage/channels/channels.service.ts
islathehut: I added cleanup for those listeners.  This did raise a separate issue of the state manager creating duplicates of event handlers.  We rerun the event handler attachment each time so if you leave, join, leave, join, etc. you'll have duplicate handlers for each join.  I think fixing that is out of scope for this but something we should fix.

packages/backend/src/nest/storage/channels/channels.service.ts
islathehut: Added

packages/backend/src/nest/auth/sigchain.service.ts
islathehut: Made a slight change and checking the logs I didn't see any duplicates.

packages/backend/src/nest/storage/channels/messages/orbitdb/PrivateMessagesAccessController.ts
islathehut: Added TODOs for both

packages/backend/src/nest/storage/channels/channels.service.ts
islathehut: Added TODOs

packages/backend/src/nest/storage/channels/messages/orbitdb/BaseMessageAccessController.ts
```diff
@@ -0,0 +1,121 @@
...
+      } else {
+        address = await AccessControlList({ storage, type: this.type, params: { write } })
+        address = posixJoin('/', this.type, address)
+      }
+
+      return {
+        type: this.type,
+        address,
+        write,
+        canAppend: this.canAppend({ ...config, ...write }, identities) as any,
```
adrastaea:
```suggestion
        canAppend: this.canAppend({ ...config, write }, identities) as any,
```
`write` is an array of strings. We want to overwrite `config.write`, not spread the array itself

packages/backend/src/nest/storage/channels/messages/orbitdb/BaseMessageAccessController.ts
```diff
@@ -0,0 +1,121 @@
...
+  LRUStorage,
+  IPFSBlockStorage,
+  AccessControllerType,
+  AccessController,
+  CanAppendFunc,
+} from '@orbitdb/core'
+import { getCrypto } from 'pkijs'
+import { NoCryptoEngineError, NotImplementedError } from '@quiet/types'
+import { posixJoin } from '../../../orbitDb/util'
+import { EncryptedMessage } from '../messages.types'
```
adrastaea: unused imports

packages/backend/src/nest/storage/channels/messages/orbitdb/BaseMessageAccessController.ts
```diff
@@ -0,0 +1,121 @@
...
+import { createLogger } from '../../../../common/logger'
+import { QuietLogger } from '@quiet/logger'
+import { SigChainService } from 'packages/backend/src/nest/auth/sigchain.service'
```
adrastaea:
```suggestion
import { SigChainService } from '../../../../auth/sigchain.service`
```

islathehut: Ugh I borked my vscode config at one point trying to resolve a different import issue and thought I caught all of the places that had these bad import statements

packages/backend/src/nest/storage/channels/messages/orbitdb/BaseMessageAccessController.ts
islathehut: Ahh that's what I get for fixing lots of things rapid fire.  Appreciate putting up a fix

packages/backend/src/nest/qss/qss-sync-manager.service.ts
```diff
@@ -752,6 +752,7 @@ export class QSSSyncManager implements OnModuleDestroy, OnModuleInit {
         teamId,
         userId,
         ...(nextStartSeq != null ? { startSeq: nextStartSeq } : { startSeq: 0 }),
+        startTs: 0,
```
islathehut, line 755: @adrastaea I don't if something went wrong when merging from develop but qss syncing was failing due to an error with this field being missing from the request payload.

adrastaea (approved):

PR #3184, opened 2026-04-13
fix: sidebar layout tweaks

holmesworcester:
## Summary

Small cluster of sidebar layout fixes, found while comparing the self profile tab against the reference design:

- **Vertically center the username next to the avatar** in the self profile tab at the bottom of the sidebar. Two causes: the flex row was missing `alignItems: center`, and `ProfilePhoto` has a default `marginBottom: 16px` inline style that was forcing the avatar up relative to the text. Fixed both.
- **Dim the username text in the Users list** (`opacity: 0.7`, `fontWeight: 300`) to match the channels-list styling. Previously it rendered at full white.
- **Remove the decorative `borderTop` above the self profile tab.** In prod it was floating orphaned above TorStatus's hidden-but-still-space-taking row, and the design doesn't actually want it in either mode.
- **Return null from `TorStatus` in non-development builds** so the dev-only indicator doesn't occupy ~56px of empty space at the bottom of the sidebar in prod. Previously it was hidden via `opacity: 0%` but still took layout space.
- **Force `overflow-x: hidden` on the sidebar's `rc-scrollbars` view.** On Linux Chromium the view had `overflow: scroll` on both axes, and at certain viewport heights the reserved horizontal-scrollbar gutter peeked out as a faint 1-pixel stripe above TorStatus (with the panels below rendering 1px wider than the scroll view).

## Test plan

- [ ] Self profile tab: avatar and username are vertically centered with each other
- [ ] Users list: nicknames render dimmed, matching the channels list
- [ ] In a prod build, no divider line and no empty 56px gap at the bottom of the sidebar
- [ ] In a dev build, TorStatus still renders (TOR label + indicator) above the self profile
- [ ] At various window heights on Linux, no thin horizontal stripe appears above TorStatus; sidebar width is consistent from top to bottom


adrastaea (approved):
please update snapshots and add to the changelog before merging, but looks good!

holmesworcester:
@copilot resolve the merge conflicts in this pull request

PR #3188, opened 2026-04-16
Android Push Notifications

### LINKED ISSUE #3156: User should be able to receive push notification on Android without background use
When QSS is enabled, users should get Android notifications via QSS by default and have to enable background Tor notifications manually.

This issue is for a spike to get Android notifications working in a full path in as similar a way to iOS notifications as makes sense, looking for any scope simplifications that are possible on Android.

Acceptance criteria:

- User gets full message notifications even when background usage is disabled
- Background usage is disabled by default when QSS is enabled (we can add a setting for it in the future)

adrastaea:
### Pull Request Checklist


.github/secrets/decrypt_secrets.sh
```diff
@@ -6,12 +6,14 @@ gpg --quiet --batch --yes --decrypt --passphrase="$IOS_PROFILE_KEY" --output ./.
 gpg --quiet --batch --yes --decrypt --passphrase="$IOS_NSE_PROFILE_KEY" --output ./.github/secrets/match_AppStore_comquietmobile_QuietNotificationServiceExtension.mobileprovision ./.github/secrets/match_AppStore_comquietmobile_QuietNotificationServiceExtension.mobileprovision.gpg
 gpg --quiet --batch --yes --decrypt --passphrase="$IOS_CERTIFICATE_KEY" --output ./.github/secrets/Certificates.p12 ./.github/secrets/Certificates.p12.gpg
 gpg --quiet --batch --yes --decrypt --passphrase="$IOS_FIREBASE_KEY" --output ./.github/secrets/GoogleService-Info.plist ./.github/secrets/GoogleService-Info.plist.gpg
+gpg --quiet --batch --yes --decrypt --passphrase="$IOS_FIREBASE_KEY" --output ./.github/secrets/google-services.json ./.github/secrets/google-services.json.gpg
```
holmesworcester: should we rename this key if we're using it for both?

packages/backend/src/nest/auth/sigchain.service.spec.ts
```diff
@@ -177,6 +177,43 @@ describe('SigChainService - listener lifecycle', () => {
...
+      Object.defineProperty(process, 'platform', { value: originalPlatform })
+      if (originalQpsAllowed == null) {
+        delete process.env.QPS_ALLOWED
+      } else {
+        process.env.QPS_ALLOWED = originalQpsAllowed
+      }
+    }
+  })
+
   it('emits device credentials for the NSE on ios', async () => {
```
holmesworcester, line 217: On Android, can't we just run full LFA and use whatever credentials we have, even new credentials that we get in the process of connecting to QSS and syncing LFA?

Another way of putting this is: on iOS when validating/decrypting notifications we only have the keys we knew about at the time of the last foreground run, right?

On Android, is this also going to be the case, given the current implementation style? 

Did we do it that way just because it's cheaper to be similar to the iOS implementation? Is that actually the case? Does it have to be? Is this best addressed in further work or is a discussion helpful?

packages/mobile/android/app/src/main/java/com/quietmobile/Push/Base58.kt
```diff
@@ -0,0 +1,63 @@
+package com.quietmobile.Push
```
holmesworcester: Should this be a library or is that not worth it?

packages/mobile/android/app/src/main/java/com/quietmobile/Push/MsgpackDecoder.kt
```diff
@@ -0,0 +1,377 @@
+package com.quietmobile.Push
+
+import java.io.ByteArrayOutputStream
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
+import java.nio.charset.StandardCharsets
+
+object MsgpackUndefined
+
+object MsgpackDecoder {
```
holmesworcester, line 14: should this be a library?

packages/mobile/android/app/src/main/java/com/quietmobile/Push/QssAuthService.kt
```diff
@@ -0,0 +1,55 @@
+package com.quietmobile.Push
```
holmesworcester, line 1: Related to my question above, should we be using LFA for this on Android if we can? Can we?

packages/mobile/src/components/ContextMenu/menus/CommunityContextMenu.container.tsx
```diff
@@ -45,9 +52,30 @@ export const CommunityContextMenu: FC = () => {
...
+    }
+  }, [backgroundTorEnabled, dispatch])
+
   const items: ContextMenuItemProps[] = [
     { title: 'Create channel', action: () => redirect(ScreenNames.CreateChannelScreen) },
     { title: 'Add members', action: () => invitationContextMenu.handleOpen() },
+    ...(Platform.OS === 'android' && Config.QSS_ALLOWED === 'true' && community?.qssEnabled === true
+      ? [
+          {
+            title: backgroundTorEnabled ? 'Disable background Tor' : 'Enable background Tor',
```
holmesworcester, line 72: I made a separate issue for this: https://github.com/TryQuiet/quiet/issues/3230

holmesworcester (commented):
Most of my questions flow from the first question of, how much easier was it to get this working on Android following the iOS way of doing things vs. following the existing foreground cross-platform way of doing things (LFA, QSS, etc.)?

Do we have a sense of the cost of doing it the other way, in case we decide to in the future?

packages/backend/src/nest/auth/sigchain.service.spec.ts
adrastaea: I did it this way for a couple reasons:
1. Part of the ticket was to receive notifications when background usage was disabled, so I wanted to have as native a notification pathway as possible if we had totally shut down the node backend.
2. It's simpler to keep the implementation as similar as possible to ios so that its easier to reason about.
3. If we want to keep this up to date with LFA in the background, we can just open or leave open a QSS connection in the backend, receive new keys, and then do the native process. This pattern allows more flexibility in that regard.

.github/secrets/decrypt_secrets.sh
adrastaea: Ah yeah, I think I used that as a placeholder. I'll update to it's own key before merging

packages/mobile/android/app/src/main/java/com/quietmobile/Push/MsgpackDecoder.kt
adrastaea: There's some compatibility issues with msgpack because we use `moreTypes` in the qss serializer to support js maps which is not supported in the . I replaced the Base58 implementation with a library, and I changed this object to use the msgpack core where possible with an extension to support the js bonus types

packages/mobile/android/app/src/main/java/com/quietmobile/Push/Base58.kt
adrastaea: Replaced with a library

holmesworcester (approved):

PR #3195, opened 2026-04-26
Update GitHub actions

bitmold:
Recently Holmes shared this screenshot from a GitHub actions run:
image

In it, it shows that 3 actions have still not been updated to do their CI things in node 24 and that GitHub will soon force actions to run with nodejs24. The actions are:

- `Saionaro/extract-package-version` this can be trivialy updated to their new 1.4.0 release which explicitly uses node 24
- `TryQuiet/base64-to-file` This is [Quiet's unmodified fork ](https://github.com/tryquiet/base64-to-file) of https://github.com/timheuer/base64-to-file. Quiet's fork had no new commits, and there have been other fixes by the original maintainer. [In the near future, there will be a release supporting node24 that Quiet can update to.](https://github.com/timheuer/base64-to-file/issues/60). [Quiet's fork of this repo should be closed](https://github.com/tryquiet/base64-to-file)
- https://github.com/actions-ecosystem/action-slack-notifier the slack action has not seen any activity in years. I could fork it to run under node 24, however it seems pretty unlikely that if it were forced to run in node24 anything will break. We can test this.

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -51,7 +51,7 @@ jobs:
 
       - name: Extract version
         id: extract_version
-        uses: Saionaro/extract-package-version@fdb5b74adc1278ddb777dfed4c988b9d098bb48d # v1.2.1
+        uses: Saionaro/extract-package-version@v1.4.0
         with:
           path: packages/desktop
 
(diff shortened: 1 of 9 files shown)


bitmold:
@holmesworcester if it's necessary to update the slack action, all that probably needs to happen is to fork https://github.com/actions-ecosystem/action-slack-notifier and update it's [`action.yml`](https://github.com/actions-ecosystem/action-slack-notifier/blob/main/action.yml) file:

```yml
runs:
  using: node12
  main: dist/index.js
```

That action has been running with node 12, but will most likely behave the same in node 24 (it's been running fine on GitHub with node 20 already...). But again, I don't think making this change is going to be necessary....

Issue #3196, opened 2026-04-30
Mobile user should be able to give us logs


holmesworcester:
Right now it is difficult to get log files off mobile devices.

Success Criteria:

* Mobile users can send a log file to Quiet devs, either by exporting the file to a readable directory or by sending it as an email attachment. 
* This feature is unobtrusive for normal users
* It is limited to dev/alpha builds  
* We can use it on the joining screen
* We can send a zipped data_dir too, to the extent possible

holmesworcester:
I'm going to knock this back and add a send logs link on the join screen.

holmesworcester:
https://github.com/TryQuiet/quiet/pull/3213 is waiting for review. 

Sharing zipped data directory for debugging purposes in alphas is working on iOS and Android

kingalg:
Version: 7.1.0-alpha-32
Done.

PR #3198, opened 2026-04-30
feat(mobile): dev/alpha-only "Share logs" menu item

holmesworcester:
## Summary

- Adds a **Share logs** action to the community context menu on mobile, gated to non-production builds (`Config.NODE_ENV !== 'production'`). Visible in `development` and `staging`/alpha bundles, hidden in production.
- Reads the on-device log directory (`RNFS.DocumentDirectoryPath/logs`), bundles all log files plus a metadata header (app version, platform, device, sensitivity warning) into a single `quiet-logs-<timestamp>.txt` under `RNFS.CachesDirectoryPath/quiet-logs-share/`, and shares via the native share sheet with the To: field pre-filled with `logs@tryquiet.org`.
- Storybook stories added: `ContextMenu → Community (dev/alpha — Share logs)` and `SendLogs → Trigger share sheet` for on-device testing.
- Jest test added (`CommunityContextMenu.test.tsx`) verifying the gate hides the item in production and shows it in dev/staging.

Refs https://github.com/TryQuiet/quiet/issues/3196.

## Why this is dev/alpha-only

A pre-implementation audit of every `logger.*` and `console.*` call across `packages/{mobile,desktop,backend,common,state-manager,logger,identity,types}` found that current Quiet logs leak (in info-level entries, unconditionally written to the file every user would share):

- The user's **Tor onion service private key** and **libp2p peer private key** (full `Identity` object logged on every `sendMessage`).
- **Plaintext message content** (outgoing and incoming, plus mobile push-notification payloads).
- **Community PSK, root CA private key, and invitation seeds** (full `Community` and `InvitationData` payloads logged in `state-manager` and `local-db`).
- **Onion-address ↔ peerId** correlation (the deanonymization tuple in one line).

Sharing a current Quiet log would be equivalent to handing over the user's full identity, their community's secrets, and a sample of their plaintext chats. Until a redactor lands in `@quiet/logger` (`findAllByKeyAndReplace` infrastructure exists; needs wiring + a denylist) and the high-leverage call sites are scrubbed, this feature **must not** be exposed to public/production builds.

## Why a single-file bundle

First implementation used `Share.open({ urls: [...] })` to attach the log files individually. Gmail returned **"couldn't attach file"** because:

1. Logs live in `DocumentDirectoryPath` (app-private internal storage on Android), which is not exposed by `react-native-share`'s `FileProvider` (`share_download_paths.xml` only maps `cache-path` and a few external paths).
2. Even after staging into `CachesDirectoryPath`, the combination of `ACTION_SEND_MULTIPLE` + `EXTRA_EMAIL` + multiple `text/plain` URIs is fragile in Gmail — the URI grants don't always transfer through the email composer's process boundary.

Concatenating into a single `text/plain` file and using `ACTION_SEND` (single-attachment path) avoids both issues.

## Test plan

- [x] Jest unit test (`CommunityContextMenu.test.tsx`) — 3 cases covering production / development / staging.
- [x] Storybook (`com.quietmobile.storybook.debug`) on Pixel 7 — both stories render, share sheet opens, Gmail accepts the attachment.
- [x] standardDebug build with `ENVFILE=.env.development` (`com.quietmobile.debug`) on Pixel 7 — full app flow, "Share logs" appears in the community menu, share sheet opens with To: pre-filled, Gmail accepts the attachment.
- [x] standardDebug build with `ENVFILE=.env.production` (`com.quietmobile.debug`) on Pixel 7 — production env baked into the bundle; "Share logs" item is hidden in the community menu (manual confirmation).
- [ ] iOS smoke test (no simulator available in this environment).
- [ ] standardRelease (`prod` variant, Hermes + Proguard) on a CI runner — recommended as a follow-up regression layer; the runtime gate doesn't depend on minification but worth verifying once CI has a signed release build.

## Follow-ups (not in this PR)

- Log redaction in `@quiet/logger` (the actual blocker for exposing this feature publicly).
- Optional: build-time-replaced gate so production bundles dead-code-eliminate the entire `sendLogs` import, enabling a CI bundle-grep that statically proves the feature cannot ship to production. The current runtime gate is correct but leaves the strings in the bundle — fine for a sanity check, not a static guarantee.

### DIFF
--- a/packages/mobile/.storybook/index.js
+++ b/packages/mobile/.storybook/index.js
@@ -41,6 +41,7 @@ configure(() => {
   require('../src/components/FileAttachmentPreview/FileAttachmentPreview.stories')
   require('../src/components/NewUsernameRequested/NewUsernameRequested.stories')
   require('../src/components/ModalBottomDrawer/drawers/ServerOffer.drawer.stories')
+  require('../src/utils/sendLogs.stories')
 }, module)
 
 const StorybookUIRoot = getStorybookUI({
--- a/packages/mobile/src/components/ContextMenu/ContextMenu.stories.tsx
+++ b/packages/mobile/src/components/ContextMenu/ContextMenu.stories.tsx
@@ -51,9 +51,9 @@ const community_dev_items: ContextMenuItemProps[] = [
     },
   },
   {
-    title: 'Send logs to devs',
+    title: 'Share logs',
     action: () => {
-      logger.info('clicked on send logs')
+      logger.info('clicked on share logs')
       void sendLogs()
     },
   },
@@ -87,7 +87,7 @@ storiesOf('ContextMenu', module)
       />
     )
   })
-  .add('Community (dev/alpha — Send logs)', () => {
+  .add('Community (dev/alpha — Share logs)', () => {
     return (
       <ContextMenu
         title={'Rockets'}
--- a/packages/mobile/src/components/ContextMenu/ContextMenu.stories.tsx
+++ b/packages/mobile/src/components/ContextMenu/ContextMenu.stories.tsx
@@ -6,6 +6,7 @@ import { ContextMenu } from './ContextMenu.component'
 import { ContextMenuItemProps } from './ContextMenu.types'
 
 import { createLogger } from '../../utils/logger'
+import { sendLogs } from '../../utils/sendLogs'
 
 const logger = createLogger('contextMenu:stories')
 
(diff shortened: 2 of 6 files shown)


holmesworcester:
Tested on iOS too.  On iOS the email doesn't prefill. It's possible to prefill it for users with Mail.app installed but it seems better to not assume that.  Adding a note to instruct users to send to logs@tryquiet.org

PR #3209, opened 2026-05-06
fix: raise getCaptchaSiteKey ack timeout from 2s to 5s (stalls create-community under sustained latency)

holmesworcester:
## The bug

`QSSClient.getCaptchaSiteKey` ([packages/backend/src/nest/qss/qss.client.ts:285](https://github.com/TryQuiet/quiet/blob/7.1.0/packages/backend/src/nest/qss/qss.client.ts#L285)) hardcodes a **2 second** ack timeout for the first round-trip of the captcha exchange:

```ts
public async getCaptchaSiteKey(): Promise<string | null> {
  this.logger.debug('Requesting hCaptcha site key from QSS')
  try {
    const response = await this.sendMessage<GetCaptchaSiteKeyResponse>(
      WebsocketEvents.GET_CAPTCHA_SITE_KEY,
      { ts: ..., status: ... },
      true,
      2000              // <-- 2 s ack timeout
    )
```

When this timeout fires before QSS's response arrives, the create-community flow stalls **permanently**. The error is silently swallowed, no retry is scheduled, and nothing else fires the flow until the user force-quits and reopens the app.

This is a **stuck-state bug**, not a slow-network annoyance: even waiting 5 minutes won't recover; even clearing the network problem mid-stream won't recover. The only path back is a full socket reconnect.

## When users hit this

The 2 s timeout fires whenever the round-trip to QSS exceeds ~2 s, sustained. Concretely:

- **Cellular dead spots / 1-2 bar coverage.** Multi-second RTT is normal. A user trying to create their first community in an elevator, basement, or underground transit will trip this.
- **Captive WiFi.** Hotels, cafes, airports — first connections through the portal often see 3-10 s of stalling.
- **Congested public WiFi.** Conference centres, cafes — bandwidth fine on average, tail RTT brutal.
- **Cellular handover.** Phone hopping between cells (driving, walking out of a building) — bursty 1-10 s RTT.

In all these cases the user sees: tap "Create Community", spinner appears, **stays forever**. No error message, no progress. Force-quit-and-retry sometimes works (because the network was transient); sometimes the user has to leave the building to get a real connection.

## How the failure unfolds (code path)

Trace from `QSSService.connect()` to permanent stall:

1. **Socket connects successfully.** Even with high latency, `_waitForConnect` (10 s timeout in `7.1.0`) eventually wins.
2. **`QSS_CONNECTED` fires.** `QSSService._handleQssConnected` ([qss.service.ts:199-202](https://github.com/TryQuiet/quiet/blob/7.1.0/packages/backend/src/nest/qss/qss.service.ts#L199-L202)) emits `QSS_HANDLE_SIGN_IN`.
3. **`_handleQssHandleSignIn` decides to create.** For a fresh single-user owner ([qss.service.ts:238-243](https://github.com/TryQuiet/quiet/blob/7.1.0/packages/backend/src/nest/qss/qss.service.ts#L238-L243)) it calls `createCommunity()`.
4. **`createCommunity` gates on captcha.** [qss.service.ts:686-692](https://github.com/TryQuiet/quiet/blob/7.1.0/packages/backend/src/nest/qss/qss.service.ts#L686-L692). If `captchaVerified` is false, calls `requestCaptchaVerification()`.
5. **`requestCaptchaVerification` calls `getCaptchaSiteKey()` first.** Which is where the 2 s timeout lives.
6. **The timeout fires before the response arrives.** Under sustained latency ≥ ~2 s, `socket.io`'s `emitWithAck.timeout(2000)` rejects with "operation has timed out".
7. **`sendMessage` swallows the error.** [qss.client.ts:269-272](https://github.com/TryQuiet/quiet/blob/7.1.0/packages/backend/src/nest/qss/qss.client.ts#L269-L272) — `try/catch` logs and returns `undefined`.
8. **`getCaptchaSiteKey` returns null.** Sees a non-success response, logs `"Failed to get hCaptcha site key from QSS: no reason provided"`, returns null.
9. **`requestCaptchaVerification` returns false.** "Cannot request captcha verification without a site key".
10. **`_createCommunityImpl` returns false.** Outer `createCommunity` swallows it via try/catch.
11. **Nothing re-fires the flow.** And here's the dead end:
    - `_handleQssConnected` only emits `QSS_HANDLE_SIGN_IN` on the `QSS_CONNECTED` event. The socket is still connected; that event won't fire again.
    - `_handleCommunityAdded` only emits it on `COMMUNITY_ADDED`, which fires once at community-set time.
    - `_scheduleReconnect` ([qss.service.ts:488-512](https://github.com/TryQuiet/quiet/blob/7.1.0/packages/backend/src/nest/qss/qss.service.ts#L488-L512)) only schedules a reconnect when the *socket* itself errors. Pure latency doesn't break the socket; the socket stays alive forever.

So the user is stuck on a healthy socket with a defunct auto-flow. There is no internal recovery path.

### Run-time evidence

From the test's debug logs on `7.1.0`:

```
20:58:46.508 INFO  Establishing connection with QSS
20:58:53.746 INFO  Connection established                       ← +7.2 s socket.io handshake
20:58:55.748 ERROR Error while sending message to QSS: operation has timed out
20:58:55.750 INFO  Received response from QSS for hCaptcha site key request
20:58:55.750 WARN  Failed to get hCaptcha site key from QSS: no reason provided
20:58:55.751 WARN  Can't create community on QSS because captcha verification failed
[final] qssSetup=false at +67447ms
```

A single ack timeout (exactly 2 000 ms after "Connection established"), exactly one auto-flow attempt, then 60+ s of silence. The same socket would happily handle messages, but nothing asks it anymore.

## The fix

One line change in `packages/backend/src/nest/qss/qss.client.ts`:

```diff
@@ -278,7 +278,14 @@ public async getCaptchaSiteKey(): Promise<string | null> {
         status: CommunityOperationStatus.SENDING,
       },
       true,
-      2000
+      // Was 2000 — too tight under sustained downstream latency. With 3 s ±
+      // 1.5 s on the QSS link the ack can arrive 1.5–4.5 s late, the
+      // timeout fires before the response, sendMessage swallows the error,
+      // this returns null, and the create-community auto-flow stalls
+      // permanently because nothing re-fires QSS_HANDLE_SIGN_IN.
+      // 5 s matches the default `timeoutAck` in `sendMessage`.
+      5000
     )
```

`getCaptchaSiteKey` was the only QSS message in the codebase using a 2 s ack timeout — every other call to `sendMessage` either uses the default 5 s `timeoutAck` or an explicit longer one. This brings it in line.

A more durable fix (worth doing alongside, but separate concern) would be to **reschedule `QSS_HANDLE_SIGN_IN` after a transient `createCommunity` failure**, so a future timeout regression can't strand the user. Sketch:

```ts
// qss.service.ts: in _handleQssHandleSignIn
if (!result && !initStatus.qssSetup) {
  setTimeout(() => this.emit(QSSEvents.QSS_HANDLE_SIGN_IN), QSS_RECONNECT_DELAY_MS)
}
```

That would also harden against other transient causes (temporary captcha service hiccup, GEN_PUB_KEYS dropped, etc.) without requiring the user to leave/rejoin to recover. Not in this PR.

## Reproduction

A new test reproduces the failure deterministically:

[`packages/backend/src/nest/qss/stress/scenarios/stalled-under-latency.stress.spec.ts`](packages/backend/src/nest/qss/stress/scenarios/stalled-under-latency.stress.spec.ts)

The test uses [Toxiproxy](https://github.com/Shopify/toxiproxy) — a small TCP proxy that injects controllable network latency — to simulate the slow-network condition: 3 s ± 1.5 s downstream delay (modelling cellular dead spot or captive-portal-after-auth).

Four cases:

1. **PoC.** Under 3 s ± 1.5 s downstream latency, `qssSetup` must reach `true` within 60 s. **Fails on `7.1.0`** (without the fix); passes with this PR's one-line change.
2. **Diagnostic — extended wait.** Even with a 180 s wait, `qssSetup` stays false. Distinguishes "stuck" from "slow".
3. **Diagnostic — clear latency mid-test.** Removing the latency partway through does **not** recover. Confirms there is no timed retry, no progress monitor.
4. **Positive control — force a reconnect.** Explicitly toggling the proxy off/on (so the QSS socket drops and reconnects) **does** recover. Isolates the missing recovery path.

Tests 2 and 3 are negation-style diagnostics: they pass on `7.1.0` because they assert the bug exists. With the fix, those would change colour (which is fine — they're documenting the bug, not gating on it). The PoC and positive control are the load-bearing assertions.

## Before / after

Same machine, same QSS server, same Toxiproxy proxy.

**On `7.1.0` (no fix):**
```
PoC: under 3 s ± 1.5 s downstream latency, qssSetup must reach true within 60 s
✕ FAIL — qssSetup still false after 67 s
```

**With this PR's fix:**
```
PoC: ...
✓ PASS — qssSetup === true at +22 s
```

Verified across 4 consecutive runs.

## How to run

```bash
NODE_OPTIONS="--experimental-vm-modules" \
  ./node_modules/jest/bin/jest.js --config jest.stress.config.js \
  --runInBand --colors=false --forceExit \
  ./src/nest/qss/stress/scenarios/stalled-under-latency.stress.spec.ts
```

Requires a local QSS server (`npm run start:qss`) and a Toxiproxy proxy (`npm run start:qss:toxi`). Node 20.20.1 for backend tests.

## Note

Replaces holmesworcester/quiet#11, which targeted the workflow-stripped mirror at `holmesworcester/quiet:7.1.0`. This PR retargets the same patch directly at upstream `TryQuiet/quiet:7.1.0`.


holmesworcester:
Dropped the Toxiproxy stress test from this PR — it depends on `packages/backend/src/nest/qss/stress/{harness,invariants}.ts` which exist on the holmesworcester/quiet fork (added by separate stress-tooling PRs there) but are not present on `TryQuiet/quiet:7.1.0`. The webpack build was failing because the spec's `import { bootQssHarness } from '../harness'` could not be resolved.

This PR now contains only the load-bearing fix: bumping the `getCaptchaSiteKey` ack timeout from 2 s to 5 s in `qss.client.ts`. The reproduction methodology and red/green evidence in the description still apply — they just can't be re-run in this repo without the harness.

adrastaea (approved):

PR #3212, opened 2026-05-06
fix(qss): widen handshake timeout via QSS_HANDSHAKE_TIMEOUT_MS (default 30s)

holmesworcester:
## The bug

`QSSClient._waitForConnect` ([packages/backend/src/nest/qss/qss.client.ts:142-182](https://github.com/TryQuiet/quiet/blob/7.1.0/packages/backend/src/nest/qss/qss.client.ts#L142-L182)) hardcodes a **10 second** deadline for the QSS WebSocket handshake to finish:

```ts
const timer = setTimeout(() => {
  socket.off('connect', onConnect)
  socket.off('connect_error', onError)
  this.logger.error('QSS client failed to connect within timeout, closing socket')
  this._closeSocket(socket)
  reject(new QSSConnectionError(`Client didn't connect in time!`))
}, 10_000)
```

When this timer fires, `connect()` rejects, the QSS service falls back to `_scheduleReconnect` ([qss.service.ts:488-512](https://github.com/TryQuiet/quiet/blob/7.1.0/packages/backend/src/nest/qss/qss.service.ts#L488-L512)) which retries with exponential backoff starting at 50 ms.

10 s is enough for a healthy network. It is **not enough** for users on networks where round-trip latency is multiple seconds, especially when a transient TCP reset forces a fresh attempt.

## When users hit this

`7.1.0` connects to QSS over plain WebSocket (`CLIENT_TRANSPORTS = ['websocket']` — [qss.const.ts:1](https://github.com/TryQuiet/quiet/blob/7.1.0/packages/backend/src/nest/qss/qss.const.ts#L1)), so the connect path is: TCP → TLS → HTTP upgrade → engine.io open packet. That's at least one full round-trip after the upgrade before `socket.io` fires `connect`.

Real-world conditions that produce >10 s here:

- **Cellular dead spots / 1-2 bar coverage.** Multi-second RTT, occasional TCP RST as towers hand off. A user opening Quiet in an elevator, on a train, or in a basement.
- **Cellular handover.** When a phone hops between cells (driving, walking out of a building) packets get queued and bursty latency goes 1-10 s.
- **Captive WiFi after auth-redirect.** Hotels, cafes, airports. Connections see 5-30 s of stalling while the portal authenticates the device.
- **Congested public WiFi.** Conference centres, cafes — bandwidth is fine on average but tail RTT is brutal, and access points sometimes RST stale connections.

In each case, the user-visible symptom is: app gets stuck on "Connecting to QSS…", eventually fails with a generic error, and the user has to retry by hand. Reconnect *would* succeed if the per-handshake budget were a bit larger.

## How the failure unfolds

Each retry from `_scheduleReconnect` calls `_waitForConnect` again with the same 10 s ceiling. The schedule starts at 50 ms and doubles up to 60 s. In a 30 s window:

| t (s) | event |
| ---: | --- |
| 0 | attempt 1 starts |
| 10 | attempt 1 hits the 10 s `_waitForConnect` ceiling → reject; retry in 50 ms |
| 10.05 | attempt 2 starts |
| 20.05 | attempt 2 hits the same ceiling → reject; retry in 100 ms |
| 20.15 | attempt 3 starts |
| 30+ | attempt 3 still in flight |

Three attempts, each capped at the same 10 s ceiling. None of them finishes when the WebSocket handshake legitimately needs 11-15 s.

The only thing that ever lets the user through is a *better* network — when their connection improves, one of the retries lands inside the budget. If their network stays bad, they're stuck.

## The fix

Make the per-handshake deadline a configurable constant, default 30 s, env-overridable via `QSS_HANDSHAKE_TIMEOUT_MS`. Values are clamped to a minimum of 1 s.

Files changed:

- `packages/backend/src/nest/qss/qss.const.ts` — new `QSS_HANDSHAKE_TIMEOUT_MS` constant
- `packages/backend/src/nest/qss/qss.client.ts` — replaces the hardcoded `10_000` with the constant

Why 30 s:

- It's the same order of magnitude as a single bad-network round-trip plus the engine.io handshake — letting the worst-case slow handshake actually complete instead of burning the whole reconnect budget on three identical attempts.
- It aligns with `socket.io`'s own default `timeout` (20 s) with a small margin.
- The reconnect schedule's exponential backoff already saturates at 60 s, so the ceiling on "how long before the user knows something's wrong" doesn't change much for an actually-dead endpoint.

The trade-off: a longer per-handshake timeout means a slightly slower fail-fast when the QSS endpoint is genuinely unreachable (wrong DNS, server down). The fail-fast was already not very fast — three attempts × 10 s + retry delays = ~30 s before the user sees anything. With the fix, one attempt × 30 s + retry delays ≈ same ballpark, but with a much higher chance of success on a slow-but-working network.

Alternative considered: keep 10 s and add an internal retry inside `_waitForConnect`. Structurally trickier (the retry has to drop the now-broken socket and re-call `socket.connect()` from scratch, plus reconcile with captcha re-verification in the outer service) and adds more knobs without changing the user-visible outcome.

## Reproduction

A new test reproduces the failure deterministically:

[`packages/backend/src/nest/qss/stress/scenarios/connect-timeout-under-latency.stress.spec.ts`](packages/backend/src/nest/qss/stress/scenarios/connect-timeout-under-latency.stress.spec.ts)

The test uses [Toxiproxy](https://github.com/Shopify/toxiproxy) — a small TCP proxy that injects controllable latency, jitter, and connection resets — to simulate the slow-network conditions described above. Specifically:

- ~4.6 s downstream latency with ±1.4 s jitter (modelling cellular dead spot)
- ~15 % upstream `reset_peer` (modelling a flaky middlebox / captive portal)
- A 6.3 s outage early in the connect attempt

Three test cases:

1. **Single attempt.** Flaky on `7.1.0` (~20 % red); reliably green with the fix.
2. **5-attempt aggregate.** Boots a new socket per iteration. Reliably red on `7.1.0`; green with the fix.
3. **Recovery.** Apply the toxics, miss the deadline, then *clear* the toxics. Asserts the service reconnects within another 30 s. **Passes on `7.1.0`** — confirms the bug is purely a tight-deadline issue, not a state-machine wedge or a leak.

## Before / after

Same machine, same QSS server, same Toxiproxy proxy.

**On `7.1.0` (no fix), 5-attempt aggregate:**
```
Tests:       1 failed, 2 skipped, 3 total
  expect(failures).toEqual([])
  - Array []
  + Array [3]      ← 3 of 5 attempts failed
Time:        101.3 s
```

**With this PR's fix:**
```
Tests:       2 skipped, 1 passed, 3 total
Time:        77.1 s
```

Full spec (3 cases including recovery):
```
Tests:       3 passed, 3 total
Time:        90.2 s
```

## How to run

```bash
NODE_OPTIONS="--experimental-vm-modules" \
  ./node_modules/jest/bin/jest.js --config jest.stress.config.js \
  --runInBand --colors=false --forceExit \
  ./src/nest/qss/stress/scenarios/connect-timeout-under-latency.stress.spec.ts
```

Requires a local QSS server (existing `npm run start:qss`) and a Toxiproxy proxy on `host:8474` (`npm run start:qss:toxi`). Node 20.20.1 for backend tests.

## Note

Replaces holmesworcester/quiet#10, which targeted the workflow-stripped mirror at `holmesworcester/quiet:7.1.0`. This PR retargets the same patch directly at upstream `TryQuiet/quiet:7.1.0`.


holmesworcester:
Dropped the Toxiproxy stress test from this PR — it depends on `packages/backend/src/nest/qss/stress/{harness,invariants}.ts` which exist on the holmesworcester/quiet fork (added by separate stress-tooling PRs there) but are not present on `TryQuiet/quiet:7.1.0`. The webpack build was failing because the spec's `import { bootQssHarness } from '../harness'` could not be resolved.

This PR now contains only the load-bearing fix: `QSS_HANDSHAKE_TIMEOUT_MS` constant + the `qss.client.ts` change. The reproduction methodology and red/green evidence in the description still apply — they just can't be re-run in this repo without the harness.

packages/backend/src/nest/qss/qss.const.ts
```diff
@@ -3,6 +3,30 @@ export const QSS_RECONNECT_DELAY_MS = 50
...
+ * inside the same `_waitForConnect` and is consistent with socket.io's own
+ * default `timeout` (20 s) plus a small margin for the slow-network case.
+ *
+ * Override with the QSS_HANDSHAKE_TIMEOUT_MS env var if a deployment wants to
+ * tighten or further widen it. Values are clamped to a minimum of 1 s to
+ * avoid accidentally disabling the timer.
+ */
+export const QSS_HANDSHAKE_TIMEOUT_MS = Math.max(
+  1_000,
+  Number(process.env.QSS_HANDSHAKE_TIMEOUT_MS ?? 30_000) || 30_000
```
adrastaea, line 27: Is there a particular reason you want to make this configurable by env var?

adrastaea (changes requested):
I think the environment variable config isn't very useful. I'm going to implement an exponentially increasing timeout capped at 30s.

PR #3229, opened 2026-05-13
fix(backend): purge uploads and downloads dirs on leave community (#3225)

### LINKED ISSUE #3225: Leaving the community doesn't clean all data from previous community
**Version**: 7.1.0-alpha
**System**: all

**Issue**: Leaving the community doesn't clear all data from the previous community, such as uploaded and downloaded files. 

**Steps to recreate**:
1. Be a part of a community that exchanges pictures.
2. Leave the community.
3. Check your Quiet "downloads" and "uploads" files - they should be empty, but are not.

holmesworcester:
## Human Summary

There are two commits. One includes additional directories in purge. The other ensures that if purge is interrupted e.g. by the mobile OS, a "leave in progress" flag is stored and purge will continue on restart.

Without this second commit, an interrupted purge would be forgotten and would not happen until the user created or joined a new community.  

I don't have a Windows machine to test on, so while I don't think there is any Windows-specific issue here, I would be missing it if there was.  I could catch this with an end-to-end test if we think that's appropriate.

## Agent Summary
- Adds `uploads` and `downloads` to the prefix filter in `_purgeDataDirectories()` so user-content files (sent/received images, attachments) are removed when a user leaves a community. Closes #3225.
- Adds a regression test that seeds 25 varied sentinel files (UUID-like names, unicode, spaces, hidden files, nested subdirs, 0-byte files) across `uploads/`, `downloads/`, and four already-purged dirs as anchors; asserts the entire `quietDir` tree returns to baseline after `purgeData()`.
- **Defense in depth**: also runs the same purge at backend startup when no community is active in storage. Catches the case where a leave-community crashes mid-way (process killed, OS-terminated on mobile, power loss) and the user reopens the app but doesn't immediately create or join another community — today those orphaned files persist indefinitely.

## Why these dirs
`uploads/` and `downloads/` are constructed at `ipfs-file-manager.service.ts:177` (staging for files the user is uploading) and `:955` (where downloads land). They've never been in the purge filter — confirmed via blame; no version of this function has included them, and no comment documents an exclusion. Sin of omission, not a deliberate choice. Filed as both bug and security in #3225 since leftover user content survives a "leave community."

## Lock tolerance / Windows
The two new prefixes flow through the same `removeFilesFromDir(..., { throwOnError: false, maxRetries: 1, retryDelay: 100 })` call as the existing seven, so they inherit the lock-tolerance behavior added in `c7d0837932`. The existing `purgeData continues when a data directory is locked` test covers the umbrella case.

## Mobile
Mobile inherits this fix transitively. `state-manager/src/sagas/app/closeServices.saga.ts:11` emits `SocketActions.LEAVE_COMMUNITY` via `socket.emitWithAck`; the backend handles it at `connections-manager.service.ts:828` and runs the full `leaveCommunity()` chain — so the same `purgeData()` runs on iOS and Android via nodejs-mobile. The mobile saga blocks on the backend's ack callback before proceeding, so no race with subsequent native cleanup.

## quietDir paths (where to look when testing)
The backend's `quietDir` is resolved per-platform:
- **Desktop prod**: `~/.config/Quiet7/Quiet/` (Linux), `~/Library/Application Support/Quiet7/Quiet/` (macOS), `%APPDATA%\Quiet7\Quiet\` (Windows). Comes from `loadMainEnvs.ts:32` (`DESKTOP_DATA_DIR = "Quiet7"`) → `backendManager.ts:201` (joins with `'Quiet'`) → `app.module.ts:87`.
- **Desktop dev** (`NODE_ENV=development`): `Quietdev` in place of `Quiet7`.
- **iOS**: `Documents/backend/files7/` via `DataDirectory.create()` → `--dataPath` arg → `app.module.ts:87`.
- **Android**: `<context.filesDir>/backend/files7/` via `Utils.kt:13` → `--dataPath` arg → `app.module.ts:87`.

The `~/.quiet/` value of the `QUIET_DIR_PATH` const is only a fallback when no `APP_DATA_PATH`/`--dataPath` is provided — never in deployed builds, only relevant for backend Jest tests.

## Test plan
- [x] `cd packages/backend && npx cross-env NODE_OPTIONS="--experimental-vm-modules" jest storage.service.spec --runInBand --forceExit` — **13 passed** locally (12 preexisting + 1 new). Test fails on `7.1.0` without this commit (verified before the fix landed).
- [x] `cd packages/backend && npx cross-env NODE_OPTIONS="--experimental-vm-modules" jest connections-manager.service.spec --runInBand --forceExit` — **15 passed** locally (13 preexisting + 2 new for the startup-purge defense).
- [ ] Desktop manual (leave path): create a community, exchange an image with another peer, leave, inspect `~/.config/Quiet7/Quiet/uploads/` and `~/.config/Quiet7/Quiet/downloads/` (substitute `Quietdev` for dev) — both should be empty/absent.
- [ ] Desktop manual (startup-purge defense): after a clean leave, drop `~/.config/Quiet7/Quiet/uploads/sentinel.bin`, fully quit the app, relaunch — sentinel should be gone before the join/create screen renders.
- [ ] Android manual: `adb shell run-as com.quietmobile ls -la files/backend/files7/uploads` before and after leave — populated → empty.
- [ ] iOS manual: same check via Xcode's container-inspection on `Documents/backend/files7/{uploads,downloads}`.


holmesworcester:
I don't have a Windows machine to test on, so while I don't think there is any Windows-specific issue here, I would be missing it if there was. 

I could catch this with an end-to-end test if we think that's appropriate.

islathehut (approved):

PR #3263, opened 2026-06-02
feat(2571): DMs for desktop

### LINKED ISSUE #2571: User should be able to start group DMs
Slack and Messenger let you start fixed-membership group DMs, where there is a single canonical thread for every group of users. We should do this too. 

This is a ticket for a fast initial implementation. 

Image

---

Figma: https://www.figma.com/design/tXuRsUfP6VnSv99dox00C1/Direct-Messages--DMs-?node-id=821-22035&t=A3haiOXjffUkbPbg-0

Prototype: https://www.figma.com/design/tXuRsUfP6VnSv99dox00C1/Direct-Messages--DMs-?node-id=823-16422&t=A3haiOXjffUkbPbg-0

Acceptance criteria:

1. User can start DM thread with other users in the same workspace
2. Messages are encrypted to participants

Nice to have:

1. Matches designs (user can start initial message immediately)
2. There is a canonical thread for each group of users (channel id is canonicalized hash of user id's e.g.)  

Non-requirements: 
1. It is impossible at the protocol level to add others at this thread; if members want to leak their messages they can!

islathehut:
### Pull Request Checklist


islathehut:
NOTE: There is still work to be done to cleanup/refine some things so reviews can wait until this comment is gone.

holmesworcester:
I'm trying to figure out if the "ID deterministic from participants" is a good idea. Some thoughts:

1. We can always enforce "same DM" in the frontend without making the ID the same. 

2. Enforcing it in the backend by making ID deterministic from members would be helpful for concurrency (Alice and Bob both create a Alice, Bob, Carol DM) *but* we'll have different keys anyway, and different roles, is this too weird?

I think what happens is, right now you'll essentially have n (concurrent DM creation) different cryptographic/role contexts for the same channel id, and all messages will be visible. Is that right? That gives everyone the properties they want at this stage, or seems to! 

@islathehut can you vouch for this making sense and behaving well in the concurrent case? 

How do you feel about random id's and frontend-only enforcement of "same members, same DM channel"?

PR #3355, opened 2026-07-13
feat(3354): Randomly generate private channel role names

### LINKED ISSUE #3354: Use random private channel role names
# Description

Private channels currently use deterministic role names.  We want them to be random and completely disconnected from the channel names and IDs.

# Acceptance Criteria

* Randomly generate role names for private channels

islathehut:
### Pull Request Checklist


packages/backend/src/nest/auth/services/members/user.service.ts
```diff
@@ -93,6 +95,12 @@ class UserService extends ChainServiceBase {
     return users[0]
```
adrastaea, line 98: This can return undefined as long as throwOnMissing is not true, so the type definition for the returned value is incorrect. Same with `getMyUser`.

adrastaea (changes requested):
one type issue, but otherwise fine.

packages/backend/src/nest/auth/services/members/user.service.ts
islathehut: Tweaked it.  Confirmed that when `throwOnMissing` is true the return signature is `Member` and when its false the return signature is `Member | Undefined`.

adrastaea (approved):

PR #3385, opened 2026-07-27
fix: Clear connected peers interval on pause and ensure libp2p doesn't redial after pause

islathehut:
### Pull Request Checklist

### DIFF
--- a/packages/backend/src/nest/libp2p/libp2p.service.ts
+++ b/packages/backend/src/nest/libp2p/libp2p.service.ts
@@ -43,6 +43,7 @@ import { LocalDbService } from '../local-db/local-db.service'
 import { TimedQueue } from '../common/timed-queue'
 import { defaultLogger } from './libp2p.logger'
 import { QSSService } from '../qss/qss.service'
+import { Libp2pConnectionGater } from './libp2p.connection-gater'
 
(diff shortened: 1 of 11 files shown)


packages/backend/src/nest/libp2p/libp2p.service.ts
```diff
@@ -339,7 +376,8 @@ export class Libp2pService extends EventEmitter implements OnModuleDestroy {
     await this.hangUpPeers()
     this.dialedPeers.clear()
     this.connectedPeers.clear()
-    // await this.libp2pDatastore?.deleteKeysByPrefix(Libp2pDatastorePrefix.PEERS)
+    // remove peers from the datastore to avoid libp2p auto-redialing after the pause
+    await this.libp2pDatastore?.deleteKeysByPrefix(Libp2pDatastorePrefix.PEERS)
```
adrastaea: This is currently a no-op because `peers` does not match `/peers/`, so nothing is deleted. Also, deleting the entire peer namespace will also delete the self records, so that could be detrimental. It would be better to enumerate the peer records and use the `.delete()` api to clear remote peer records. The test you added just proves that this command is called, not that it actually deleted the records which is why it is passing.

adrastaea (changes requested):

PR #3394, opened 2026-07-29
fix(3392): Fix validations of private channel deletions

### LINKED ISSUE #3392: Can't delete private channel
**Version**: 9.0.0-alpha.3
**System**: all
**Server**: tor and qss

**Issue**: Can't delete private channels. Checked on different systems, and both Tor and QSS-based quiet. Public channels can be deleted.

**Steps to recreate**: try deleting a private channel. Easy to reproduce. 

https://github.com/user-attachments/assets/5774c239-899b-41b3-be54-5de595c1c79c

islathehut:
### Pull Request Checklist

### DIFF
--- a/packages/backend/src/nest/storage/channels/orbitdb/ChannelMetadataAccessController.ts
+++ b/packages/backend/src/nest/storage/channels/orbitdb/ChannelMetadataAccessController.ts
@@ -23,6 +23,7 @@ import { createLogger } from '../../../common/logger'
 import { QuietLogger } from '@quiet/logger'
 import { posixJoin } from '../../orbitDb/util'
 import type { SigChain } from '../../../auth/sigchain'
+import type { PrivateChannelMappings } from '../channels.types'
 
 const TYPE = 'channelmetadataaccess'
 const codec = dagCbor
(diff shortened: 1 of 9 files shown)


packages/backend/src/nest/storage/channels/orbitdb/ChannelMetadataAccessController.ts
```diff
@@ -175,9 +177,29 @@ export class ChannelMetadataAccessController {
...
+      } else {
+        const key = entry.payload.key
+        if (key == null) {
+          this.logger.warn(`Channel metadata DEL rejected due to missing key`, {
+            writerId: writerIdentity.id,
+          })
+          return false
+        }
+        try {
+          const channelRoleMappings = await config.getPrivateChannelsByRolename()
```
adrastaea, line 192: It seems like this can fail on restart when initializing a db with existing private channels because `getPrivateChannelsByRolename` will throw when `this.privateChannels` is unset, but it is only set when `openPrivateChannelsDb` returns. So when we initialize the database using this access controller the necessary map is not yet initialized which I think can cause valid entries to be invalidated. Have you tested reinitializing the app with existing private channels?

adrastaea (changes requested):

packages/backend/src/nest/storage/channels/orbitdb/ChannelMetadataAccessController.ts
islathehut: Yea, I've tested restarting the app both on the owner and second user and private channels remain usable and the deletions are still honored.  I haven't seen any issues with anything being invalidated after already being validated.

adrastaea (approved):

PR #3400, opened 2026-07-31
Device Linking 2610

### LINKED ISSUE #2610: User can link devices for all communities
**Update: we can ignore this complexity because we just have one community for now. Let's do straightforward LFA device linking in current community. **

(In the future when we have multiple communities: automatic updates across all devices is way better and less confusing. Piecemeal is closer to what LFA does and we might want to do that first.)

Figma: https://figma.com/design/3RcrYKRTiFY87TpFSqZyj4/Device-linking?t=pvUNvz05PMsZGmX3-0

-- 

Users must be able to link devices. Currently, we believe it's better that all communities are mirrored across all devices, since in the case of a lost device you easily unlink it and remove it from all your communities automatically, in one step. 

(Otherwise you might have to unlink it from 10 communities!)

**Technical details**

We would still use LFA's mechanism for linking/unlinking devices.  However we would also create a p2p network that was centered around the user. Whenever the user joined a new community on one device, they would create device linking invites for that community to all their devices at the LFA layer, and then transmit them at this new user-device-network layer. 

The user-device-network layer has the same eventual consistency property our existing network has, so as these devices come online they would get these invites and join the new community. When a user leaves a community on one device, this propagates to all their devices. 

We may also want to add a QSS instance centered around each device that is created when the user first attempts to link a device.

adrastaea:
### Pull Request Checklist


adrastaea:
Screenshot 2026-08-04 at 12 47 33 PM

packages/backend/src/nest/admission/admission-coordinator.service.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+  it('refuses ordinary-member fallback after QSS ownership is claimed', async () => {
+    request.kind = AdmissionKind.MEMBER
+    qssService.startPreparedAdmission.mockRejectedValueOnce(new Error('LFA start failed'))
+
+    await expect(coordinator.coordinate(request, runtime)).rejects.toThrow('LFA start failed')
+    expect(localDbService.claimAdmissionTransport).toHaveBeenCalledWith(request.communityId, AdmissionTransport.QSS)
+    expect(runtime.startP2p).not.toHaveBeenCalled()
+  })
+
+  it('honors stored P2P ownership without starting QSS or claiming again', async () => {
```
holmesworcester, line 124: Just noting that it's very awkward that we have to handle cases where users in a QSS community can stay as p2p-only. That's what this is for, right?

packages/backend/src/nest/auth/services/invites/invite.service.ts
```diff
@@ -64,13 +65,21 @@ class InviteService extends ChainServiceBase {
     }
   }
 
-  public createDeviceInvite(validForMs: number = DEFAULT_INVITATION_VALID_FOR_MS, seed?: string): InviteResult {
-    const expiration = (Date.now() + validForMs) as UnixTimestamp
+  public createDeviceInvite(
+    validForMs: number = DEFAULT_DEVICE_INVITATION_VALID_FOR_MS,
+    seed?: string
+  ): DeviceLinkInvite {
+    const expiresAt = (Date.now() + validForMs) as UnixTimestamp
```
holmesworcester, line 72: validForMs: 0: createUserInvite treats 0 as "never expires"; here 0 silently yields Date.now(), already expired.

Not sure if this matters but this section looked weird and this was an LLM nit.

packages/types/src/errors.ts
```diff
@@ -40,6 +40,7 @@ export enum ErrorMessages {
 
   // Community
   COMMUNITY_LAUNCH_FAILED = 'Could not launch community',
+  COMMUNITY_ALREADY_INITIALIZED = 'A community is already initialized',
```
holmesworcester, line 43: Is this error user-facing? Is it log-facing?

packages/types/src/network.ts
```diff
@@ -47,6 +52,11 @@ export type InvitationAuthDataV5 = InvitationAuthDataV4 & {
   salt: string
 }
 
+export type DeviceInvitationAuthData = InvitationAuthDataV4 & {
+  userId: string
+  userName: string
+}
+
```
holmesworcester, line 59: Can we deprecate these older invite formats yet?

packages/backend/src/nest/connections-manager/connections-manager.service.spec.ts
```diff
@@ -145,6 +237,27 @@ describe('ConnectionsManagerService', () => {
     expect(launchCommunitySpy).not.toHaveBeenCalled()
   })
 
+  it('purges an interrupted device link on startup when no admitted sigchain was persisted', async () => {
```
holmesworcester, line 240: Are we sure this is the behavior we want? 

In p2p mode on iOS, e.g., linking a device will be really unstable because it will keep dying before it connects. I think for p2p we want *eventual success*.

packages/backend/src/nest/connections-manager/connections-manager.service.ts
```diff
@@ -763,6 +805,59 @@ export class ConnectionsManagerService extends EventEmitter implements OnModuleI
...
+    if (!isPSKcodeValid(inviteData.psk)) {
+      emitError(this.serverIoProvider.io, {
+        type: SocketActions.LINK_DEVICE,
+        message: ErrorMessages.NETWORK_SETUP_FAILED,
+        community: payload.id,
+      })
+      return
+    }
+
+    const communities = (await this.localDbService.getCommunities()) ?? {}
```
holmesworcester, line 828: one note is that joinCommunity above goes straight to erasePreviousCommunityArtifacts() with no equivalent check, though it has a frontend check.

Should we add a backend check for joinCommunity?

holmesworcester (commented):
Let's discuss this all again before merging from a security standpoint, but this is what I've got from looking at the code. No big comments.

PR #3401, opened 2026-08-03
fix(3393): Ensure channel deletions are honored

### LINKED ISSUE #3393: Deleting channels on the Tor version is syncing poorly
**Version**: 9.0.0-alpha.4, tor version
**System**: all

**Issue**: Deleting channels on the Tor version is syncing poorly. I could create a situation in which the Owner deleted one of the channels when another user was in the process of joining. After joining, said user still had this channel and could send messages. Tor-syncing often causes issues but what is concerning in this case is that even after the user was fully synced, they still had this channel on their list and could send messages that, if another user clicked in, would send them to the non-existing channel (see screenshot from Windows that received a message on a deleted channel and could enter this channel after clicking on the notification).

Image

**Steps to reproduce**: it requires a few devices and a bit of trial and error, since I'm not sure at which point the problem is created. What I suggest it to create one tor-based community with a few public channels and join with two other devices while deleting those channels and creating new ones. I was able to reproduce it twice while following this path. 

Below you'll find logs from the first time I've spotted this issue:
- mac - Owner
- android - user who can still see the removed channel and is able to send messages through it
- windows - user that correctly synced deletion of a channel but now still gets notifications from it and through notification is able to enter the channel 

MacOS:
[Quiet9-mac-owner.zip](https://github.com/user-attachments/files/30504084/Quiet9-mac-owner.zip)
Android:
[quiet-data-2026-07-29-android-user.zip](https://github.com/user-attachments/files/30504115/quiet-data-2026-07-29-android-user.zip)
Windows:
[win-user-29-07-26.zip](https://github.com/user-attachments/files/30504132/win-user-29-07-26.zip)

islathehut:
### Pull Request Checklist

### DIFF
--- a/packages/backend/src/nest/storage/orbitDb/keyValueIndexedValidated.ts
+++ b/packages/backend/src/nest/storage/orbitDb/keyValueIndexedValidated.ts
@@ -47,6 +47,8 @@ const valueEncoding = 'json'
 
 const logger = createLogger('orbitdb:keyValueIndexedValidated')
 
+const isValidOp = (op: string): boolean => op === OrbitDbOp.PUT || op === OrbitDbOp.DEL
+
 /**
  * Defines an index for a KeyValue database.
  * @param {string} [directory] A location for storing the index-related data
@@ -60,12 +62,8 @@ const Index =
       path: posixJoin(directory ?? './level', '/_indexedEntries/'),
       valueEncoding,
     })
-    const indexedKeysWithOp = await LevelStorage({
-      path: posixJoin(directory ?? './level', '/_indexedKeysWithOps/'),
-      valueEncoding,
-    })
 
-    const update = async (log: LogType, entry: LogEntry) => {
+    const update = async (log: LogType, entry: LogEntry, traverseFullLog = false) => {
       const keys = new Set()
       const toBeIndexed = new Set()
       const latest = entry.hash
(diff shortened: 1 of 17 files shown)


packages/backend/src/nest/storage/orbitDb/keyValueIndexedValidated.ts
```diff
@@ -87,17 +93,23 @@ const Index =
...
+              await index.put(key as string, encodeEntry(entry))
+              await indexedKeysWithOp.put(key as string, OrbitDbOp.PUT)
+            }
             await indexedEntries.put(hash, true)
-          } else if (op === 'DEL' && !keys.has(key) && isValid) {
+          } else if (op === OrbitDbOp.DEL && isValid) {
             keys.add(key)
             await index.del(key as string)
             await indexedEntries.put(hash, true)
+            await indexedKeysWithOp.put(key as string, OrbitDbOp.DEL)
```
islathehut: This is the fix.  Basically when the deletions were processed we would skip over marking the put hashes as indexed and that resulted in future reindexing treating them as fresh puts since the check for an entry being indexed is based on the hash.

adrastaea: I think there may be an issue with this though that will cause `PUT("key", "old value") -> DEL ("key") -> PUT("key", "new value")` to result in the key to still be indexed as deleted because we are storing the deletion of key persistently and skipping processing the new `PUT` because it was previously deleted. You can add
```
  it('allows a newer PUT to recreate a deleted key', async () => {
    orbitDbService = await module.resolve(OrbitDbService)
    await orbitDbService.create(ipfsService.ipfsInstance!)

    const store = await orbitDbService.open<KeyValueIndexedValidatedType<string>>(
      `recreated-key-test-store-${Date.now()}`,
      {
        Database: KeyValueIndexedValidated(),
        AccessController: IPFSAccessController({ write: ['*'] }),
        sync: false,
      }
    )

    await store.put('key', 'old value')
    await store.del('key')
    await store.put('key', 'new value')

    await expect(store.get('key')).resolves.toBe('new value')
  })
```
to the orbitdb service tests to prove this.

adrastaea (changes requested):

adrastaea (approved):

PR #3413, opened 2026-08-19
Fix/appbar contextmenu a11y hitarea

### LINKED ISSUE #3372: Tap-able area on "..." menu in top right is too small
The tap-able area on the "..." button in the top right seems too small. It's hard to hit instinctively. 

Worth checking to make sure it's standard, and auditing this generally.

kingalg:
### Pull Request Checklist

- [x] I have linked this PR to a related GitHub issue. - this fixes https://github.com/TryQuiet/quiet/issues/3372

Fixes https://github.com/TryQuiet/quiet/issues/3372 and additionally changes the size of other clickable elements in the AppBar that were not up to a11 (button Done in adding members to the private channel). Tests updated accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * iOS tor process lifecycle improvements solving crashes and improving performance [#3349](https://github.com/TryQuiet/quiet/issues/3349)
 * Update LFA to remove flaky timestamp validator [#3365](https://github.com/TryQuiet/quiet/issues/3365)
 * Improve image compression efficiency [#3364](https://github.com/TryQuiet/quiet/issues/3364)
+* Fix tapable area in two places in the Appbar [#3372] (https://github.com/TryQuiet/quiet/issues/3372)
 
 ## [8.0.0]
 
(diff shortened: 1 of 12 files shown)


holmesworcester (approved):

holmesworcester (changes requested):
Does this nitpick make sense?

> Typography doesn't set allowFontScaling={false} anywhere in the mobile package, so text scales with the OS accessibility font-size setting. With a hard width: 64, "Done" at a large accessibility scale will wrap to two lines inside a 50dp box in a 52dp bar. Before this PR the button had no fixed width, so large text simply made it wider on one line. Using minWidth: 64 keeps the a11y floor without capping growth. The icon buttons don't have this problem — their children are fixed-size 16dp images.

kingalg:
@holmesworcester, changing width to minWidth makes sense here, you are absolutely right! I fixed that and also added a few accessibility changes to the buttons we work on (and updated the tests for them). Let me know if you want me to add an explanation, but it shouldn't block anything anyway since the change is only for screen reader users.

holmesworcester (approved):

Issue #519, opened 2022-05-19
Emoji-only messages should have larger font size


holmesworcester:
Most messaging apps increase the font size of emoji-only messages because this makes the emoji more visible and fun. We should too! 

Use Slack, Signal, or WhatsApp as a reference for emoji size.

Bonus: in-line emojis should have a larger font size too. (Slack seems to render them as images at 22px.)

mike-kiss:
Hi @leblowl and @EmiM 👋 I'm a new contributor getting started with this issue. I have code ready to go, but am running into a weird situation with the pre-commit hooks. When I try to commit, lerna runs and I see my code change from the ' character to the " character and then back to ' again, after which it reports eslint/prettier errors as if it settled on " . I'm assuming this is a config problem on my end, but if you've seen this before, please let me know! Screenshot attached.

![Image](https://github.com/TryQuiet/quiet/assets/142341848/c3fd7e24-9806-4256-a08d-014cac2ab530)

islathehut:
I added that pre-commit hook and it seems to have issues with linting on files that aren't configured to be linted.  I haven't had a chance to look into it but for now if you run the linter with --fix (you can do this at the root by running _npm run lint:all_) then you can commit with --no-verify.  This is just an issue with lint-staged which is only on the pre-commit.

mike-kiss:
Thanks! This worked and I've put up a PR. Let me know if I missed anything or if there are needed changes.

holmesworcester:
I think just that issue with tests that Lucas replied to. Thank you for this! Quiet will be 50% more emotive!

kingalg:
@holmesworcester, it's not in the issue specifications, but was this only for desktops? It's implemented correctly on desktop, but on iOS and Android, the size didn't change.

holmesworcester:
Let's create a new issue for mobile that refers to this issue. Can you do that? For these 'good first issue' issues we are okay with people doing desktop only.

kingalg:
Sure, this is an issue for mobile: Use Slack, Signal, or WhatsApp as a reference for emoji size. And I'm closing this one. 

Version: 2.2.0-alpha.0

holmesworcester:
Fixed — closing.

Implemented by 71c8b226c ("Feat/519 big emoji messages", #2389, 2024-04-03), which predates the 9.0 line and ships in `@quiet/desktop@9.0.2`.

Current state on `origin/develop`:
[...]

PR #3414, opened 2026-08-24
Feature/2422 emoji only message font size

### LINKED ISSUE #2422: Emoji-only messages should have larger font size on mobile
Thanks to https://github.com/TryQuiet/quiet/issues/519, we already have emoji-only messages with a larger font on the desktop; we need to implement the same for mobiles. 

Use Slack, Signal, or WhatsApp as a reference for emoji size.

kingalg:
### Pull Request Checklist

- [x] I have linked this PR to a related GitHub issue - https://github.com/TryQuiet/quiet/issues/2422

Fixes #2422 — emoji-only messages now render at fontSize: 28 instead of the default 14, matching existing desktop behavior (#519 ).

- **packages/common/src/index.ts**: exported isAllEmoji — it already existed and was already tested, but wasn't part of @quiet/common's public API (desktop reaches it via a deep relative import instead of the package entry point).
- **Message.component.tsx**: the message-body markdown rule now picks fontSize based on isAllEmoji(message.message).
- **Message.test.tsx (new file)**: 3 focused tests — default text, emoji-only, and mixed text+emoji (the negative case, added after review). Verified they actually catch a regression by deliberately breaking the fix and confirming the test fails with a clear diff, then restored it.

I had a dilemma about what to do with tests because we have a disabled message.test already but at the end decided on the following approach: 
**New test file vs. reviving Message.test.disabled.tsx**: that file is a pre-existing, already-broken ~1400-line snapshot (avatar SVG + MathJax rendering), disabled during a past RN upgrade and not run by Jest. Fixing it properly is a separate, unrelated effort, so it's left untouched — flagged in-line with a one-line comment pointing at it so it's not mistaken for overlapping coverage. I'm gonna check and, if needed and it's not created yet, create an issue to fix the disabled test.


holmesworcester (approved):
Approved, but if you'd like to do this bonus item from the linked issue, that would be great! 

I don't know if desktop does this yet but I *think* it does

> Bonus: in-line emojis should have a larger font size too. (Slack seems to render them as images at 22px.)

kingalg:
**TLDR**: @holmesworcester the requested change is done; tests are modified and slightly improved; It wasn’t done on desktop, and we still need to do it.





This turned out to be bigger than I thought (or I kept running into things I wanted to fix). Several points to explain:

* We don’t have this implemented on desktop; I’ll make an issue for that, but here I wanted to focus on mobile.
* We need to consider using emoji-regex instead of isAllEmoji; it’ll fix some of our issues, like automatically changing numbers to emojis. I don’t know why we didn’t use it; maybe there is a legit reason behind it, but it’s worth investigating and potentially changing for both desktop and mobile. I don’t have enough info to handle it in this PR.
* You’ll see that I moved the emoji-sizing helper functions out of Message.component.tsx into a new Message.utils.tsx. I did it because it matches a pattern already used elsewhere in this codebase (Typography.utils.ts). This makes them directly testable without the full component + mock setup and keeps the component file focused on rendering rather than logic.
* Another change in the tests you may have questions about (**packages/mobile/src/components/Message/Message.test.tsx**): Our test mock for the markdown library was oversimplified in a way that let a real bug slip through. The emoji-sizing feature looked correct in tests but didn't actually work in the app. I fixed the mock to accurately simulate how the library renders text, so tests now catch this class of bug instead of hiding it.
* I decided to add some exclusions that we didn't have before, following other similar apps' examples (like code_inline/code_block/fence).



Changes that I’ve made:

* **packages/common/src/emojis.ts** — added splitEmoji(), splits text into emoji/non-emoji runs.
* **packages/common/src/emojis.test.ts** — tests for splitEmoji.
* **packages/mobile/src/components/Message/Message.utils.tsx** — new file: emoji-sizing helpers (renderWithInlineEmoji, isPlainMessageText), moved out of the component for testability.
* **packages/mobile/src/components/Message/Message.component.tsx** — inline emoji in a message now render bigger (22px), scoped to plain message text only.
* **packages/mobile/src/components/Message/Message.test.tsx** — tests for the above; mock rewritten to match how the markdown library actually renders text.

holmesworcester (approved):

holmesworcester:

Three files conflicted, and two of them were more than they looked.

**`packages/common/src/index.ts`** — union. This branch's `emojis` export beside `develop`'s `dms`
and `presence` exports.

**`Message.component.tsx`** — union. This branch's emoji imports and font-size constants sit beside
`develop`'s copy-on-long-press constants, and the two `./Message.utils` imports become one.

**`Message.utils` — a real collision.** This branch added `Message.utils.tsx` and `develop` added
`Message.utils.ts`, and both resolve as `./Message.utils`. `develop`'s markdown-source helpers are
folded into the `.tsx`, which has to stay `.tsx` because `renderWithInlineEmoji` returns JSX, and
the `.ts` is removed. Every existing importer keeps working unchanged:
`Message.blankLines.test.tsx` still gets `normalizeMessageWhitespace` and `toMarkdownSource`, and
`Message.component.tsx` gets all four.

**`Message.test.tsx` — add/add, and not concatenable.** Each file installs its own module-scope
`jest.mock` of `@ronradtke/react-native-markdown-display`, one stubbing the text and paragraph
rules and the other the link rule, so putting them in one file would silently disable half the
coverage. `develop`'s file keeps the canonical name and this branch's three cases move to
`Message.emoji.test.tsx`, which is the split the folder already uses for `Message.blankLines`,
`Message.downloads` and `Message.status`. No case is dropped from either side.

## Gate

|---|---|---|
| mobile | clean | 98 suites, 421 passed |
| common | clean | 9 suites, 110 passed |
| desktop | clean | 129 of 129 suites, 598 passed |

PR #3493, opened 2026-09-11
Keep the desktop app resident while the backend is running

### LINKED ISSUE #53: Desktop app should try to prevent itself from being suspended or sleep / powersave turning on.
Sometimes we have a problem with registration where the registrar user has the app open and their computer powered on and connected, but where their computer has gone on suspend. 

I think this is easy to fix because Electron lets us request that the OS not suspend when the app is running. https://www.electronjs.org/docs/latest/api/power-save-blocker

We should do this. 

We should confirm that it does not prevent sleep/suspend over a user action requesting sleep/suspend, e.g. when the laptop lid is closed.

holmesworcester:
Fixes #53

## What

Holds one `powerSaveBlocker('prevent-app-suspension')` for exactly the backend's lifetime: started right after the fork, released first thing in before-quit; both helpers idempotent because before-quit fires twice on the wait-for-backend path. Explicit user sleep (lid close) still wins — cited from Apple IOPMLib and Microsoft SetThreadExecutionState docs, since Electron's page is silent on it. No opt-out shipped: battery trade-off stated.

## Evidence

4 new main-process tests, all red on unmodified main.ts, green after; suite 268→272 passing with the same single pre-existing failure. Lead verified placement and re-ran main.test.ts (20/20).

## Reviewer notes

- Full report with suite logs: local review bundle `lhf/review/issue-53.html`.

### DIFF
--- a/packages/desktop/src/main/main.test.ts
+++ b/packages/desktop/src/main/main.test.ts
@@ -4,7 +4,7 @@ import * as main from './main'
 import * as backendHelpers from './backendHelpers'
 
 import { autoUpdater } from 'electron-updater'
-import { BrowserWindow, app, ipcMain, Menu } from 'electron'
+import { BrowserWindow, app, ipcMain, Menu, powerSaveBlocker } from 'electron'
 import { waitFor } from '@testing-library/dom'
 import path from 'path'
 import { composeInvitationDeepUrl, getValidInvitationUrlTestData, validInvitationCodeTestData } from '@quiet/common'
@@ -65,6 +65,9 @@ jest.mock('child_process', () => {
 })
 
 jest.mock('electron', () => {
+  // Faithful little fake: real Electron hands back an id and tracks whether that id is still active.
+  const startedPowerSaveBlockers = new Set<number>()
+  let nextPowerSaveBlockerId = 1
   return {
     ...jest.requireActual('electron'),
     app: {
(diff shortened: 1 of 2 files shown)


holmesworcester:

Two files conflicted, both unions.

- **`main.ts`** — `develop`'s import order wins, which puts `./appImageEnvironment` first so the
  host-child environment is cleaned before anything can spawn a process. `powerSaveBlocker` is
  re-applied to the electron import.
- **`main.test.ts`** — the electron import takes both `powerSaveBlocker` and `session`, and the
  mock object keeps this branch's `powerSaveBlocker` fake beside `develop`'s `session` and
  `systemPreferences` fakes, which came in with #3516's camera permission work.

## Gate

|---|---|---|
| desktop | clean | **129 of 129 suites, 590 passed** |

This is the first fully green desktop run in the train. Everything before it reported
`src/main/main.test.ts` as a known failure, because that suite cannot resolve `backend-bundle`
in a worktree that has not built the backend. It does `jest.mock('backend-bundle', jest.fn())`,
so only module resolution matters, and a stub at the gitignored `packages/backend-bundle/bundle.cjs`
is enough to let it load. Since this PR's four new tests live in that very suite, gating it any
other way would have proved nothing. The stub is local to the worktree and is not part of this
change.

PR #3494, opened 2026-09-11
fix(desktop): persist and restore the main window size (#2103)

### LINKED ISSUE #2103: Persist windows size in desktop application
The window size of the electron app is hard-coded.
image

We should configure the application in such a way that, upon resizing the application, its dimensions are persisted locally. When booted again, the application should check for this persisted dimension and set it accordingly.

holmesworcester:
Fixes #2103

## What

Size saved to its own electron-store file (window-state.json) from a debounced resize handler and flushed on close; restored in createWindow and clamped to the current display's work area (600x400 floor mirrors setMinimumSize). Skips maximised/fullscreen sizes; malformed data → defaults. Position deliberately not persisted.

## Evidence

8 new tests written first (4 red on unmodified code), all green after; full desktop suite 276/284 with the same single pre-existing failure. Lead verified flush runs inside the 'close' handler before hide/destroy and re-ran main.test.ts (24/24).

## Reviewer notes

- Full report with suite logs: local review bundle `lhf/review/issue-2103.html`.

### DIFF
--- a/packages/desktop/src/main/main.test.ts
+++ b/packages/desktop/src/main/main.test.ts
@@ -25,6 +25,11 @@ const mockWindowClose = jest.fn()
 const mockSetMovable = jest.fn()
 const mockSetAlwaysOnTop = jest.fn()
 const mockIsDestroyed = jest.fn()
+const mockIsMaximized = jest.fn()
+const mockIsFullScreen = jest.fn()
+const mockGetPrimaryDisplay = jest.fn()
+const mockStoreGet = jest.fn()
+const mockStoreSet = jest.fn()
 
 const spyCreateWindow = jest.spyOn(main, 'createWindow')
 const spyGetPorts = jest.spyOn(backendHelpers, 'getPorts')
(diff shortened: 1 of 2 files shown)


holmesworcester:

One file conflicted, `main.ts`, in two places, and both are imports.

- `develop`'s import order wins, which puts `./appImageEnvironment` first so the host-child
  environment is cleaned before anything can spawn a process. This branch's `screen` joins the
  electron import beside `develop`'s `powerSaveBlocker`, which landed with #3493 a moment ago.
- This branch's `electron-store` type import is kept beside `develop`'s external-link,
  camera-permission and e2e captcha imports.

Nothing in the window-state logic itself conflicted.

## Gate

|---|---|---|
| desktop | clean | 129 of 129 suites, 598 passed |

`src/main/main.test.ts`, where this PR's eight tests live, runs here because the worktree carries
a stub at the gitignored `packages/backend-bundle/bundle.cjs`. That suite does
`jest.mock('backend-bundle', jest.fn())`, so only module resolution matters. The stub is local and
is not part of this change.

PR #3499, opened 2026-09-11
fix(desktop): close Settings when a background action switches channel (#1349)

### LINKED ISSUE #1349: Settings modal allows background actions
Version: 1.1.0-alpha.6

Currently we have some actions that can be perform in background even if Settings modal is open. Most important things like typing and sending messages are blocked. Others are not blocked (and probably shouldn't be) but should trigger closing Settings and they are not:
1. Use CTRL+k and switch channel in background - user switch channel but can't see it as Settings are still open.
2. If there is new message popup and the user click it it moves him to the channel in background - but as above, he can't see it because Settings are still open.

Both actions doesn't sound very bad but while we perform them "Settings" should close. In both cases user can expect that if he choose to go to the message in specific channel (point 2) or specific channel itself (point 1) he should be moved to the channel window. He actually is but can be unaware of it as Settings modal is still open.

holmesworcester:
Fixes #1349

## What

Both triggers (Ctrl+K search pick, notification click) converge on publicChannels.actions.setCurrentChannel, so one root-saga takeEvery covers both: a 20-line desktop saga that closes only accountSettingsModal when it is open. No state-manager change, no effect on other modals.

## Evidence

4 rtl-tests driving the real root saga via prepareStore: both paths assert the channel actually changed before asserting Settings closed; startup guard + 'does not close unrelated modals'. 3 of 4 fail with the registration reverted. Full desktop 96/97 with the same pre-existing failure. Lead re-ran 4/4 and read the saga.

## Reviewer notes

- Full report with suite logs: local review bundle `lhf/review/issue-1349.html`.

### DIFF
--- a/packages/desktop/src/renderer/sagas/index.saga.ts
+++ b/packages/desktop/src/renderer/sagas/index.saga.ts
@@ -1,6 +1,7 @@
-import { communities } from '@quiet/state-manager'
+import { communities, publicChannels } from '@quiet/state-manager'
 import { all, takeEvery } from 'redux-saga/effects'
 import { customProtocolSaga } from './invitation/customProtocol.saga'
+import { closeSettingsOnChannelSwitchSaga } from './modals/modals.saga'
 import { startConnectionSaga } from './socket/socket.saga'
 import { socketActions } from './socket/socket.slice'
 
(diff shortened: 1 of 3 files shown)


holmesworcester:
Merge note: one conflict, `packages/desktop/src/renderer/sagas/index.saga.ts`. develop had switched `customProtocolSaga` to `takeLeading` and forked `watchDeviceLinkExpirySaga`; resolved as a union so both survive alongside this PR's `setCurrentChannel` `takeEvery`. Separately, develop added `index.saga.test.ts`, which asserts the root saga's complete effect list, so the new effect had to be named there (that commit is in the squash body; the test's subject is unchanged). Post-merge desktop: 112/113 suites, 400 tests, 142/142 snapshots.

Full rationale is in merge commit `15cbcbd7d` on `fix/1349-settings-close-on-nav` (GitHub's squash drops merge-commit messages).

PR #3500, opened 2026-09-11
fix(common): decode percent-encoded file names for attachment previews

### LINKED ISSUE #1701: filenames on mobile shouldn't use %20 for spaces
Right now filenames for attached files on mobile use "%20" in the place of spaces. They should use hyphens or spaces instead, and it should be consistent with desktop. 

Screenshot: https://www.loom.com/i/cf3541a39cfe485285a42e6b5995d6f3

holmesworcester:
Fixes #1701

## What

Decoded at the single shared derivation point (packages/common fileData): percent-decode the picked file's basename with a try/catch fallback for names that are not valid encoding, then basename() again so a decoded %2F can never smuggle a path separator into a name the backend interpolates into a path. The send saga already decodes the path, so the displayed name now matches the file actually attached. Not decoded at render time (old messages keep %20; avoids double-decoding).

## Evidence

common 13/13 (incl. 100%25.txt→100%.txt, malformed unchanged, desktop path with spaces unchanged, literal % unchanged) and mobile render test that reproduces the screenshot with My%20File; zero snapshots updated. Lead re-ran both. Known edge: a desktop file whose real name contains a valid escape like a%20b.txt would display as 'a b.txt' — accepted. packages/common must be rebuilt after checkout.

## Reviewer notes

- Touches `packages/common`: run `npm run build` there before running desktop/mobile suites.
- Full report with suite logs: local review bundle `lhf/review/issue-1701.html`.

### DIFF
--- a/packages/common/src/fileData.test.ts
+++ b/packages/common/src/fileData.test.ts
@@ -1,4 +1,4 @@
-import { getFilesData } from './fileData'
+import { decodeFileName, getFilesData } from './fileData'
 
 describe('fileData ', () => {
   it('should return proper FilePreviewData', () => {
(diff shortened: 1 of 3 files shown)


holmesworcester:
Merge note: one conflict at the top of `packages/common/src/fileData.ts`. develop had added its `nextPreviewId` counter where this PR added `decodeFileName`; resolved as a union, leaving develop's `preview_<pid>_<ts>_<n>` id scheme untouched. Post-merge common 9/9 suites / 109 tests, mobile 81/84 suites / 307 passed. Desktop was also run by hand (111/112, identical to baseline) because CI's desktop job self-skips unless `packages/desktop/**` changed, and this PR changes shared `@quiet/common`.

Full rationale is in merge commit `91bd24e3d` on `fix/1701-decode-filenames` (GitHub's squash drops merge-commit messages).

Issue #1063, opened 2022-11-21
Switching from channel list to channel should be super fast


holmesworcester:
In Signal and WhatsApp, switching from the conversation list to the message list is super fast. Is there a way we can make this feel as fast. For example, can we have a screen or two of messages loaded in state on the channel list already, and then fill in the rest of the channel list with a lazy load as soon as the user clicks the channel. 

What do other React Native messaging apps (like Mattermost) do here? Discord is in React Native too. What can we learn about how they do it?

PR #3502, opened 2026-09-11
fix(mobile): stop list rows dimming while the list is being dragged

### LINKED ISSUE #1495: User should see indication that they have tapped
Right now there are no tapped states on iOS or Android for channels and menu items (we used to have them on Android for channels) so users can't tell when they've tapped the screen. 

I think we should be using TouchableOpacity on channels and menu items in action sheets (like settings, delete, etc.)  And TouchableHighlight on buttons. See: https://reactnative.dev/docs/handling-touches

holmesworcester:
Fixes #1495

## What

Tap feedback itself already shipped in PR #1584; what remained were the maintainer's two follow-ups. Fix: `delayPressIn={150}` (the reporter's own number, one shared documented constant) on the three touchables that live inside FlatLists — channel tile, action-sheet menu item, member row — so a drag/flick claims the responder before the row ever dims. The second requirement (highlight lasts only for the tap) needs no code: nothing sets delayPressOut and TouchableOpacity has minPressDuration 0; tests pin that absence. Agent dropped the inherited activeOpacity restyle as scope creep (issue #1063 plans to delete the effect).

## Evidence

3 prop-pinning tests fail on unmodified code (Received: undefined); zero snapshots change because delayPressIn never reaches the host view. Full mobile 54/54, 153 tests. Lead re-ran 26/26. MOTION: three scrollable stories added (ChannelList, ContextMenu, UpdateChannelMembership) for the on-device check; report contains the Pressability state-machine argument. Flagged, not fixed: Button still uses TouchableWithoutFeedback (no feedback at all); taps under 150 ms dim less.

## Reviewer notes

- Full report with suite logs: local review bundle `lhf/review/issue-1495.html`.

### DIFF
--- a/packages/mobile/src/components/ChannelList/ChannelList.stories.tsx
+++ b/packages/mobile/src/components/ChannelList/ChannelList.stories.tsx
@@ -78,6 +78,28 @@ storiesOf('ChannelList', module)
       ]}
     />
   ))
+  // Issue #1495: a list long enough to scroll, so the tap highlight can be
+  // checked against a drag. Flick through it - no tile should dim - then tap
+  // one and the dim should appear and clear with the tap.
+  .add('Long list (scroll to check tap feedback)', () => (
+    <ChannelList
+      // @ts-ignore
+      community={{
+        name: 'Quiet',
+      }}
+      tiles={Array.from({ length: 30 }, (_, index) => ({
+        name: `channel-${index + 1}`,
+        id: `channel-${index + 1}`,
+        message: 'Text from latest chat message. Lorem ipsum dolor sit amet, consectetur adipiscing elit.',
+        date: '1:55pm',
+        unread: index % 4 === 0,
+        isPublic: index % 5 !== 0,
+        redirect: (id: string) => {
+          logger.info(`Clicked ${id}`)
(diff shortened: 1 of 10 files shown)


holmesworcester:
Superseded by a forward-port: https://github.com/TryQuiet/quiet/pull/3653.

Two of the three touchables this PR fixed are still there and still needed the fix — the context-menu row and the channel-membership row — but `ChannelTile` is gone: https://github.com/TryQuiet/quiet/pull/3529 replaced the mobile channel list with the design library's Community home and deleted it. On the new screen the channel row, the Add members row, the member row and the create-channel plus are all `Pressable`s inside one scroll view, and all four lit up on contact.

The forward-port re-applies the intent to the current components, with the same 150ms from the maintainer's comment above, spelling it `unstable_pressDelay` on the `Pressable`s and `delayPressIn` on the two `TouchableOpacity` rows. The shared constant and the "released inside the delay still fires onPress" guard both carry over.

The tests changed shape: the flash is transient, so each one now asserts *between* the finger landing and the list claiming the touch, and 8 of them are red with the component change reverted. Closing this one unmerged.

PR #3503, opened 2026-09-11
fix(mobile): let users search the channel list by name

### LINKED ISSUE #773: User should be able to search channels by channel name
Designs: https://www.figma.com/file/y8h6w8PYR9jyI3zjYHL9Cl/Mobile?node-id=456%3A18822

User should be able to search the channel list by channel name.

Acceptance:
1. User enters a string and all / only all channels with that string display below
2. User can click on channel item to go to that channel view

Non-requirements:
* Full text search of messages
* Searching users

holmesworcester:
Fixes #773

## What

A 'Search channels' Input above the channel FlatList; case-insensitive substring filter on the trimmed query (plain containment is exactly the acceptance criterion, so fuse.js — already a dep — was rejected as needing an unspecified threshold); 'No channels found' empty state; keyboardShouldPersistTaps so a result can be tapped without dismissing first. State lives in the component so the story and RNTL tests exercise it without a store. Empty query renders byte-identically to before; spinner condition still uses the unfiltered list.

## Evidence

6 of 7 new tests red with the component reverted (the 7th is the no-regression guard); one inline snapshot regenerated. Full mobile 147→153 tests, zero failures, verified independently by a second agent and re-run by the lead (53/53). Flags: the shared Input hardcodes an inner testID 'input', so the list now carries one under the composer's screen — same shape as CreateCommunity→CreateUsername which is green in detox CI, but only a device run settles it; the storybook baseline channellist--default.png is stale; Figma frame not accessible, so inline-field vs modal is unconfirmed against design.

## Reviewer notes

- Full report with suite logs: local review bundle `lhf/review/issue-773.html`.

### DIFF
--- a/packages/mobile/src/components/ChannelList/ChannelList.component.tsx
+++ b/packages/mobile/src/components/ChannelList/ChannelList.component.tsx
@@ -1,33 +1,75 @@
-import React, { FC } from 'react'
+import React, { FC, useState } from 'react'
 import { FlatList, View } from 'react-native'
 import { defaultTheme } from '../../styles/themes/default.theme'
 import { Appbar } from '../Appbar/Appbar.component'
 import { ChannelListProps } from './ChannelList.types'
 import { ChannelTile } from '../ChannelTile/ChannelTile.component'
+import { Input } from '../Input/Input.component'
 import { Spinner } from '../Spinner/Spinner.component'
+import { Typography } from '../Typography/Typography.component'
 import { capitalizeFirstLetter } from '@quiet/common'
 
 export const ChannelList: FC<ChannelListProps> = ({ community, tiles, communityContextMenu }) => {
+  const [query, setQuery] = useState('')
+
   let communityName = '...'
   if (community?.name) {
     communityName = capitalizeFirstLetter(community.name)
   }
+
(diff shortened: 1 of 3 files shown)


holmesworcester:
Closing unmerged: the design of record for this screen has no channel search, so there is nothing to forward-port this onto.

`ChannelList.component.tsx`, the file this PR adds the search field to, was deleted by `3628b9940` (https://github.com/TryQuiet/quiet/pull/3529). Its replacement is the design library's Community home. The designer's note beside that frame — Quiet Design Library `0j7Nna9zWmfOSNmRmQK1Uh`, page *Structure & Nav*, note `6220:24045` "Requirements from Holmes" for the **For V1 2025** variant — opens with:

> Top bar without search

and `CommunityHome/CommunityAppbar.component.tsx` follows it: the title bar's right-hand zone, where the prototype put a search glyph, is empty. Re-applying this PR would mean inventing a search affordance the current design does not draw.

https://github.com/TryQuiet/quiet/issues/773 stays open as the feature request. Before it can be built it needs a design decision about where search lives on the new Community home — the old Figma frame it links (`y8h6w8PYR9jyI3zjYHL9Cl` Mobile `456:18822`) predates the library.

PR #3506, opened 2026-09-11
fix(mobile): let the message entry field scale back down

### LINKED ISSUE #2655: Message entry field doesn't scale back down
Steps to reproduce on Android:

1. Send a message with many lines

Expected: the message entry field scales back down to a normal size after scaling up to fit the large message. 

Actual: the message entry field remains awkwardly large.

holmesworcester:
Fixes #2655

## What

Root cause: Input fed onContentSizeChange's height straight back into the same TextInput's style height. On Android (old architecture) a definite height makes Yoga skip the text measure function, the view is never re-laid-out, and the content-size watcher's height falls back to the frame JS pinned — so no smaller size can ever be reported. Fix: multiline inputs get min-height instead of height; wrapper maths (54 floor, height+20) unchanged; single-line inputs byte-identical; iOS unaffected because it measures with unbounded height.

## Evidence

Second agent verified the mechanism against RN 0.77.3 sources (Yoga CalculateLayout.cpp:301, ReactEditText.java:246/1346, ReactTextInputManager.java:1211, RCTBaseTextInputShadowView.mm:84) and flagged one incomplete clause in the commit message (onLayout is not the only driver; the frame fallback is the sharper reason) — conclusion unaffected. Reverting Input.styles.ts fails 2 of 3 new tests ('Expected: 54, Received: 220'). Full mobile 53/53, 150 tests; one Chat snapshot line changed. Pre-existing and untouched: no maxHeight on the composer, so it can still grow unbounded upward.

## Reviewer notes

- Full report with suite logs: local review bundle `lhf/review/issue-2655.html`.

### DIFF
--- a/packages/mobile/src/components/Chat/__tests__/__snapshots__/Chat.test.tsx.snap
+++ b/packages/mobile/src/components/Chat/__tests__/__snapshots__/Chat.test.tsx.snap
@@ -3036,7 +3036,7 @@ exports[`Chat component renders component 1`] = `
                       style={
                         [
                           {
-                            "height": 54,
+                            "minHeight": 40,
                             "paddingBottom": 0,
                             "paddingTop": 12,
                             "textAlignVertical": "center",
--- a/packages/mobile/src/components/Input/Input.styles.ts
+++ b/packages/mobile/src/components/Input/Input.styles.ts
@@ -8,7 +8,21 @@ export const StyledTextInput = styled(TextInput)<{
 }>`
   ${({ height, multiline }) => css`
     text-align-vertical: center;
-    height: ${Math.max(40, height)}px;
+    /*
+     * A multiline input must keep an auto height. Its own box is what React Native measures to
+     * produce onContentSizeChange, so pinning that box to the last reported height freezes the
+     * measurement: on Android a definite height stops Yoga consulting
+     * ReactTextInputShadowNode.measure(), the view is never re-laid-out, ReactEditText.onLayout()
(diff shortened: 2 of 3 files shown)


holmesworcester:
Correction to this PR's squash body on `develop`: the line "Single-line inputs still pin their height and render byte-identically" was true against this PR's original base (`1ed08d8fb`) but is not true of what landed. develop's "Input 2.0" work had already moved single-line height off `StyledTextInput` onto `StyledWrapper` (`INPUT_HEIGHT` 48), so there was no single-line pin left here to preserve.

What actually changed on merge: only the multiline branch, `height: ${Math.max(40, height)}px` -> `min-height: 40px`. develop's `flex: 1`, wrapper height and the `COMPOSE_MIN_HEIGHT` 54 / `height + 20` maths are untouched. The rendered change is one snapshot line in `Chat.test.tsx.snap`: `"height": 54` -> `"minHeight": 40`.

Consequently the third test, which asserted single-line inputs pin height 54 on the field, described develop's old layout and failed regardless of this fix. It was re-expressed against the current design (single-line pins neither `height` nor `minHeight`; the wrapper carries `INPUT_HEIGHT`). The two real regression cases are unchanged, and develop's pre-existing single-line inline snapshot in the same file still passes - that is the evidence single-line rendering is untouched.

Full conflict rationale is in the merge commit `253f26976` on `fix/2655-input-shrink` (GitHub's squash drops merge-commit messages).

PR #3507, opened 2026-09-11
Link #channel mentions to the channel they name (desktop)

### LINKED ISSUE #1745: Users should be able to tap/click #some-channel to go to that channel
If someone mentions a public channel, users should be able to click on the channel name to go to that channel.

(I'm not sure how this should work for private channels and users that don't have access to them, but we can look at how Slack does it.)

holmesworcester:
Fixes #1745

## What

A 150-line remark plugin walks the parsed mdast and turns `#name` text tokens that resolve to a channel the user can see into link nodes, so markdown's own rules hold for free: code spans/fences, existing links and GFM autolinked URLs (example.com/#general) are never touched, and a word boundary is required before `#` (foo#bar stays text). Trailing sentence punctuation is trimmed (#general. links 'general'). Href is `#channel/<id>` because react-markdown's uriTransformer rewrites custom schemes to javascript:void. One optional `channelLinks` prop travels the same path as openUrl (Channel → ChannelComponent → ChannelMessages → BasicMessage → TextMessage; one line per hop); click dispatches setCurrentChannel. Unknown/private channels stay plain text.

## Evidence

Cypress before/after of a message with #general, #nope, `#general` in code and a URL fragment; 16 RTL tests (5 red on unmodified code; the invisible-channel guard now also asserts openUrl is never called). Full desktop 96/97, zero snapshots changed; typecheck + eslint clean on 9 files. Lead read the plugin and re-ran 20/21 with the known pre-existing failure. Largest footprint in the batch (9 files, +382) — a feature, not a bug fix.

### Screenshots

**after**

**before**

## Reviewer notes

- Full report with suite logs: local review bundle `lhf/review/issue-1745.html`.

### DIFF
--- a/packages/desktop/src/renderer/components/Channel/Channel.tsx
+++ b/packages/desktop/src/renderer/components/Channel/Channel.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect } from 'react'
+import React, { useCallback, useEffect, useMemo } from 'react'
 
 import { shell, ipcRenderer, webUtils } from 'electron'
 
@@ -7,6 +7,7 @@ import { users, messages, publicChannels, communities, files, network, settings 
 import { FileMetadata, CancelDownload, FileContent, FilePreviewData } from '@quiet/types'
 
 import ChannelComponent, { ChannelComponentProps } from './ChannelComponent'
+import { ChannelLinkNavigation } from '../widgets/channels/TextMessage'
 
 import { useModal } from '../../containers/hooks'
 import { ModalName } from '../../sagas/modals/modals.types'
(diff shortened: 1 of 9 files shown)


holmesworcester:

`develop` has moved a long way since this was cut at `1ed08d8fb`, most of it the onboarding
redesign. Two files conflicted, both trivially.

- **`Channel.tsx`** — union of the two React imports: `useMemo` from here, `useState` from
  `develop`.
- **`BasicMessage.tsx`** — `develop` now filters `MessageType.Empty` out before mapping. Its
  version is kept and this branch's `channelLinks` prop is re-applied to the
  `NestedMessageContent` inside that filter.

## One integration fix

`visibleChannels` is read from `publicChannels.selectors.publicChannels`, which yields nothing
until the community's channels are in the store. The composer renders before that, and `develop`'s
newer composer tests exercise exactly that state, so this branch's `channelLinks` memo, which
called `.map` on it directly, threw in three of them. The memo now treats the selector's result as
possibly empty. No behaviour changes when the selector has a list; the difference is only that
rendering before it does no longer throws.

|---|---|---|
| desktop | clean | 128 of 129 suites, 569 passed, no snapshots changed |
| common | clean | 9 suites, 109 passed |

PR #3508, opened 2026-09-11
fix: drop whitespace-only text messages before sending (#2778)

### LINKED ISSUE #2778: The first message after creating a channel can be sent empty
Version: Version: 4.0.3
System: all desktops, iOS, android

Usually, users can't send empty messages. There is one exception:
1. Create a new channel.
2. Enter space as the only element in the text area.
3. Send a message.
4. A message appears in the channel without any text or sign. 

This will not happen in any other circumstance. 

Image

holmesworcester:
Fixes #2778

## What

No first-message code path exists: both screens guarded with a bare truthiness check a single space satisfies, and the shared saga broadcast whatever it got. 'First message only' is VISIBILITY — messages merge by sender within 300 s and only a group's first message draws avatar/nickname/timestamp; a fresh channel forces the first user message to open a group, so the blank one renders fully dressed. Fix: drop whitespace-only Basic messages without media in the shared sendMessage saga (desktop + mobile identical), tighten both UI guards to match, never trim content (leading spaces are markdown-significant), info/image/file messages exempt.

## Evidence

2 saga tests + desktop rtl test red on unmodified code (the store literally receives message ' '); mobile test is a regression guard only — mobile already trims for an iOS autocorrect workaround, which is exactly why the fix belongs in the shared saga (stated plainly). state-manager 201/201, desktop 270 passed with the known failure, mobile 149/149. Lead re-ran 8/8, 18/20 (2 skipped), 2/2 and confirmed the rebuilt lib carries the guard. Reviewers must `npm run build` in state-manager first.

## Reviewer notes

- Touches `packages/state-manager`: run `npm run build` there before running desktop/mobile suites.
- Full report with suite logs: local review bundle `lhf/review/issue-2778.html`.

### DIFF
--- a/packages/desktop/src/renderer/components/Channel/Channel.tsx
+++ b/packages/desktop/src/renderer/components/Channel/Channel.tsx
@@ -67,8 +67,9 @@ const Channel = () => {
 
   const onInputEnter = useCallback(
     (message: string) => {
-      // Send message out of input value
-      if (message) {
+      // Send message out of input value. Whitespace-only input has nothing to
+      // send, but any attached files below still go out.
+      if (message.trim()) {
         dispatch(messages.actions.sendMessage({ message }))
       }
       // Upload files, then send corresponding message (contaning cid) for each of them
--- a/packages/desktop/src/rtl-tests/channel.main.test.tsx
+++ b/packages/desktop/src/rtl-tests/channel.main.test.tsx
@@ -464,6 +464,102 @@ describe('PublicChannel', () => {
     expect(await screen.findByText(messageText)).not.toHaveStyle('color:#B2B2B2')
   })
 
+  it('does not send a whitespace-only message as the first message in a new channel', async () => {
+    const { store } = await prepareStore(
+      {},
+      socket // Fork state manager's sagas
+    )
+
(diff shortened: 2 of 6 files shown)


holmesworcester:
Merge note: three conflicts, all the same shape - develop made the destination channel explicit while this PR made blankness a reason not to send, and both inserted at the same spot. Resolved by keeping develop's guard first in each: the saga keeps its missing-`channelId` early return, and `Channel.tsx` / `Channel.screen.tsx` keep their guards and `channelId` payloads with `if (message)` becoming `if (message.trim())`. `sendMessage.saga.test.ts` auto-merged but did not compile - develop made `WriteMessagePayload.channelId` required - so a real `channelId` was added to the four new cases. That is load-bearing: without one, the two "drops a ... message" cases would have passed on the channelId guard without reaching the blank-message check. Verified by negative control (guard removed -> exactly those two fail, other eight pass). Post-merge: state-manager 66/66 / 279 tests, mobile 82/85 / 308 passed, desktop 111/112 / 398 tests.

Full rationale is in merge commit `866b3ad99` on `fix/2778-empty-first-message` (GitHub's squash drops merge-commit messages).

PR #3509, opened 2026-09-11
Say invite link, not invitation code, in join errors (#2441)

### LINKED ISSUE #2441: Incorrect text ("invite code") on join screen
Right now the join screen says "invite code" instead of "invite link", which is confusing to users because we don't offer invite codes anymore. This just came up in a demo and it's a tiny change, so I'm adding it to the next sprint. 

It should say "invite link" instead of "invite code", both in the text above the from field and in the placeholder text in the field itself.

Here's the screen:

Image

holmesworcester:
Fixes #2441

## What

Both remaining user-facing strings now read 'Please check your invite link and try again' (desktop fieldsErrors enum + mobile JoinCommunity). Added the only regression test either platform has for this wording. Flagged a near-duplicate 'invitation link' string in common/const.ts as a follow-up, deliberately untouched.

## Evidence

Cypress before/after of the desktop join screen (the 'before' shows label 'Paste your invite link' above error 'invitation code'). Desktop 95/96 identical before/after; lead re-ran the mobile test after fixing a worktree snapshot-path artifact.

### Screenshots

**desktop-after**

**desktop-before**

## Reviewer notes

- Full report with suite logs: local review bundle `lhf/review/issue-2441.html`.

### DIFF
--- a/packages/desktop/src/renderer/forms/fieldsErrors.ts
+++ b/packages/desktop/src/renderer/forms/fieldsErrors.ts
@@ -13,7 +13,7 @@ export enum CommunityNameErrors {
 }
 
 export enum InviteLinkErrors {
-  InvalidCode = 'Please check your invitation code and try again',
+  InvalidCode = 'Please check your invite link and try again',
 }
 
 export enum ChannelNameErrors {
--- a/packages/mobile/src/components/JoinCommunity/JoinCommunity.component.tsx
+++ b/packages/mobile/src/components/JoinCommunity/JoinCommunity.component.tsx
@@ -54,7 +54,7 @@ export const JoinCommunity: FC<JoinCommunityProps> = ({
 
     if (!submitValue) {
       setLoading(false)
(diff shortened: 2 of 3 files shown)


holmesworcester:
Heads-up for the record: this renamed `InviteLinkErrors.InvalidCode` and updated the mobile component and its test, but `packages/e2e-tests` still asserted the old "invitation code" wording. `admissionTimeout.test.ts` therefore went red on every develop-based branch about an hour later, failing the `Run admission timeout recovery test` step with `NoSuchElementError` on the case `keeps the guest on Join Community when an invalid invitation is submitted`.

Fixed in `2adde19d7`, merged as `cbcc7afcd` via https://github.com/TryQuiet/quiet/pull/3591.

Worth remembering: a user-facing copy rename needs a grep of `packages/e2e-tests` too, since those assertions are plain literals (e2e-tests cannot import the desktop renderer).

PR #3516, opened 2026-09-13
onboarding: scan QR codes on desktop (camera permission via Electron)

holmesworcester:
Desktop "Join with QR code" and Link devices → "Scan QR code" rendered a link/paste field. They now open the camera and scan the code, per the prototype's sheets (`2811:2460`, `2811:2587`). A scanned code takes exactly the path a pasted link takes.

## What changed

**Camera permission through Electron** (`packages/desktop/src/main/cameraPermission.ts`, wired in `main.ts` once the window exists)
- `session.defaultSession.setPermissionRequestHandler` + `setPermissionCheckHandler`: `media` is granted only to the main window's own renderer showing the app's `index.html` (recognised by file path — a `file:` origin is opaque) and only for **video**; audio is refused. Every other permission keeps Electron's default grant, which the renderer's `new Notification(...)` relies on. Nothing else about the session (or the captcha window's own CSP/handler) changes.
- The renderer asks main first over IPC (`camera:request-access`), then calls `getUserMedia`.
- E2E only: with `IS_E2E=true` and `E2E_FAKE_CAMERA_FILE=<clip.y4m>` the app appends Chromium's `use-fake-device-for-media-stream`, `use-fake-ui-for-media-stream`, `use-file-for-fake-video-capture` switches before `ready`.

**Permission flow per OS**
| OS | before getUserMedia | then |
|---|---|---|
| macOS | `systemPreferences.askForMediaAccess('camera')` — the system prompt, once; later calls return the stored answer | request handler grants video to the app renderer; a denial shows the *denied* state |
| Windows | `systemPreferences.getMediaAccessStatus('camera')` — a `denied` privacy setting shows the *denied* state without prompting | same |
| Linux | nothing to ask | `getUserMedia` decides: `NotAllowedError` → *denied*, `NotFoundError` / `NotReadableError` → *no camera* |

**Scanner** (`packages/desktop/src/renderer/components/Onboarding/qrScanner/`)
- `useQrScanner`: `getUserMedia({ video: { facingMode: 'environment' }, audio: false })` into a `<video>`, a frame drawn to a canvas and decoded with **jsQR** every 100 ms (~10 fps); the stream is released when a code is accepted or on unmount (the container closes the modal in the same tick a code is accepted, so the released camera — Chromium paints an ended track black — is not a state the user sees; the decoded stories show it).
- `decodeQr.ts`: `decodeQrImage` (jsQR) and `parseScannedCode` — `getInvitationCodes` from `@quiet/state-manager`, the paste field's parser, so member and device links are validated identically.
- `QrScannerComponent`: the camera fills the 375 column edge to edge (375×384, flush with the title bar when there is no copy above it), the framed square in the middle; a code that is not a Quiet invitation shows the paste field's own error (`Please check your invitation code and try again`) and scanning continues.

**Wiring** — `JoinCommunity.tsx`: options → *Join with QR code* → scanner; `LinkDevices.tsx`: *Scan QR code* → scanner with the sheet copy. A decoded member link dispatches `communities.actions.joinCommunity({ inviteData })` and opens Choose username; a device link dispatches `linkDevice({ inviteData })` — the same handler the paste field calls. When the camera cannot be used, the scanner's *Paste a link* goes to the paste step (`Paste a link to Join`); back returns to the scanner.

**Dependencies** — `jsqr@1.4.0` (desktop dependency; pure JS, no dependencies, 280 KB unpacked / 360 KB on disk) and `qr.js@0.0.0` (devDependency in desktop and e2e-tests; 92 KB; it is `react-qr-code`'s own encoder and was already in desktop's tree transitively) so tests, stories and the e2e fake camera make real QR codes of real links. `npm install` cannot resolve the lerna-linked `@quiet/*` packages from the registry, so the `package.json` / `package-lock.json` entries were written by hand with the registry's `resolved` + `integrity` (verified against `npm pack`) and the tarballs extracted into `node_modules` — the same result `npm install --save-exact` produces.

**Stories** — `Screens/Onboarding`: Join with QR code (scanning), Requesting, Decoded, Invalid, Denied, No camera; Scan QR code, Decoded, Denied; the Walkthrough's QR steps scan too (camera picker in the story chrome). No camera exists under Storybook, so `storyCamera.ts` makes `getUserMedia` return a `canvas.captureStream()` — a real QR code of `validInvitationDatav4[0]` composed by the app's `composeInvitationShareUrl`, a blank feed, or the failure modes — and the real `<video>` → canvas → jsQR pipeline runs. All under `chromatic: { disableSnapshot: true }`.

**E2E** (`packages/e2e-tests`) — `fakeCamera.ts` writes a Y4M clip (4:2:0, 640×480, 5 frames) of a real QR code with `qr.js`; the client under test starts with `E2E_FAKE_CAMERA_FILE`. New scenario *Get started → Join with QR code → scan → username*; the device-linking scenario now scans the device link instead of pasting it. Selectors: `joinWithQrCode` / `scanQrCode` wait for the scanner, `usePasteLinkFromScanner` for the fallback.

**Docs** — `ONBOARDING.md` phase-1 item 4 and the component map say desktop scans.

## The undesigned states

The prototype draws one state per sheet: the camera. *Requesting camera access…*, *Camera access was denied.* and *No camera is available.* have no frame; their copy is the implementation's minimum (`SCANNER_COPY` in `QrScannerComponent.tsx`), each with the Open invite link frame's own *Paste a link* text as the route to the paste step. The stories label them as such. They need a designer's pass.

## Verified

- `npx tsc --noEmit -p tsconfig.json`: desktop clean, e2e-tests clean. Prettier (root 3.9.6) and eslint clean on every changed file.
- Desktop jest, the touched suites (`src/main`, `Onboarding`, `CreateJoinCommunity`): 99 passed, 1 skipped (pre-existing), 0 failed; run 4× in a row. New tests: `cameraPermission.test.ts` 15 (page matching, video-only grant, non-media untouched, per-OS access, fake-camera switches), `decodeQr.test.ts` 9 (real QR images of the fixture links through jsQR and the parser), `QrScannerComponent.test.tsx` 7 (jsdom camera: decode once, invalid keeps scanning, denied / no camera / not readable, intro, release on unmount), `LinkDevices.test.tsx` 3, `JoinCommunity.test.tsx` +3 (scan member → username, scan device → linkDevice, denied → paste → back).
- E2E `onboarding.test.ts` over `LOCAL_TRANSPORT=true` with the AppImage (`distUbuntu` → `linux:copy`, `xvfb-run` + fluxbox): 7 passed, 1 skipped (the pre-existing `it.skip` for device removal), 0 failed — three runs on three successive builds (the last one the rebased tree; the two later comment-only edits do not change the build) — including *Join with QR code → scan → username* (the fake camera plays the owner's real invite link; the scan lands on Choose username and the member joins) and *B links through Link devices → Scan QR code* (scans A's device link).
- Frozen static Storybook (`build-storybook -o …/qr-sb/static`, served on 127.0.0.1:6320) screenshotted through headless Chrome over CDP and looked at: every state renders as intended — the scanning frame, the request text, the decoded stories record the dispatch (joinCommunity / linkDevice with the parsed fixture) and release the camera, the invalid story shows the error under a visible code and keeps scanning, denied / no-camera show the message and *Paste a link*; the walkthrough's Join with QR code lands on Choose username and Scan QR code returns to Get started with linkDevice recorded.

## Not done / notes

- **Mobile untouched.** The mobile branch has no scanner hook to reuse (only `react-native-qrcode-svg` for display; JoinCommunity / LinkDevices are paste and deep link), so a mobile scanner would mean a camera dependency — out of this PR, as ONBOARDING.md already records.
- `design/onboarding-impl` moved while this was built; the branch is rebased onto `f3eecd9a0` (the interaction-states commits; the scanner's *Paste a link* is that commit's `TextLink`) and merges cleanly with the current tip (`8d62aec75`, checked with `git merge-tree`).
- Not touched: #3400's AdmissionCoordinator, admission deadlines, harness caps.


holmesworcester:
## Brought up to date with `design/onboarding-impl` (root `2d4f0c091`)

Head is now `6a942167e`. Base is unchanged (`design/onboarding-impl`); GitHub reports `MERGEABLE` (state `UNSTABLE`, i.e. CI, not conflicts), and `git merge-tree --write-tree origin/design/onboarding-impl origin/design/onboarding-qr-scan` exits 0.

The root brought two things this branch had to meet: **Electron 32 → 44.3.0**, and a **device-link consent sheet**.

### Resolution log

| Path | Resolution |
| --- | --- |
| `src/main/main.ts` | Both sides added an import on the same line; both kept. The camera wiring in `app.on('ready')` and the `applyFakeCameraSwitches` call merged cleanly around develop's Electron 44 changes. |
| `JoinCommunity.tsx` | develop's file with this branch's steps applied: `pasteQrCode` becomes `scanQrCode` (the camera) plus `pasteFromQrCode` (the fallback). develop's join-error field reporting, `leave` handler and consent gate are all kept. |
| `LinkDevices.tsx` | This branch's scanner and paste fallback, on develop's consent gate. |
| `e2e-tests/src/tests/onboarding.test.ts` | The device-link scenario scans rather than pastes, and no longer asserts on a device list. |
| `packages/desktop/package-lock.json` | Two adjacent entries, both kept. |

### Where the root deleted things

develop **dropped the linked-device list** — no backend on this line enumerates a user's devices (https://github.com/TryQuiet/quiet/issues/3636) — and with it the `linkedDevices` selector and the `getLinkedDevices` saga this branch read. So:

- `LinkDevices.tsx` no longer selects `linkedDevices` or dispatches `getLinkedDevices`, and `LinkDevicesComponent` is rendered without the `linkedDevices` prop. The entry screen offers *Display QR code* and *Scan QR code* and nothing enumerates.
- The e2e device-linking scenario dropped its `hasNoLinkedDevices()` assertion.
- `ONBOARDING.md` step 6 still promised the "Linked devices / No linked devices" list. Corrected, with the issue cited.

### Consent now covers the scanned path

develop gates a **pasted** device link behind a consent sheet. A camera decodes whatever is put in front of it, so a decode is no more deliberate than a paste — arguably less. Both containers therefore route a `DeviceInvitationData` through `setPendingDeviceInvite` and only dispatch `linkDevice` from `confirmDeviceLink`, carrying `deviceLinkConsent: true`.

Tests follow that change rather than being relaxed to fit it: `LinkDevices.test.tsx` asserts the consent sheet appears for a **scanned** code, that nothing is dispatched before it is confirmed, and that **declining links nothing**; `JoinCommunity.test.tsx` does the same for a device link scanned there; the e2e `LinkDevicesModal` gained `confirmScannedDeviceLink()`.

The Screens/Onboarding scanner stories recorded `linkDevice({ inviteData })` as what a decoded device link dispatches. That is no longer what either container does, and a design-review surface that states the app's behaviour has to state the behaviour the app has, so the recorded line now names the consent step and the real payload.

### Electron 44 — what was actually exercised

`packages/desktop/node_modules/electron` resolves to **44.3.0** (Chrome 152.0.7977.78). Checked against the **shipped `electron.d.ts`**, not the docs:

- `setPermissionRequestHandler(handler(webContents, permission, callback, details))` — 4-arg shape unchanged; `'media'` is still in the permission union; `details` is `PermissionRequest | FilesystemPermissionRequest | MediaAccessPermissionRequest | OpenExternalPermissionRequest`.
- `MediaAccessPermissionRequest` still has `mediaTypes?: Array<'video' | 'audio'>` and inherits `requestingUrl: string` from `PermissionRequest`.
- `setPermissionCheckHandler(handler(webContents: WebContents | null, permission, requestingOrigin: string, details: PermissionCheckHandlerHandlerDetails))` — `details.mediaType?: 'video' | 'audio' | 'unknown'` and `details.requestingUrl?: string` both still exist.
- `systemPreferences.askForMediaAccess('microphone' | 'camera'): Promise<boolean>` and `getMediaAccessStatus('microphone' | 'camera' | 'screen')` unchanged.
- The main window still runs `nodeIntegration: true, contextIsolation: false`, so the renderer's `ipcRenderer.invoke` path in `renderer/camera.ts` works.

Nothing in this branch's permission code needed changing. **`tsc --noEmit` on `packages/desktop` and `packages/e2e-tests` both exit 0.**

Then it was **run**, not just typechecked. A harness loads this branch's compiled `registerCameraPermissionHandlers` and installs it on a real Electron 44.3.0 default session with the same predicate `main.ts` installs, then calls `getUserMedia` from real windows under `xvfb` with `--use-fake-device-for-media-stream`:

| Check | Result |
| --- | --- |
| The app's renderer, on the app's own `index.html`, gets the camera | granted, track `fake_device_0` live |
| Another window is refused | `NotAllowedError` |
| The app's window on a *different* `file://` page is refused | `NotAllowedError` |
| A repeat `getUserMedia` still succeeds (the **check** handler's path) | granted |
| A non-media permission is still granted by default (the app relies on this for renderer `Notification`) | `Notification.requestPermission()` → `granted` |
| `resolveCameraAccess('linux')` | `granted` |

6/6. The negative controls matter: they show the handler discriminates rather than granting everything.

**One finding worth recording.** Running the same harness *with* `--use-fake-ui-for-media-stream` makes both negative controls pass the camera through — that switch bypasses the permission handler entirely. `applyFakeCameraSwitches` sets **both** switches under `IS_E2E`, so the e2e camera tests prove the scanner decodes, but they do **not** exercise the permission gate. `src/main/cameraPermission.test.ts` is what covers it, and it passes.

### Verified

- Desktop unit tests: **117 of 118 suites pass, 442 tests, 0 failures, 142 snapshots unchanged**. `cameraPermission`, `decodeQr`, `QrScannerComponent`, `LinkDevices` and `JoinCommunity` suites all pass.
- `jsqr@1.4.0` and `qr.js@0.0.0` installed from the registry; both tarball SHA-512s match the lockfile `integrity` entries exactly.
- **Storybook**: static build served on `:8103`, 28 stories captured with Electron 44 under xvfb and looked at individually — every scanner state (scanning, requesting, decoded, invalid, denied, no camera, for both Join with QR code and Scan QR code), plus Link devices, Display QR code, Get started, the walkthrough, Settings and the sidebar. No story errors, no regressions.

### Not verified

- **The packaged app was not launched.** `main.ts` requires `backend-bundle`, whose `bundle.cjs` is a webpack artifact not built in this worktree (the backend has no `node_modules` here), so the real binary could not start. That is why the permission module was exercised directly on Electron 44 instead — which tests the same code, but not the app's startup wiring.
- For the same reason `src/main/main.test.ts` cannot run here; it is the suite that asserts the camera handlers are registered on ready. It fails identically on the root branch.
- No desktop e2e run. The scan scenarios and the new `confirmScannedDeviceLink()` step are matched against what the merged app renders, not proven by a run.
- macOS and Windows branches of `resolveCameraAccess` are covered only by `cameraPermission.test.ts`; this box is Linux.

holmesworcester:
## Electron 44 hardening, from what the runtime probe showed

Head is now `b902d3963`. Two follow-ups to the merge, in one commit.

**1. The check handler now answers when Electron omits `mediaType`.** Running this branch's `registerCameraPermissionHandlers` against the real Electron 44.3.0 binary (Chrome 152.0.7977.78) under xvfb showed the check handler being called for `media` with `details.mediaType` **absent** while `getUserMedia` is in flight, and only carrying `'video'` afterwards:

```
REQ granted=true url=file:///...index.html?dataPort=1234#/ mediaTypes=["video"]
CHK result=false mediaType=undefined   (x2)
RENDERER GUM_OK tracks=1
CHK result=true mediaType="video"
RENDERER PERMQ camera=granted
```

So `details.mediaType === 'video'` answered "no camera here" to a question about the camera. Nothing was broken by it — the request handler is what gates the stream, and `getUserMedia` succeeded — but this handler is what `navigator.permissions.query` and repeat calls read, so it should answer honestly. It now refuses only the microphone: `details.mediaType !== 'audio'`.

The new test in `cameraPermission.test.ts` covers the absent case and `'unknown'`, and re-checks that the two things which must still be refused still are (another renderer, and the microphone). **It fails against the old expression** — verified by reverting the one term and re-running.

**2. Electron 45 is recorded as a thing to revisit, not fixed.** On 44, a screen-capture request arrives as `media` with an empty `mediaTypes`, which `onlyVideo` denies. Electron 45 reports it as `display-capture` instead, which the grant-everything-else default in the request handler would let through silently. The app does not capture the screen, so `display-capture` should be denied explicitly at that upgrade. That is now a comment on the default rather than a surprise later.

### What the Electron 44 evidence was, in full

Everything below is quoted from the runtime and typing checks, not from docs:

- `'media'` is still the camera permission string in 44; the permission union grew from 20/17 values to 44 but purely by addition — nothing renamed or removed.
- `MediaAccessPermissionRequest` (`mediaTypes?: Array<'video'|'audio'>`, `securityOrigin?`) and `PermissionRequest` (`isMainFrame`, `requestingUrl`) are **byte-identical** to v31. `PermissionCheckHandlerHandlerDetails` gained only three `fileSystem` fields; `mediaType?: 'video'|'audio'|'unknown'` is unchanged.
- `askForMediaAccess('microphone'|'camera')` and `getMediaAccessStatus('microphone'|'camera'|'screen')` are character-identical to v31.
- Breaking changes 33→44 touching media: **none**. 36.0's `app.commandLine` lowercasing affects switch *names*, not values — verified on 44.3.0 that `use-file-for-fake-video-capture` keeps an uppercase path verbatim, so the E2E fake-camera switch survives. 39.0 (`NSAudioCaptureUsageDescription`) and 44.0 (`chromeMediaSource: 'desktop'` ids) both concern `desktopCapturer`, which this branch does not use. The `display-capture` change is 45.0, not 44.
- The fake camera still works on Chromium 152: the app read the branch's own Y4M file (centre 255,255,255 / corner 56,56,56 at 640x480), not Chromium's default rolling pattern.
- Linux needs no extra flags for a V4L2 camera; only PipeWire-only cameras would want `--enable-features=WebRtcPipeWireCamera`, which is pre-existing and not a 32→44 regression.
- The new handler goes on `session.defaultSession`; the existing hCaptcha deny-all handler is on `session.fromPartition('persist:hcaptcha')`. Different sessions, so neither clobbers the other.

### Correction to the earlier comment

`src/main/main.test.ts` **does** run and pass now. `backend-bundle`'s `bundle.cjs` is a webpack artifact; bootstrapping the workspace produced it, so the environmental gap I reported is closed. Desktop unit tests on this head: **118 of 118 suites, 460 tests, 0 failures, 142 snapshots unchanged**, `tsc --noEmit` exit 0.

The dependent PR #3519 has been merged forward onto this head (`cf78d4aaa`) so the chain stays consistent; its own suite is green there too (121 suites, 491 tests).

holmesworcester:

`develop` now carries the onboarding redesign as thirteen squash commits. Five files conflicted.
Rule: the chain's structure wins and this branch's scanner is re-applied on top of it.

**`JoinCommunity.tsx`.** `develop`'s trail-based back navigation, its `TITLED_STEPS` map and its
Recover account step are kept as they are. This branch's `scanQrCode` and `pasteFromQrCode` steps
replace `develop`'s `pasteQrCode` stand-in, so "Join with QR code" opens the camera and the
scanner's "Paste a link" falls back to the paste field. `develop`'s heading ternary goes with the
stand-in: the fallback reads "Paste a link to Join", as this branch had it. Navigation is
`develop`'s `go()` throughout, so the scanner and its fallback join the same trail and the back
arrow retraces them.

**`JoinCommunity.test.tsx` and `LinkDevices.test.tsx`.** Unions. `develop`'s Account recovery
cases and its no-bar-title cases sit beside this branch's camera cases. `develop`'s "takes the
pasted link for Join with QR code since desktop has no camera" case is gone, because desktop now
has one.

**`OnboardingScreens.stories.tsx`.** `develop`'s file, with this branch's `WithCamera` helper and
its nine scanner stories grafted on, and `develop`'s two no-camera paste stand-ins removed. Bar
captions use `develop`'s `droppedBar` rather than this branch's `bar`.

**`packages/e2e-tests/src/tests/onboarding.test.ts`.** This branch's scan case replaces
`develop`'s "join with QR code → paste link" case, whose premise was the missing camera.
`develop`'s walk-back case is untouched.

## Two integration fixes, both following a rule one side already stated

**The Scan QR code step would have had no title at all.** `LinkDevices.tsx` set `withoutTitle`
unconditionally, on the reasoning `develop` writes out: every step repeats the frame's bar title
as its own large heading, so the bar keeps only the glyph. That is no longer true of the scan
step, because this branch replaced its paste form with the camera, which draws a viewfinder and
the sheet's intro and no heading. `develop`'s own comment anticipates the case, "a step without a
heading would take one", so `withoutTitle` is now `step !== 'scan'` and the scanner keeps the bar
title the frame gives it. Its test asserts the bar title is present and the heading is not.

**Declining a scanned device link.** `cancel-device-link` no longer exists: #3524 made the consent
an Agree & join card whose back glyph is the way out, and the card's own test asserts there is no
second button. The link-devices test now clicks `deviceLinkConsentModalBack`.

## Not carried over

The walkthrough's camera toggle. `develop` restructured that component, and grafting this branch's
camera threading into it is not mechanical, so the walkthrough's QR steps still show the paste
stand-in. All nine standalone scanner stories are present and do run the real video to canvas to
jsQR pipeline. Worth a follow-up.

|---|---|---|
| desktop | clean | 125 of 126 suites, 527 passed |
| mobile | clean | 92 suites, 385 passed |
| e2e-tests | clean | not run: they need a packaged AppImage |

PR #3520, opened 2026-09-13
mobile: scan QR codes in the join and link-devices flows

holmesworcester:
Mobile half of the QR scanning work (desktop: #3516). "Join with QR code" and Link devices → "Scan QR code" open a real camera sheet instead of the paste form; a scanned code does exactly what the same link pasted into "Paste a link to Join" does.

## Dependency

- **react-native-vision-camera 4.7.3** (pinned; lockfile carries only the new entry). Its `useCodeScanner` decodes QR codes natively (ML Kit on Android, AVFoundation on iOS) with no frame-processor / worklets / reanimated dependency. Builds against the branch's RN 0.81 + new architecture + Hermes and the x86_64 emulator build (branch `build/android-x86_64`, #3515) without changes. react-native-camera-kit was not needed.
- `VisionCamera_enableCodeScanner=true` in `android/gradle.properties` bundles the ML Kit barcode model (2.4 MB) so scanning works offline and without Google Play services.
- **Patch** `patch/vision-camera/CameraDevicesManager.kt.patch`, applied by `prepare` through the existing `apply-patch.cjs`: VisionCamera builds its device list from CameraX's ProcessCameraProvider, which initialises asynchronously; `getConstants()` and `initialize()` run first and hand JS an empty list, and nothing re-sends it (onCameraAvailable skips ids it already knew). `useCameraDevice('back')` then stays undefined for the process lifetime and the sheet said "No camera is available." — reproduced on the API 36 emulator where CameraX took 5.7 s. The patch emits `CameraDevicesChanged` once the provider is ready.
- Android: `CAMERA` permission + `<uses-feature android.hardware.camera required=false>` (paste still works without a camera). iOS: `NSCameraUsageDescription` in Info.plist so it compiles; **`pod install` / Podfile.lock not run here (no macOS)** — needs a Mac before an iOS build.

## Permission flow

`useCameraPermission()` on mount → status `requesting` (overlay "Requesting camera access…") → OS prompt. Granted → `scanning`. Denied → `denied` overlay with "Paste a link". No back camera / camera error → `unavailable` overlay with "Paste a link". Permission granted later in system settings is picked up on foreground (the hook re-checks on AppState change).

## What changed

- `components/QrScanner/QrScannerSheet.component.tsx` — the two sheets (2811:2460 Join with QR code, 2811:2587 Scan QR code): Appbar with close ✕ and centred title, optional intro, camera edge to edge at the sheets' 375×384, the 216pt framed square (same numbers as desktop). Presentational.
- `components/QrScanner/QrScanner.component.tsx` — camera container: permission, `useCodeScanner({ codeTypes: ['qr'] })`, `parseScannedCode` (= `getInvitationCodes`, the paste field's parser, same as desktop's `decodeQr.ts`). Invalid code → inline error `INVALID_INVITATION_ERROR` (the paste field's own copy, now exported from JoinCommunity.component) and scanning continues; valid code → `onDecoded` once, camera released. `active` pauses the camera while another screen covers the sheet and rescans (same code accepted again) when it returns.
- `screens/ScanQrCode/ScanQrCode.screen.tsx` — one screen, `variant: 'join' | 'deviceLink'` (title/intro per flow, copy verbatim from the prototype). Dispatches through the new `hooks/useInvitationAction.ts`, extracted from PasteInviteLinkScreen and shared with it: device link → `linkDevice` + replace with ConnectionProcess; member link → `joinCommunity` + Choose username. Join with QR code therefore accepts a device link, exactly like paste. Splash until the backend socket is up, like paste. Denied/no camera → `replaceScreen` to the paste form of the flow (`inviteLink` = "Paste a link to Join", or `deviceLink`).
- `JoinCommunity.screen.tsx`, `LinkDevices.screen.tsx` route the QR rows to the scanner; the paste form's `qrCode` variant had no caller left and is removed.
- Storybook (`QrScanner.stories.tsx`, registered in `.storybook/index.js`): every state with a stand-in camera, plus "Live camera (device only)".
- Jest: `parseScannedCode.test.ts`, `QrScanner.test.tsx`, `ScanQrCode.screen.test.tsx` (22 tests); `react-native-vision-camera` is mocked globally in `setupTests.tsx` via `tests/mocks/reactNativeVisionCamera.tsx`.

## Verified

- `npx tsc --noEmit` in packages/mobile: the same 10 pre-existing errors as the base branch (`../state-manager`, `../types` — `@localfirst/*` typings), nothing new. Full jest: 67 suites / 211 tests pass (3 skipped, as before). prettier + eslint clean on every touched file.
- x86_64 debug APK (branch merged locally with `build/android-x86_64`) on a fresh API 36 x86_64 headless emulator with the virtual-scene camera and real QR posters (`-virtualscene-poster wall=…`, camera driven over the emulator's gRPC `rotateVirtualSceneCamera` / `setVirtualSceneCameraVelocity`). Looked at, screenshots in the session scratchpad `mqr/shots/`:
  - `05-join-qr-permission.png` OS camera prompt; `06-join-qr-denied.png` denied overlay + Paste a link; `07-paste-fallback.png` → "Paste a link to Join".
  - `20-join-scanning.png` Join with QR code scanning (camera + framed square); `20-after.png` **real decode of the `validInvitationDatav4[0]` member link → Choose username** (logcat: `utils:invite Invitation data … kind: member parsed` → `Starting joinCommunitySaga` → `Navigating to UsernameRegistrationScreen`).
  - `30-invalid-link-devices.png` Link devices; `30-invalid-scan-sheet.png` Scan QR code sheet with intro; `30-invalid-step-2.png` a non-Quiet QR in the viewfinder → "Please check your invitation code and try again", still scanning.
  - `40-device-step-2.png` / `41-device-now.png` **real decode of a device link (fixture + userId/userName) → "Joining now!" connection process** (logcat: `kind: device parsed` → `Linking this device from a device link` → `Starting linkDeviceSaga` → `Backstack after replace: … LinkDevicesScreen -> ConnectionProcessScreen`).
  - `51-sb-1..10.png` all ten on-device Storybook stories (storybookDebug x86_64 build).

## Undesigned / flagged

- The requesting / denied / no-camera overlays and their copy ("Requesting camera access…", "Camera access was denied.", "No camera is available.", "Paste a link") are not in the prototype — same wording as desktop, minimal.
- `NSCameraUsageDescription` copy ("Quiet uses the camera to scan invite and device-link QR codes.") is mine.
- The sheets are stack screens (the branch's pattern for the QR display sheet), not bottom drawers; the camera keeps the sheets' 375×384 proportion with white below on tall phones.

## Not done

- iOS: compiles only in the sense of the Info.plist/Podfile inputs; no `pod install`, no iOS build here.
- The "Live camera" Storybook story renders the frame but showed a black preview in the storybook APK on the emulator (the main app's camera preview works); not investigated.
- Device-link decode was verified up to the connection process; the fixture's onion address is fake, so joining itself does not complete.

## Follow-ups after the first review

- **Decoded state** (879:16178 in the Device linking file, Dec 2024): a spinner sits over the camera once a code is read, until the dispatch navigates. The prototype's 2026 sheets have no such frame; 879:15946 matches the scanning state already, and 879:15508 ("Linking devices" progress) is the QR *display* side, not applied here.
- **Notifications prompt never stacked on the camera ask.** The scanner requests CAMERA only (VisionCamera's own request code; the app's POST_NOTIFICATIONS request uses code 200, results routed by code). The notifications prompt comes from the push-notifications saga once at store-ready — on the emulator ~0.4 s before Get started appears — and a fresh install → deny at launch → Join with QR code shows the camera prompt alone (`shots/70-notif-launch.png`, `72-notif-first-prompt.png`, `81-after-launch-prompt.png`, `82-after-scanner-prompt.png`; one REQUEST_PERMISSIONS launch per prompt in logcat). The bundled prompt the user saw was not reproducible from this branch's logic; to make it impossible anyway, the launch-time request now waits while the scanner sheet is the current screen and fires when it is left (saga test added). Second launches only re-check.


holmesworcester:
## Merged `design/onboarding-impl` (2d4f0c091) into this branch

Head is now `c2d3f9885`. Merge, not rebase. Two textual conflicts and one semantic conflict the merge could not see. The result is mobile-only; `packages/backend` and `3rd-party` are byte-identical to the root.

### The one that mattered: device-link consent

The root put a consent drawer in front of device linking — `linkDevice` now takes a **required** `deviceLinkConsent` flag and a confirmed endpoint (`confirmedDeviceLinkPayload`). This branch had moved the paste screen's submit path into a shared `useInvitationAction` hook so the scanner could reuse it, and that hook dispatched `linkDevice({ inviteData })` directly, the pre-consent behaviour.

Taking the hook as written would have let a **scanned** device link bypass the consent gate the root had just added. Resolved by making the hook itself consent-aware:

- it clears the join error and joins a community directly, as the root does;
- a device invitation is **handed back to the caller** rather than acted on;
- both the paste screen and the new scanner screen show `DeviceLinkConsentDrawer` and dispatch `linkDevice(confirmedDeviceLinkPayload(data))` only on confirm;
- the scanner's camera stops while the drawer is up, so nothing is scanned behind it.

The branch's intent (one submit path shared by pasting and scanning) is kept; the root's security behaviour is kept. The root's own paste-side consent test passes unchanged against the refactored hook.

### The other two

| Conflict | Resolution |
|---|---|
| `JoinCommunity.component.tsx` | The root rewords the invalid-invitation error to **"Please check your invite link and try again"**. Kept the root's wording, inside this branch's shared `INVALID_INVITATION_ERROR`, so the scanner still shows the same error the paste field does. |
| `JoinCommunity.screen.test.tsx` (semantic, no markers) | The root's test asserted that "Join with QR code" falls back to the paste form *"since this build has no scanner"*. This build has one, which is the point of this PR, so the test now expects `ScanQrCodeScreen`. The sheet still falls back to the paste form when the camera cannot be used; that is the sheet's own job and is still covered. |

### Native config and the patch — verified, not assumed

`react-native-vision-camera` was **not installed** in this worktree, so it was fetched at the pinned 4.7.3 and checked for real:

- `patch/vision-camera/CameraDevicesManager.kt.patch` applies cleanly to real 4.7.3 (`git apply --check` exit 0), and the repo's own `scripts/apply-patch.cjs` lands it.
- The symbols it calls exist in 4.7.3: `sendAvailableDevicesChangedEvent()` is defined in the same class, and `reactContext` is a `ReactApplicationContext`, which has `hasActiveReactInstance()`.
- It is wired into install: `prepare` runs `patch-react-native-vision-camera`.
- Android `CAMERA` permission plus `uses-feature … required="false"` survived; `VisionCamera_enableCodeScanner=true` survived in `gradle.properties`; iOS `NSCameraUsageDescription` survived in `Info.plist`.

### Verified here

- `tsc --noEmit` clean for `packages/mobile`.
- Mobile tests: **86 suites passed, 341 passed, 46 snapshots passed** (3 suites skipped, as on the root). No snapshot needed updating.
- New tests for the consent gate on the scan path: consent is asked before any link, the camera stops behind the drawer, declining links nothing and returns to scanning, and Join with QR code goes behind the same consent.

### Not verified here

- **No APK was built and no device or emulator was used**, so the camera itself, the ML Kit scanner and the patched `CameraDevicesManager` were not exercised on a real Android device. The patch was verified by application and by reading 4.7.3, not by running it. The CameraX device-list race this patch fixes can only be confirmed on hardware.
- Mobile Storybook cannot render in this environment. The `QrScanner` stories were read rather than looked at; they cover scanning, requesting, invalid code, denied, no camera and decoded, for both flows, using a stand-in camera surface.

holmesworcester:

`develop` now carries the onboarding redesign as nine squash commits. Resolution rule: where this
branch and the chain touch the same thing, the chain's structure wins and this branch's QR feature
is re-applied on top of it.

**`App.tsx` — union.** This branch's `ScanQrCode` screen registration and `develop`'s
`RecoverAccount` registration are independent additions at the same two places. Both are kept.

**`JoinCommunity.component.tsx` — chain's copy map, this branch's deletion kept.** `develop`'s
longer doc comment for `COPY` wins, amended to say the QR flows now open the camera sheet rather
than standing in for it. This branch's exported `INVALID_INVITATION_ERROR` is kept, and so is its
removal of the `qrCode` variant, because the QR flows no longer render the paste form at all.
`develop`'s `titleHidden: true` is kept on `inviteLink`, so #3521's no-top-bar rule is not
regressed.

**`ScanQrCode.screen.tsx` — rewired to the chain's consent screen.** #3524 deleted
`ModalBottomDrawer/drawers/DeviceLinkConsent.drawer.tsx` and made Agree & join a screen of its
own rather than a sheet, which is what `PasteInviteLink.screen.tsx` does on `develop` today. This
branch still imported the deleted drawer, so the scanner now presents `DeviceLinkConsent` the same
way: when a scanned device link is pending it returns the consent screen instead of laying a
drawer over the camera. Declining returns to the scanner, which restarts and can read the same
code again, so the behaviour this branch specified is unchanged; only the presentation follows the
chain. The `active={active && !deviceLinkInvite}` pause is gone with the drawer, because the
scanner is unmounted while consent is open.

## Two test changes, and what they do not weaken

**`JoinCommunity.test.tsx`** — the `qrCode` row is dropped from the chain's `it.each` over paste
variants. That case was added by #3521 to assert the bar zone is empty on a variant this branch
deletes. The `deviceLink` row still carries the same assertion, so the chain's rule stays covered
on every variant that still exists.

**`ScanQrCode.screen.test.tsx`** — two assertions were re-expressed against the consent screen
rather than the drawer, keeping the same claims. "The camera stops behind the drawer" becomes
"the camera is not mounted at all while consent is up", which is strictly stronger. Declining now
presses the Agree & join back arrow, because the chain's `DeviceLinkConsent` has no decline button
by design and its own test asserts `device-link-cancel` is absent; the test additionally asserts
the consent screen is gone afterwards. Nothing about what is dispatched, or not dispatched, changed
in either case.

|---|---|---|
| mobile | clean | 91 suites, 383 passed |
| desktop | clean | 121 of 122 suites, 486 passed |
| state-manager | clean | 69 suites, 300 passed |
| common | clean | 9 suites, 109 passed |

unless the backend bundle is built. It fails identically on `develop`. Detox was not run; this
host has no usable emulator.

PR #3521, opened 2026-09-13
onboarding: no top bar on full-screen h1 stages (mobile + desktop)

holmesworcester:
Applies the decision **No top bar title on full-screen h1 stages** (ONBOARDING.md; prototype f6Nr5b5wtvk6Xoh1HJZ8Dd, title nodes `visible:false`) to the app on both platforms and to the Screens/Onboarding stories. The bar zone stays for the back/close glyph; the title text and the hairline go; the large heading is the title. Titled bars remain on sheets (QR code, Scan QR code, Join with QR code), the community home and the non-onboarding modals. Link devices is not touched here (design/link-devices-paste applies the same rule to it).

Stacked on design/onboarding-entry; merges design/storybook-grid @ 365a1e662 first (the decision text, flow.json `hiddenTitle`, the Device-linking desktop exports).

### Desktop (`ui/Modal` gains `withoutTitle`)

| screen | frame | before | after |
|---|---|---|---|
| Join community | 2811:2562 | bar "Quiet" + back arrow | back arrow only, no hairline |
| Account recovery | 2811:2535 | bar "Account recovery" + back | back arrow only |
| Open invite link | 2811:2455 | bar "Join with invite link" + back | back arrow only |
| Paste a link to Join | 3190:10892 | bar "Join with invite link" + back | back arrow only |
| Join with QR code (paste stand-in) | 2811:2460 | bar "Join with QR code" | unchanged (sheet keeps its titled bar) |
| Create a community | 2811:2451 | bar "Create a community" + back | back arrow only |
| Choose username | 2811:2371 | bar "Create a community" + close | close glyph only |
| Want a server? | 2922:10009 | `withoutHeader` (no bar) | unchanged |
| Get started | 2811:2550 | no bar | unchanged |
| Link devices | 2811:2575 | bar "Link devices" | left to design/link-devices-paste |

Glyph position unchanged (the entry branch's placement in the 60px zone). Content starts 24 under the zone (OnboardingBody's top padding), as in the Device-linking desktop frame 879:20987.

### Mobile (`Appbar` gains `withoutTitle`)

The mode renders the library's 60-tall bar zone with the glyph box 28×28 at (14, 16), a 16px icon (measured from the frames: arrow bbox 20–36 × 24–36), a 48 touch target via hitSlop, no title, no hairline. Content is top-anchored 24 under the zone (heading top at 84 as in the frames) instead of vertically centred.

| screen | frame | before | after |
|---|---|---|---|
[...]

PR #3514, opened 2026-09-13
onboarding: Get started, Recover account, Join with invite link

holmesworcester:
Stacked on #3512 (`design/onboarding-impl`, the phase-1 onboarding redesign): Get started as the entry, the Recover account screen with its two routes, and the wired Join with invite link path, together in one diff. Five commits, `69e63e4b7..d57ddd5d9`.

## What changed, per screen

**Get started (Figma 2811:2550) — the entry.** `Onboarding/GetStarted.tsx` opens itself once the app is connected without a community, unless an invitation is already being processed, and Join a community / Create a new community / Link devices each return to it from their back arrow. `GetStarted.test.tsx` pins those three facts. The component and container were already on the base; this branch pins the entry behaviour and records it in the library.

**Recover account (Figma 2811:2535).** Join community's third row is enabled and opens Account recovery: title bar "Account recovery", the frame's own "Icon=Vpn key" glyph at 64px, "Recover account", "Locked out? You can recover with a linked device or ask an admin to send you an invite link.", and three rows — *Use linked device* → Link devices, *Use invite link* → Join with invite link (the prototype's two links), *More options* (drawn but unlinked in the file, so present and inert; its glyph is the frame's Material `more_horiz`, gray 50). There is no recovery mechanism: both routes are the existing flows. Desktop: `Onboarding/RecoverAccountComponent.tsx` as a step of the Join community modal, which now keeps a trail of visited steps so the back arrow retraces Recover account → Join with invite link the way it came; *Use linked device* hands over to the Link devices modal. Mobile: `RecoverAccount` component + screen registered in the stack, reached from `JoinCommunityOptions` — mobile is touched only because its neighbouring screens are already on the base.

**Join with invite link (Figma 2811:2455).** Wired as the design has it: Join community → Open invite link (the Monster screen) → *Paste a link* → Paste a link to Join → join. When the user follows the screen's instruction and opens the invite link instead, Electron hands it to the renderer as a custom-protocol invitation and `customProtocol.saga.ts` handles it like every deep link: `joinCommunity`, then Choose username on top. The join modal keeps its step underneath, so closing Choose username returns to the screen the link arrived on rather than to Get started. Nothing in the modal duplicates or blocks that path; an RTL test in `customProtocol.test.tsx` mounts Get started, Join community and Choose username, opens the invite-link screen and dispatches the deep link.

[...]

PR #3523, opened 2026-09-13
onboarding: Join community shows the prototype's heart-chat illustration (mobile + desktop)

holmesworcester:
The prototype's Join community frame (2811:2562) opens with the heart-chat illustration — instance `I2815:2504;6181:27547` (Graphic--heart-chat, 219×160) directly under the title bar, the title 24 below it, the rows 24 below that, top-anchored. Neither platform showed it: the mobile screen had no illustration and centred its content vertically; the desktop screen (the same column in the Modal full-window shell, per the layout canon) had none either.

## What changed (packages/mobile)

- `src/assets/icons/png/graphic_heart_chat{,@2x,@3x}.png` — the Figma images-API export of that instance at 1×/2×/3× (never redrawn). The SVG export of this node embeds a 1024×1024 raster (1.2 MB), so the PNG set is the faithful and lighter asset, as the branch already does for the Get started logo and the monster.
- `JoinCommunityOptions.component.tsx` — `Image` at `GRAPHIC_SIZE` (219×160) centred under the Appbar, then the title, then the rows; the column is top-anchored (`gap` 24) as in the frame (graphic y 60, title y 244, rows from y 302). The Appbar is untouched (title handled by #3521).
- `JoinCommunityOptions.stories.tsx` (registered in `.storybook/index.js`) and `__tests__/JoinCommunityOptions.test.tsx` with a snapshot and a graphic-size assertion.

## What changed (packages/desktop)

- `src/renderer/static/images/onboarding/heart-chat@2x.png` — the same 2× export (438×320), registered in `Onboarding/icons.ts` as `heartChatIllustration` with `HEART_CHAT_SIZE` (219×160), the way the monster and the Get started mark are.
- `Onboarding/JoinCommunityOptionsComponent.tsx` — the graphic in `OnboardingBody`'s `leading` slot, drawn at 219×160 and centred above the h1, exactly as `OpenInviteLinkComponent` does with the monster; the body's 24 gap puts the title 24 below it and the rows 24 below that.
- `Onboarding/OnboardingBody.tsx` — a `flushLeading` prop (padding-top 0) so the graphic sits directly under the bar as the frame draws it (y 60 = the bar's height); every other screen keeps its 24 top padding.
- `Onboarding/JoinCommunityOptionsComponent.test.tsx` — the graphic is an `img` (aria-hidden) before the heading, the heading before the rows, the body flush, and the two rows still route. The Screens/Onboarding · Join community story and the Components/Onboarding · JoinCommunityOptions story render the component and so show the graphic at both widths.

## Verified

- Mobile: `npx tsc --noEmit` in packages/mobile: the same 10 pre-existing errors as the base branch, none new. Jest: the new suite passes (2 tests, 1 snapshot). prettier + eslint clean on touched files. Looked at on the x86_64 API 36 emulator (branch merged locally with `build/android-x86_64`): `shots/80-join-illustration.png` — illustration under the title bar, "Join community", the three rows.
- Desktop: `tsc --noEmit -p tsconfig.json` clean; jest `Onboarding/*` + `CreateJoinCommunity/JoinCommunity` 18 passed; prettier clean. Looked at in the desktop demo AppImage (the design stack): Get started → Join a community shows the graphic above "Join community"; Open invite link still shows the monster.


holmesworcester:
## Merged `design/onboarding-impl` (2d4f0c091) into this branch

Head is now `24a6ff72c`. Merge, not rebase. **No conflicts** — the root's onboarding work and this branch's illustration touch different things. All 14 files of this branch's contribution survive the merge unchanged, and `packages/backend` and `3rd-party` are byte-identical to the root.

### Snapshot churn

One snapshot updated, `JoinCommunityOptions.test.tsx.snap`, and the whole diff is inside the Appbar: the root makes the bar taller (`minHeight` 52 → 60, `maxHeight` 52 → 64), adds an 8px `hitSlop`, a 44pt minimum touch target, and stretches the side slots. No part of the illustration's own output moved. Absorbed in `24a6ff72c`.

### Verified here

- `tsc --noEmit` clean for `packages/mobile` **and** `packages/desktop`.
- Mobile: **84 suites passed, 319 tests passed, 47 snapshots passed** (3 suites skipped, as on the root).
- Desktop: **114 suites passed, 404 tests passed, 142 snapshots passed**.
- **Looked at** the desktop Join community screen rendered through `lightTheme` with the real image exports: the heart-chat graphic sits flush under the bar with no padding above it (`flushLeading`), the heading 24 below, then the three-way choice with Recover account greyed out.
- **Looked at** the illustration asset itself. The desktop and mobile copies are the same file (`md5 ffc3c117…`), and the mobile exports are exactly 219×160 / 438×320 / 657×480 for 1×/2×/3×, matching `GRAPHIC_SIZE`.

### Design note

This screen keeps `<Appbar title={'Quiet'} back={…} />`. That bar title is **pre-existing** and untouched by this branch — this PR adds no bar title to an h1 screen. The `withoutTitle` prop from the h1-no-bar work is not in this base yet, so removing the title here is out of scope for this PR.

### Not verified here

- Mobile Storybook cannot render in this environment, and no APK was built, so the mobile screen was checked by reading its story and its snapshot rather than by looking at it on a device. The illustration is a plain `Image` at a fixed `GRAPHIC_SIZE`, and the asset is byte-identical to the desktop one that was looked at.
- `src/main/main.test.ts` fails to load with `Cannot find module 'backend-bundle'`, a build artifact that was never produced in this worktree. Pre-existing and unrelated; this PR touches no main-process code.

holmesworcester:

`develop` now carries the onboarding redesign as ten squash commits. Resolution rule: the chain's
structure wins and this branch's illustration is re-applied on top of it.

**`JoinCommunityOptions.component.tsx` (mobile).** The bar takes `develop`'s
`<Appbar withoutTitle …>`, which is #3521's no-top-bar rule, in place of this branch's
`title={'Quiet'}`. The old vertical centring is gone, as the chain has it. The illustration goes
back as the first child of the column.

One judgement call, flagged because it is the only place I did not take `develop`'s line verbatim:
`develop` pads the column 24 at the top, so content starts 24 under the bar zone. This branch's
frame reading puts the illustration flush at y 60, which is the bar zone's own height, with the
24 falling between the illustration and the title. I kept it flush, so the padding is dropped and
the `gap` supplies the 24. Two reasons: the illustration is what the frame puts 24 above the
title, so a 24 above the illustration as well would add a gap the frame does not draw; and the
desktop column is already flush, through the `flushLeading` prop this branch added to
`OnboardingBody`, which never conflicted. Padding it would have split the two platforms. Every
other screen keeps `develop`'s 24, because none of them has a leading graphic.

**`JoinCommunityOptionsComponent.tsx` (desktop).** Doc comment only. Union: this branch's
illustration geometry plus `develop`'s Recover account routing, which replaced this branch's now
stale "no mechanism yet, so its row is present but disabled".

## Test and story changes

- **`onRecoverAccount` is a required prop now** (#3514). This branch's desktop test, mobile test
  and mobile story predate it and did not pass it, so all three do.
- **`JoinCommunityOptions.test.tsx` — "routes the two enabled rows and keeps Recover account
  disabled" becomes "routes all three rows".** The disabled assertion was true when this branch
  was written and is false on `develop`: #3514 gave Recover account a screen. The case now presses
  all three rows and asserts all three handlers fire, so it tests more than it did, not less.
- **One snapshot regenerated**, `JoinCommunityOptions.test.tsx.snap`. It was recorded against the
  titled bar; the chain's glyph-only 60-tall bar zone changes the markup. The regenerated file
  carries `testID="appbar_without_title"`, no `appbar_title`, the 60 height and
  `join-community-graphic`, which is exactly the intended combination.

|---|---|---|
| mobile | clean | 92 suites, 385 passed |
| desktop | clean | 122 of 123 suites, 487 passed |

Issue #3368, opened 2026-07-23
Unclear messages when QSS / hCAPTCHA is failing due to network unavailability


holmesworcester:
Steps to reproduce: 

1. Start Quiet and start creating a community with a server
2. Cut internet access before choosing a username

Expected: some clear sign that we are loading a CAPTCHA or something, and a timeout with a useful error message if it doesn't work.

Actual: no useful message, just a "joining" screen, which also doesn't make much sense because we were creating a community not joining one.

FWIW the CAPTCHA did reappear when I enabled Internet again. And it might have thrown an error eventually, but I should have seen something sooner I think.

PR #3525, opened 2026-09-13
design: the Tor explanation is for Tor communities, not for communities on a server

holmesworcester:
Joining or creating a community on a server showed the Tor explanation: a screen about why Tor is slow, shown to someone not waiting on Tor. This splits progress into the two screens the design draws and branches on whether the community is hosted on a server.

## The two screens

| | over Tor | on a server (QSS) |
|---|---|---|
| design | `Joining now` (`0j7Nna9zWmfOSNmRmQK1Uh` **5978:19161**), status from `Joining` (`y8h6w8PYR9jyI3zjYHL9Cl` **1316:34596**) | mobile `Joining now` **5978:19142** / **2894:3382**; desktop Draft 6 Frame 1320 **1430:48030** |
| shows | globe, **Joining now!**, bar, status **Connecting via Tor**, the explanation, **Learn more about Tor and Quiet** | desktop: bar, **Joining community "Rockets"**, second line. mobile: globe, title, bar, the status line the app already printed |
| Tor copy | yes, it is about Tor | none |

Both carry `data-testid="joiningPanelComponent"`, the marker the rtl tests and `e2e-tests/src/selectors.ts` use for "the joining screen is up".

## A second place Tor was named

The QSS screen's status line comes from `ConnectionProcessInfo`, and one member read `Connecting to community members via Tor`. The backend emits it at the end of storage initialisation for **every** community (`connections-manager.service.ts:1191`), so removing the explanation alone would have left that line on the server screen, which is most of what the report is about.

The clause is dropped, not replaced: `Connecting to community members`. The sentence stays the app's own, and Tor is still named where it is the subject, in the Tor screen's designed status line. Worth knowing while reviewing: a server-hosted community still spawns a Tor hidden service (`spawnTorHiddenService`, unless `LOCAL_TRANSPORT`), so the old wording was not false, just not what the person was waiting on.

Sweeping the rest of the enum for Tor, hidden-service and onion vocabulary turned up one more: `SPAWNING_HIDDEN_SERVICE` read `Spawning hidden service for community` and now reads `Setting up your community’s address`, plain and true on both paths. The key is unchanged. `TOR_BOOTSTRAP_COMPLETE` keeps its wording because it is compared, never displayed, and `Initializing libp2p` / `Initializing IPFS` are developer vocabulary rather than Tor vocabulary, so this rule leaves them.

That one was not reachable, which is worth saying plainly rather than claiming a fix that was not: `connection.slice.ts` stores only four phases into `connectionProcess.text` — `CONNECTION_STARTED`, `BACKEND_MODULES`, `CONNECTING_TO_COMMUNITY`, `LOADING_MESSAGES` — and drops every other value the backend emits. So nobody could read the hidden-service wording today. It is fixed so the next phase wired in cannot reintroduce Tor under the bar, and the test is written to hold that line: `ServerJoiningPanel.test.tsx` runs over **every** value the enum can hold, not the four in play, asserting none renders `Tor`, `hidden service` or `onion`. It fails against the old wording.

## How "uses a server" is decided

New selector `communities.usesServer`. The progress screen is up across three stages that each know the answer from a different place:

1. **Joining** reads the invite: v5 carries `qssEnabled` / `qssEndpoint`. Available before any community record exists, which is most of the joining window.
2. **Creating** reads the owner's own submission. The slice keeps it as `pendingCreate` — the name they typed and their answer about a server — from the moment `createCommunity` is dispatched until the backend replies, because the record does not exist yet either.
3. **Afterwards** the community record the backend handed back is authoritative (`Community.qssEnabled`), and the pending create is dropped on `addNewCommunity`.

`communityInProgress` reads the heading's two halves — the name, and whether the person waiting is the owner — from the same pair. Reading them off the record alone opened the create path's first frame headed `Joining community`, with no name, because on that path the record is the last thing to arrive. `rtl-tests/loadingPanel.test.tsx` renders the panel with nothing dispatched but `createCommunity`, asserts no record exists, and asserts the heading already says Creating; it fails against the old derivation.

`joinCommunitySaga` now gates the terms of service on the same `inviteUsesServer` helper it used inline, so the two questions cannot drift apart.

## Screens changed

| platform | file | change |
|---|---|---|
| desktop | `LoadingPanel/JoiningPanelComponent.tsx` | now the branch between the two panels |
| desktop | `LoadingPanel/TorJoiningPanel.tsx` (new) | the explanatory screen, restyled to **5978:19161** |
| desktop | `LoadingPanel/ServerJoiningPanel.tsx` (new) | the Draft-6 layout, **1430:48030** |
| desktop | `LoadingPanel/LoadingPanel.tsx` | selects `usesServer`, the community name, and whether a sidebar is mounted |
| mobile | `ConnectionProcess/ConnectionProcess.component.tsx` | the same branch; teal fill; the 4px type scale's variants instead of raw sizes |
| mobile | `ConnectionProcess/ConnectionProcess.screen.tsx` | selects `usesServer` |
| both | `ui/ActionProgress` | cherry-picked from `design/action-progress-states` (`d0098af22`), the bar these screens draw |

`StartingPanelComponent.tsx` is **not** changed. App start is a different screen with its own design (`2811:2770`), and this branch is about joining and creating.

## The sidebar in the desktop frame

**1430:48030** draws the split view with the sidebar greyed, because it draws the app mid-creation. The app's `Sidebar` returns `null` until both a current community and a current channel exist, which is most of the window this screen covers. So the panel dims the sidebar only when one is actually mounted, and otherwise is the frame's chat area alone, filling the window. Both cases have a story.

## The opening sentence, checked on each platform

The designer's paragraph (**5978:19167**) opens `You can exit the app - we'll notify you once you're connected!`. That was checked against each app rather than taken on faith, and it does not hold on either.

**Desktop — false outright.** Closing the window quits the app and the join stops. The sentence is now `Keep Quiet open while you connect.`, the truthful equivalent in the same register.

**Mobile — half true, and it is the *other* half that holds.**

| claim | holds? | evidence |
|---|---|---|
| you can leave the app | **yes** | `CommunicationModule.syncBackendWorkerState` (`Communication/CommunicationModule.java:382-390`) takes the `enqueueRequests` + `wake` branch whenever `!QuietStorage.isTeamQssEnabled()`, which is exactly a Tor community. The hibernate branch below it is reserved for communities *on a server* whose user has not enabled background Tor, the opposite of what one might expect. `SHOULD_RUN_BACKEND_WORKER=true` in `.env.production`, `.env.staging` and `.env.development`, and the worker runs as a `dataSync` foreground service (`AndroidManifest.xml:73-75`). |
| we'll notify you when connected | **no** | The only notifications the app posts are incoming messages — `notificationHandler.notify(message, username)` from `BackendWorker.kt:406` (socket `pushNotification`) and `CommunicationModule.java:167` (FCM) — plus the foreground service's own persistent one (`Backend` / `Continuously synchronise data and receive incoming messages`). Nothing fires on join completion, and there is no local-notification library on the JS side. |

So on mobile the permission stays and the promise goes: `You can exit the app while you connect.`

The bold timing sentence and everything after it are the designer's, unchanged, on both platforms. The tests assert the false claims are *absent*, not only that the true ones are present.

Second line on the Tor screen: the app's phase text stays there, as decided.

## Verified

- `tsc --noEmit`: **0 errors under `src/`** in desktop and mobile. state-manager has 32 errors on this branch and **the same 32 on the base** (`design/agree-and-join`), all in the auth gitlink and the invite selectors, none in changed files.
- jest:

| package | result |
|---|---|
| desktop | 107 suites, 354 passed of 360 (5 skipped, 1 todo), 127 snapshots |
| state-manager | 60 suites, 222 passed |
| mobile `ConnectionProcess` | 11 passed |

  The rest of the mobile suite is broken on the base and on this branch identically: 44 failed suites, 23 failed tests, 23 failed snapshots, the same set on both, from a pre-existing transform error. This branch turns 81 passing mobile tests into 85.
- Prettier clean on every changed file.
- Snapshots: one updated (`ServerJoiningPanel`, new), one written. `JoiningPanelComponent.test.tsx` replaces two large markup snapshots with behaviour tests for both branches, including `expect(baseElement.textContent).not.toContain('Tor')` on the server screen.

### Screenshots looked at

Frozen static build served on 127.0.0.1:6533, captured headless and read back. The Tor shot is from after the opening sentence changed:

| story | screenshot |
|---|---|
| `components-loadingpanel--joining-over-tor` | `/mnt/storage/holmes-tmp/claude-1000/-home-holmes/49757000-0775-4515-b349-71bfb4788e68/scratchpad/aps/shots4/components-loadingpanel--joining-over-tor.png` |
| `components-loadingpanel--joining-on-a-server` | `…/shots4/components-loadingpanel--joining-on-a-server.png` |
| `components-loadingpanel--creating-on-a-server` | `…/shots3/components-loadingpanel--creating-on-a-server.png` |

Every new story sets `chromatic: { disableSnapshot: true }` per story, so the existing `JoiningPanel` and `StartingPanel` stories keep their Chromatic baselines.

## Noticed, not fixed

`state-manager` re-exports a `LoadingPanelType` of its own (`sagas/network/network.types.ts`) carrying only `StartingApplication` and `Joining`, while the slice, the sagas and the panel all use the four-member enum from `@quiet/types`. Anything importing the name from `@quiet/state-manager` cannot refer to `Creating` or `Failed`. Left alone; the new test imports the real one and says why.

## Not done

- **Mobile Storybook is not gated.** The mobile stories are `storiesOf` entries in the React Native Storybook, which this box cannot render headlessly. The two new mobile variants are covered by jest, not by a screenshot.
- **`SPAWNING_HIDDEN_SERVICE`** still reads `Spawning hidden service for community` and shows on the server screen too. It is Tor vocabulary but it is accurate, since the hidden service is spawned either way, so it is left alone rather than reworded on my own.
- **The error / timeout state** (#3368) is undesigned and keeps the app's existing message, as instructed.
- **`StartingPanelComponent`** still renders the logo alone. Its design (`2811:2770`) draws bar, status and message, and ONBOARDING.md lists it as unsettled. Out of scope here.


holmesworcester:
## Brought up to date with `design/agree-and-join` (now `d67469917`, which carries the root)

Merged, not rebased. Head is now `b28c9ea72`.

### Conflicts and how they were resolved

**`JoiningPanelComponent.tsx`** — the substantive one. This branch had already split the single joining panel into two screens (`TorJoiningPanel`, `ServerJoiningPanel`) with this file as the router; the parent had meanwhile added develop's admission-reset failure UI *inside* that single panel. A failed reset is a state, not a transport — there is no progress to draw on either screen — so it is drawn **before** the split, by a new `ResetFailedPanel` carrying develop's heading, message and `retry-admission-reset` action in the same shell and globe. The reset props pass straight through. The rtl tests that press that button pass unchanged.

**`communities.slice.ts`** — kept both sides: the parent's `admissionResetStatus` / `admissionResetResult` / `joinCommunityError` fields and their reducers, and this branch's `pendingCreate`. `createCommunity` takes the named action, because this branch's body reads the payload. `linkDevice` is the parent's no-op, which is develop's current behaviour; this branch's older body (clearing a draft join) is dropped rather than reintroduced.

**`LoadingPanel.tsx`** — union of the two import lists (all five symbols are used in the merged file) and both prop sets on the panel: this branch's `usesServer` / `communityName` / `withSidebar` and the parent's reset props.

**`ConnectionProcess.screen.tsx`** (mobile) — this branch's `usesServer` selector together with the parent's newer error handling (`currentCommunityId`-based check, `launchError`, `invalidInvite`), which is what drives the reset path.

**`loadingPanel.test.tsx`** — both sides added whole tests at the same place; both kept. The two `LoadingPanelType` aliases the sides introduced were unified on one.

### One test updated

`joining.process.test.tsx` (mobile) read the connection phase from `connection-process-text`. On a Tor community that element now carries the transport ("Connecting via Tor", Figma 1316:34596) and the phase moves to `connection-process-secondary` — which is the point of this branch. The assertions follow the phase to its new line and also pin the status line.

### Verified

- Typecheck: desktop, mobile and state-manager all clean (the workspace libraries were rebuilt first, since this branch adds a selector the built `lib/` did not have).
- Tests on a real install: desktop **121 suites, 484 tests, 142 snapshots**; mobile **87 suites, 333 tests, 46 snapshots**; state-manager **69 suites, 300 tests, 13 snapshots**. The 13 `usesServer` selector cases pass, including the pending-create and pending-join paths. No snapshots were updated.
- The reset-failure path is covered end to end by the rtl suite: the retry action, the pending receipt cleanup without a community entity, the final persistence failure surviving the root reset, and the invalid-invite error after finalization all pass against the new structure.
- Static desktop Storybook built and 68 stories screenshotted in Electron under xvfb and looked at individually. The new ones read correctly: `ActionProgress` (teal bar, status, optional secondary line), Joining over Tor (globe, "Joining now!", "Connecting via Tor" over the phase, then the Tor explanation and the link), Joining on a server and Creating on a server (bar, "Joining/Creating community “Rockets”", the phase, **no Tor explanation**), and the in-community variant with the sidebar.
- **No-bar check: no story shows a bar title over a large heading**, on either platform.

### Not verified

- `ResetFailedPanel` has no Storybook story; it is covered by the rtl tests only.
- Nothing was run on device or in the packaged app, and no e2e suite was run.

holmesworcester:
### Follow-up: the `check` job's Lint step

The `check` workflow ran red on this branch after the merge above. Cause: the repo's eslint runs prettier 2.3.0 through `eslint-plugin-prettier`, which formats slightly differently from the standalone prettier binary the new test files here were written with. CI lints **mobile** but not desktop (`packages/desktop` has no `lint-ci` script, so `lerna run lint-ci` skips it), so the mobile file was the actual failure; the desktop ones are fixed too for consistency.

Fixed with `eslint --fix` on the files this branch adds, and re-merged down the chain so every PR carries it. `packages/mobile` and `packages/state-manager` now report **0 errors** (warnings only, as on the root). Tests re-run after the reformatting and still pass.

Two desktop files still fail lint — `TextMessage.blankLines.cy.tsx` and `TextMessage.blankLines.test.tsx` — but they fail identically on `design/onboarding-impl` and on `develop`, with the same configs and the same prettier version, and desktop is not linted in CI. Pre-existing, not touched here.

PR #3512, opened 2026-09-12
Onboarding redesign phase 1 on desktop + mobile, with #3400 device linking

holmesworcester:
Implements phase 1 of the onboarding redesign from `packages/desktop/src/renderer/design-system/ONBOARDING.md` on desktop and mobile, stacked on #3483 (→ #3423 → #3422), with `design/storybook-grid` (tokens, specimens, the spec) and the device-linking feature from #3400 brought in. Draft: stacked, nothing here is meant to merge before its base.

## Stack and merges

- Base: `docs/qss-notification-regression-plan` (#3483). Merged `design/storybook-grid` cleanly.
- **#3400 (`feat/2610-device-linking`) came in as a single 3-way patch of `origin/develop..origin/feat/2610-device-linking` (43 commits), not as a merge commit.** A true `git merge` produced 71 conflicted files, most of them unrelated to device linking (ios/*, android/*, `Sidebar.tsx`, LFA identity specs, tor binaries…), because #3400 sits on develop's squashed "Merge 9.0.0 into develop" (#3459, 328 files) which collides with the 10.0.0 line this branch is built on. A merge commit would also have marked #3459 as merged while its tree is not. The patch conflicted in 9 files, all resolved keeping both sides:
  - `state-manager` communities slice / master saga (+test): pendingJoin draft-resume (#3426) kept, `linkDevice` reducer + saga branch added; `linkDevice` clears a draft join the way `createCommunity` does.
  - `mobile` deepLink saga (+test): add/add, both test blocks kept; device-invite types imported, lodash not re-added.
  - `backend/libp2p.auth.ts` (12 hunks, hand-merged): the per-transport LFA session model from 6ac454a2 (sessions keyed by connection id, `connection:close`, `rememberAuthenticatedPeer`) with #3400's coordinated-admission gate layered on (`AdmissionAuthContext`, `failedAdmissionPeers`, `advanceToNextBufferedPeer`, `LOCAL_TRANSPORT`). **Please review this file.**
  - `backend/libp2p.service.ts`: `hangUpAuthTransport` runs inside #3400's 1s auth-error grace timer + admission-gate check; `LOCAL_TRANSPORT` switch wraps the connection-protector timeout.
  - `backend/connections-manager` (+spec): #3400's `bootstrapCommunityFromInvitation` helpers beside the mutex-serialised create/join (#3424); `linkDevice` is serialised behind the same mutex (my addition).
[...]

PR #3526, opened 2026-09-14
desktop: sidebar aligned with the Quiet Design Library (V1)

holmesworcester:
Stacked on #3512 (`design/onboarding-impl`). Desktop only — mobile's home screen is a separate branch.

## What

The desktop channels-and-users column now renders the Quiet Design Library's **V1 release variant**: the community tile, name and caret at the top; an "Add members" row; a "Channels" section whose header carries the (+); the members listed as the library's people rows; and the signed-in user pinned to the bottom behind a hairline.

The geometry is read off the Figma nodes rather than eyeballed, and it is all in one file (`sidebarMetrics.ts`) with the node ids beside each number. The glyphs are the library's own SVGs, transcribed with `white` swapped for `currentColor`. `theme.ts` is untouched.

Two new primitives replace per-panel styling: `SidebarRow` (the library's `List item` / `List item--people`) and `SidebarHeader` (`List title`). Because every row is now one full-width component, the selected, hover and pressed overlays span the whole 220px column in every section, instead of only the part the old markup happened to paint.

## Why

The sidebar was the last large desktop surface still carrying its own inlined geometry — each panel had its own row height, padding, type and opacity, none of them traceable to the library. Aligning it closes that gap and gives the remaining desktop work a row primitive to build on.

## Design nodes

Quiet Design Library, file `0j7Nna9zWmfOSNmRmQK1Uh`, page "Structure & Nav", section **"Desktop and mobile sidebar and titlebar"** (`6107:21330`):

| Node | What |
| --- | --- |
| `6218:16416` | **Desktop sidebar — V1 release, Content=For V1 2025** (the target) |
| `6222:13638` / `6220:24045` | The designer's V1 requirements note |
| `4233:12959` | Team — the community header row |
| `3797:16103` | List title — Default / Hover |
| `3797:16113` | List item — Default / Hover / Selected |
| `4606:16448` | List item--people — Single / Group |
| `4716:7702` | Profile summary — the own-user row |
| `3797:16806` | List group (Type=Placeholder is the skeleton) |
| `3797:15261` | controls--mac — the window-controls space |
| `5196:15387` | Community icon top-level |
| `5573:9585` / `5573:9597` / `5573:9602` | The designer's state annotations |
| `5439:58626` | The full component set — **not** the target; see below |
| `5446:61649` | Search input — omitted for V1 |
| `5582:19302` | drop-line — not used; Quiet has no channel reordering |

## Scope, from the designer's V1 note

The note (`6222:13638`) reads: *"Top bar without search / Channel list with unread badges / Add members button / No Floating Action Button / No dark mode yet / Create channel "+" button / DMs with avatars and badges / Tapped state for all clickable stuff."*

- **No search field.** The library's full set puts one under the community row; V1 does not. The app's channel search is untouched and still opens on Ctrl/Cmd+K: `SearchModal` is mounted from `Root.tsx`, not from the sidebar, so only the field goes.
- **No dark mode.** Nothing in the sidebar varies by theme, and the app's dark theme keeps today's `sidebarBackground` until a dark sidebar is designed for release. The dark Storybook story stays only as a guard that the column does not break under it, and its name says so.
- **A pressed state on everything clickable.** The library has no Pressed variant anywhere, so pressing reuses the Selected overlay rather than introducing a third value: rows and the community header take it full-width, the section (+) takes it on the glyph.

## State overlays

Taken from the designer's annotations beside each primitive, and confirmed against the nodes' fills:

| Primitive | Default | Hover | Selected |
| --- | --- | --- | --- |
| `List item` | none | white 5% | white 10% |
| `List item--people` | none | white 5% | white 10% |
| `Profile summary` | none | white 10% | — |
| `List title` | none | (+) 60% → 100% | — |

These are the library's exact values rather than `theme.palette.colors.sidebarHover` / `sidebarSelected`, which carry the same two overlays to within one hex step. Using the library's numbers directly also keeps the sidebar off `theme.ts` while another branch in this stack edits it. Measured on the rendered page: a hovered row composites to 4.8% white, a selected row to 9.9%.

`List title` State=Hover also fills its row white 10%. That is deliberately not applied — in Quiet a section title does nothing, only its (+) is clickable, and a hover fill would promise otherwise.

## Deviations from the design, and why

- **"Direct messages" reads "Users", with no (+).** Quiet has no direct messages. The section lists the community's members, styled as the library's people rows; there is no conversation to start from here, so the section header has no action.
- **The unread badge is a dot, not a numbered pill.** Quiet tracks unread per channel as a boolean, so there is no count to show and the component will not invent one. The row's label also goes to medium weight when unread, so the state does not depend on colour alone. `SidebarUnreadBadge` still accepts a `count` and renders the library's pill if Quiet ever tracks one.
- **No unread bubble on the community tile.** `Team` hangs a "99+" badge off it; Quiet counts no messages, for the same reason the channel badges have no number.
- **The member rows keep their online dot**, which the V1 content frame does not draw. The library's own `List item--people` has an "Online indicator", Quiet has a real signal for it (whether we hold a connection to that member's peer), and the app already drew it. Dropping it would have removed working behaviour to match a content mock.
- **Threads, Drafts and Starred are left out.** Quiet has none of them.
- **`TorStatus` is unchanged** and was not mapped onto the library's Node status. It renders only in development builds and nothing at all in a release, so there is no production slot for Node status to occupy; doing it properly means deciding what Quiet's connection states are, which is its own change.
- **No skeleton variant.** `Team`, `List group` and `Profile summary` all have a Placeholder variant; the app's sidebar returns nothing until there is a community and a channel, so there is no loading state on screen to skeletonise.
- **The community name is 20/28, not the library's 18/27.** That is the standing 4px-grid decision for the type scale, already applied across the design work.

## Wiring

Nothing new was invented: the (+) beside "Channels" opens the existing create-channel modal and is drawn only when the user actually holds the create permission; the community tile and caret open the community settings drawer the header has always opened; the bottom row opens the existing profile menu. "Add members" opens the settings drawer straight on its invite tab, which is the one behavioural addition — `SettingsComponent` now takes the tab to open on.

## e2e harness

The sidebar's markup is the library's now, and two rows changed element: the community header and the own-user row are `<button>` where they were a `<span>` and a `<div>`. Every xpath in `packages/e2e-tests/src/selectors.ts` that named a tag alongside the test id broke, and `Sidebar.openSettings()` breaking took every test that reaches Settings with it. Each xpath that reaches into the sidebar now matches on the test id alone, which is what most of that file already did. All nineteen resolve against the running app.

`switchTab()` needed a second fix. "Add members" asks for the invite tab, and a tab's drawer sits over the settings menu, so the next `switchTab()` aimed its click at a menu item and hit the invite tab's "Copy to clipboard" instead. It now returns to the menu first when a tab is open, rather than leaving every caller to remember `closeTab()`. Measured in the running app with `elementFromPoint`: before, a click on the About tab lands on `invite-a-friend`; after, it reaches the tab.

Two product-side changes keep a focused open from leaking: `SettingsComponent` reads `focusTab` once as the drawer opens rather than on every render, and `IdentityPanel` opens with explicitly empty args, because the modal reducer keeps the previous args when a caller passes none.

## Verification

- `tsc --noEmit` — no errors in desktop sources. One pre-existing error remains on the base branch, in `expireDeviceLink.saga.ts`; it is untouched here and it also blocks `build:renderer:prod` on `design/onboarding-impl`.
- Full desktop jest suite green: 102 suites, 304 passed, 127 snapshots. Snapshots regenerated deliberately.
- eslint and prettier clean on every changed file, and `tsc --noEmit` in `packages/e2e-tests` compiles.
- Frozen static Storybook, screenshotted and compared against the Figma exports: V1, unread, without-create-permission, the dark-theme guard, and hover on a channel row, a member row, the "Add members" row, the section (+), the community header and the profile row.
- Real app under xvfb, community created over the local transport. The measured column matches the nodes: 220 wide on `#511974`; the community row 220x28 with 0/16 padding; rows 26 high with 3/16 padding, label 14/20/400 at 0.7; section titles 14/20/500 at 0.7; the selected row `rgba(255,255,255,0.1)`; no search field; profile row 48 high with 12/16 padding at 90%; 32 between sections.

https://claude.ai/code/session_012s2ctGRnDDQ6wYcE3wkdEu


holmesworcester:
## Brought up to date with `design/onboarding-impl` (root `2d4f0c091`)

Head is now `78170cfaf`. Base is unchanged (`design/onboarding-impl`); GitHub reports `MERGEABLE` / `CLEAN`, and `git merge-tree --write-tree origin/design/onboarding-impl origin/design/desktop-sidebar` exits 0.

This was a real three-way merge, not a fast-forward: develop rewrote the sidebar at the same time this branch did. develop turned the "Users" section into a real **Direct messages** section (DM channel rows, group avatars, presence by user id, its own `+`) and renamed `UserProfileListItem` → `DirectMessageListItem`. This branch rebuilt the same column on the design library's primitives. 13 paths conflicted.

### Resolution log

| Path | Resolution |
| --- | --- |
| `ui/Sidebar/SidebarHeader.tsx` | Library header kept; develop's per-section test id adopted. `actionTestId` is gone — the component takes `actionTitle` and emits `sidebar-button-${actionTitle}`, which is what develop's rtl tests, `e2e-tests/src/types.ts` and `selectors.ts` expect now that more than one section is addable. |
| `ui/Sidebar/SidebarRow.tsx` | New optional `annotation` slot, so a DM row puts "you" beside the name while the unread badge keeps the right edge. |
| `DirectMessagesPanel/DirectMessageListItem.tsx` | develop's DM row rebuilt on `SidebarRow`. Keeps develop's `ProfilePhotoWithBadge`, group member count and "you"; takes geometry and hover/selected overlays from the library. |
| `DirectMessagesPanel/UserProfileListItem.tsx` | Deleted — develop deleted it. Intent re-applied in `DirectMessageListItem` above. |
| `DirectMessagesPanel/DirectMessagesPanel.tsx` | develop's DM section, in this branch's plain block rather than a flexing `Grid item`, so row overlays still span the 220px column. |
| `ChannelsPanel/ChannelsPanel.tsx` | This branch's markup + develop's `actionTitle` + develop's `ChannelType` filter, so a DM is not listed twice. |
| `ChannelsPanel/ChannelsListItem.tsx` | This branch's `SidebarRow` version. develop's only change there (channel labels at weight 400, not 300) lands by construction: `SidebarRow`'s label uses the `body2` token, which *is* 400, instead of overriding the weight. |
| `UserProfilePanel/UserProfilePanel.tsx` | This branch's version. See "design vs behaviour" below. |
| `Sidebar.stories.tsx` | This branch's typed args and flush-column harness + develop's `DirectMessages` and `DesignReference` stories. |
| `e2e-tests/src/selectors.ts` | develop's DM-list selectors, with this branch's element-agnostic xpaths re-applied to them and to develop's rewritten `addNewChannel`. `switchTab` keeps develop's logging **and** this branch's close-the-open-drawer guard. |
| `Main.test.tsx`, `ChannelsPanel.test.tsx`, `IdentityPanel.test.tsx` | Inline snapshots regenerated against the merged tree. |

`SettingsComponent.tsx` and `channel.switch.test.tsx` were auto-merged; both were spot-checked and correct.

### Design vs behaviour choices

- **Profile summary hover.** develop moved this row onto the 5% `sidebarHover` token "so the other rows use the same wash", after fixing a rounded MUI Button that made the hover a floating pill. The library gives `Profile summary` (`4716:7702`, Hover=True) a **10%** hover deliberately, unlike a list row's 5%, so it keeps its own value. develop's actual defect is fixed by construction here: this is a plain unrounded button, so the wash spans the column.
- **Unread DM.** Screenshotting the merged column showed the two kinds of row disagreeing: an unread channel goes bold *and* takes the library's `badge2`, while develop's DM row went bold only — a thin cue, and the only one, for a state the badge exists to carry. The DM row now takes the same badge, keyed `<channelId>-dm-link-unread`.
- **Stale audit note.** `RealScreens.stories.tsx` said the sidebar's channel rows were "14px at weight 300 with hardcoded 3px vertical padding — to migrate onto the tokens". This branch performed that migration, so the note described a state that no longer exists. Corrected.

### Verified

- `tsc --noEmit` on `packages/desktop` and `packages/e2e-tests`: **exit 0** (after building types → common → logger → state-manager).
- Desktop unit tests: **114 of 115 suites pass, 416 tests, 0 failures**.
- Snapshots: 2 inline snapshots regenerated (`Main.test.tsx`, `ChannelsPanel.test.tsx`). Both change because this branch intentionally changes the sidebar's DOM and the merge adopts develop's `sidebar-button-*` ids; the regenerated output contains develop's DM section *and* this branch's `SidebarRow` classes, which is the point. `SidebarHeader.test.tsx.snap` needed no regeneration.
- New tests, not placeholders: a DM channel is excluded from the Channels list while an untyped channel is not; the section `+` is named after its action and absent without the permission; the unread badge appears and does not appear; "you" annotates only the conversation with yourself; the header gives each section its own test id.
- **Storybook**: static build served on `:8103`, 35 stories captured with Electron 44 under xvfb and looked at individually — all six `Components/SidebarComponent` variants, plus `Components/Main`, `Components/Settings`, `Components/Onboarding`, `Components/Existing`, `Components/DirectMessages`, `Components/LoadingPanel` and the whole `Private channels` group. No story errors, no regressions. The V1 column, the padlock/`#` glyphs, the unread badge, the DM presence dots, the group member counts, the "you" annotation and the ellipsis on long DM titles all render as intended.

### Not verified

- `src/main/main.test.ts` cannot run in this worktree: `backend-bundle`'s `bundle.cjs` is a webpack artifact that is not built here. The merge leaves `packages/desktop/src/main/**` byte-identical to the root branch, so this is environmental, not a merge effect.
- No desktop e2e run. `selectors.ts` changes are matched against what the merged app renders, not proven by a run.
- The dark-theme sidebar story remains explicitly undesigned ("No dark mode yet", `6222:13638`); it is there to catch breakage, not to review.

holmesworcester:
## Correction: `src/main/main.test.ts` does run and pass

I reported it as unverified because `backend-bundle`'s `bundle.cjs` was missing. It is a webpack artifact, and bootstrapping the workspace produced it — after which the suite runs. On the sibling branches in this stack, which share `packages/desktop/src/main/**` with this one byte for byte, the full desktop suite is green including that file.

Nothing else in that comment changes: this head's own run was **114 of 115 suites with `main.test.ts` as the only skip-by-failure**, and it fails for that environmental reason alone.

PR #3530, opened 2026-09-14
state-manager: narrow activeInvite before reading its second field

holmesworcester:
One line, no behaviour change. Split out of #3526 so the sidebar work does not carry an unrelated fix.

## What

```
-  if (activeInvite?.id !== deviceLinkInvite.id || activeInvite.expiresAt !== deviceLinkInvite.expiresAt) return
+  if (!activeInvite || activeInvite.id !== deviceLinkInvite.id || activeInvite.expiresAt !== deviceLinkInvite.expiresAt)
+    return
```

The optional chain does not narrow `activeInvite` for the second clause, so the compiler reports `TS18048: 'activeInvite' is possibly 'undefined'`. Testing `!activeInvite` first says the same thing in a form it can follow.

## No behaviour change

`InviteResult.id` is required (`id: Base58`), so when `activeInvite` is undefined the first comparison is `undefined !== <a base58 id>`, which is true, short-circuits and returns. The second clause was never reached, and there was never a runtime error to fix. `expireDeviceLink.saga.test.ts` passes identically before and after, 4 tests both ways.

## Why it went unnoticed

It reproduces in some checkouts and not others, which is easy to get backwards. Two of us got it backwards in opposite directions before measuring it.

Asked through the compiler API, in two checkouts of the same commit with the same TypeScript 4.9.5:

| expression | errors here | clean here |
| --- | --- | --- |
| `deviceLinkInvite.id` | `Base58` | `Base58` |
| `activeInvite?.id` | **`any`** | **`Base58 \| undefined`** |
| `activeInvite`, first clause | `DeviceLinkInvite \| undefined` | `DeviceLinkInvite \| undefined` |
| `activeInvite`, second clause | `DeviceLinkInvite \| undefined` | `DeviceLinkInvite` (narrowed) |

One cell is the whole story. TypeScript narrows `a` from `a?.b === x` by reasoning about the optional chain's own type; where that type is `any` it learns nothing, `activeInvite` is never narrowed, and the second clause errors. The auth types resolve in both checkouts — `InviteResult.id` really is `Base58` on both sides — so this is not an unresolved import, and it is not the `preserveSymlinks` flag either; that was checked and makes no difference.

What differs is that the selector-side `DeviceLinkInvite` carries `id: any` in one checkout and `id: Base58` in the other, while the direct `@quiet/types` import carries `Base58` in both. The type reaches this file by two routes and they disagree — the same class of problem as the "every path must reach ONE `3rd-party/auth`" rule in the build notes, one level up at `@quiet/types`.

Which is the argument for this change rather than against it: `!activeInvite` does not depend on either comparand's type, so the file stops caring which route the type arrives by. Today a checkout compiles or does not for reasons that have nothing to do with this code, and in one where it does not, this single error is enough to stop `npm run build:renderer:prod` in `packages/desktop`.

## Verification

In a worktree whose auth submodule is built, `tsc --noEmit -p tsconfig.build.json --preserveSymlinks` in `packages/state-manager`:

| | errors |
| --- | --- |
| before | 3, including the `TS18048` |
| after | 2 |

The two that remain are a pre-existing unresolved `3rd-party/auth/packages/crypto/dist` import in two test files, untouched here. `prettier --check` and `eslint` are clean on the changed file.

https://claude.ai/code/session_012s2ctGRnDDQ6wYcE3wkdEu


holmesworcester:
Closing in favour of https://github.com/TryQuiet/quiet/pull/3637, which is this same commit on `develop`.

Not moot: `design/onboarding-impl` was merged with `develop` today and took develop's `expireDeviceLink.saga.ts`, but develop's copy still has the un-narrowed condition, so the fix is still needed. It just does not belong in the onboarding stack — it is a plain typecheck fix on a saga develop owns.

Why a new branch instead of retargeting this one: this branch's merge base with `develop` is f95d8cdb8, so a base change alone would have shown 59 commits of onboarding work in the diff, and repointing the head branch at `develop` would have needed a force push. The cherry-picked branch `fix/expire-device-link-narrowing-develop` gives a two-line diff against `develop` with no rewritten history.

PR #3539, opened 2026-09-15
Avoid repeated message-history processing

holmesworcester:
Quiet repeatedly scanned and consumed old message history as new messages arrived. This change indexes verified message IDs/hashes and avoids repeatedly searching the frontend message list. TryQuiet/auth#39 separately supplies the accompanying cache of immutable cryptographic facts. Message signatures and current authorization are still checked.

The index handles OrbitDB's actual replication behavior: one head update can import an entire backlog. It discovers newly accepted ancestors once using canonical local-log reads. Authorization changes invalidate the index; unjoined heads wait for local acceptance. A pre-existing delayed-notification race after authorization changes or close is also fixed.

Seven QSS roster checks now assert visible members instead of waiting for an unasserted Tor status. Existing offline-status, message, private-channel, and history assertions remain intact.

This PR contains no native crypto adapter, sodium loader, crypto test hooks, or new crypto implementation. Backend package.json and webpack configuration match `10.0.0`. Native OpenSSL work is preserved independently in TryQuiet/quiet#3541; this PR can be used with the existing libsodium implementation or a future native implementation.

Validation after separating native crypto:

- Actual desktop QSS: all 168 passed with zero skips on the fresh package without the adapter: 64 multiplayer, 75 private-channel, and 29 single-client tests.
- Backend: 896 standard tests, 12 Tor tests, and the large-file test passed.
- Auth: 773 passed (87 existing skips). State-manager: 205 tests and 13 snapshots passed.
- Native-free backend build/lint passed. E2E helper tests: 14 passed; E2E build/lint passed.
- Real OrbitDB regression cases cover public/private backlogs and rejection of an impersonating ancestor. The concurrent scaling test processes 1,000 shared ancestors plus 100 heads with 1,100 consumes and no completed-history rescan.

The first public-QSS run passed 61/64: restoring P2P while Tor was still bootstrapping returned false and left the settings modal open, causing two dependent UI failures. The same build and unchanged suite then passed 64/64. The involved Tor-toggle code matches the release base; the initial failure is retained in the validation record. Existing test skips remain unchanged. Daybreak Blue accepted the processing-only split after 73 backend and 8 independent auth adversarial tests using standard sodium. No introduced signature-verification, impersonation, or privilege-escalation regression was found.

Fixes #3536. Advances #3537. Depends on TryQuiet/auth#39; targets `10.0.0`. Graph traversal and contextual replay still grow with sigchain history. The earlier combined iPhone timings are not measurements of this branch alone.

Issue #3538, opened 2026-09-15
Accelerate iOS Node backend crypto without changing validation semantics


holmesworcester:
## Problem

The embedded iOS backend spends most of its active time executing libsodium's JavaScript fallback. This multiplies the cost of repeated validation and leaves significant delay even after reuse of known keys. Address this in parallel with the higher-priority message-history complexity work.

This message path runs in **Node Mobile on a separate native thread**, not the React Native WebView. Migrating the frontend from JSC to Hermes (#3462) does not by itself replace this backend runtime.

## Runtime and measured evidence

Physical iPhone 16e / iOS 18.5, published Quiet 10.0.0-alpha.0 (build 613):

- Node Mobile 18.20.4, V8 10.2.154.26-node.37.
- `typeof WebAssembly === "undefined"`; matching Node Mobile iOS build documentation specifies JIT-disabled V8.
- Startup CPU profile: 94.4% of active sampled time inside libsodium's JavaScript fallback.
- Actual ordinary-member decrypt/verify: 0.416 s for one message, 47.420 s for 100, 541.801 s for 1,000. All batches completed and verified exact contents/signatures.
- 1,000 actual message-signature checks consume about 48.4 s; symmetric message decryption only 1.48 s. Reconstructing/validating keypairs adds hundreds of seconds before caching is fixed.
- Static-team control reusing a validated role key: 1,000 messages in 57.451 s, preserving every signature check. This yielding control includes incidental live-community auth work (11 extra signature checks and 2.63 s of key lookup); no time was subtracted.
- Timers disabled versus enabled: 5.598 s versus 5.600 s for ten messages, excluding profiler overhead as the explanation.

Same release payload/fixtures on Linux: approximately 1.07 ms/message with ordinary Node/WASM, 24.1 ms without WASM but with JIT, 401.8 ms with `--jitless`. These are runtime controls on different hardware, not matched desktop/mobile UI benchmarks.

## Implementation work

- [ ] Evaluate native Ed25519/X25519 operations already available through Node Mobile's OpenSSL, or a compatible native libsodium binding. Prefer a supported implementation with a reliable mobile build/package path.
- [ ] Preserve exact key/message formats and strict validation semantics. Native OpenSSL and libsodium verification behavior must be compared on malformed/noncanonical encodings, scalar bounds and small-order points; do not silently broaden acceptance.
[...]

PR #3541, opened 2026-09-15
Use Node OpenSSL for iOS backend curve operations

holmesworcester:
iOS Node Mobile runs libsodium's JavaScript curve operations without JIT or WASM. This adapter uses the OpenSSL already shipped in Node for six Ed25519/X25519 operations, preserving sodium wire formats, output behavior, and rejection policy.

This is the native-crypto portion split out of #3539. It is kept as a draft alternative while a native libsodium implementation is being developed. It targets `10.0.0` directly, retains that branch's existing auth submodule revision, and has no dependency on the indexing or LFA caching changes.

The change contains only the iOS-gated adapter, webpack loader, documentation, and crypto tests. Capability/version checks retain the original implementation on unsupported runtimes; exceptional Ed25519 encodings still undergo sodium verification. Other platforms keep their existing implementation. No QSS release or data migration is required.

Validation: all 13 native differential/webpack tests pass on Node 18.20.4 and 20.20.1. Coverage includes 151 Ed25519 and 518 X25519 Wycheproof cases, tampering, malformed keys, atomic initialization/fallback, and real webpack activation across five platform identities. The adapter and tests are byte-for-byte the previously audited final implementation. Daybreak Blue accepted the native-only split with no concrete findings; its independent 16-test adversarial suite also passed. The earlier combined iPhone performance results do not measure this branch alone.

Related to #3538. The processing changes remain in #3539 and TryQuiet/auth#39.


holmesworcester:
We'll use libsodium natively in a later fix.

PR #3551, opened 2026-09-16
Merge 10.0.0 (and 9.1.0) into develop

holmesworcester:
Merges `10.0.0` into `develop`. These items were approved for release even though we
didn't cut a full release this time.

This also brings across **9.1.0**, which had never reached develop — 224 commits, up to
`ae16fd1a3 "feat: enable private channels without a feature flag"`. Before this PR, none of
9.0.0, 9.1.0 or 10.0.0 was an ancestor of develop.

### Why this isn't a plain `git merge`

`1ed08d8fb "Merge 9.0.0 into develop (#3459)"` landed as a **squash**, so it has a single
parent and git no longer records 9.0.0 as an ancestor of develop. The merge base git computes
for `develop..10.0.0` is therefore `00de37248` (2026-07-27), about two months stale.

Merging against that base produces 46 conflicted files, and — the part that doesn't announce
itself — for the files that *don't* conflict it treats develop's 9.0.0-era content as edits to
replay on top of 10.0.0, silently reverting 9.1.0 and 10.0.0 fixes. `-X theirs` doesn't help;
it only settles hunks that already conflict.

Since 9.0.0 *is* an ancestor of 10.0.0, 9.0.0 is the honest base. So this merge was built by
three-way merging only the real develop delta (`git diff 9.0.0 develop` = 28 files) onto the
10.0.0 tree. The result is 10.0.0 plus a 24-file develop-only delta; nothing in
`packages/desktop` or `packages/backend` differs from 10.0.0.

25 of the 28 merged clean. The three judgement calls:

- `packages/mobile/ios/TorBackgroundTransitions.swift` — develop-only, taken wholesale.
- `packages/backend/.../integration-tests/two-peers.spec.ts` and its snapshot — took **10.0.0**.
  Both sides fixed the same flake; 10.0.0's `captureTimelines()` version supersedes develop's, and
  the snapshot keys are identical on both sides.
- `packages/mobile/ios/TorHandler.swift`, `packages/mobile/android/app/build.gradle` — merged
  cleanly but reviewed by hand since both sides had touched them.

### Tests

`common` 69/69, `state-manager` 204/204, `mobile` 168 passed (10.0.0 baseline is 166; the two extra
are develop's Appbar a11y tests). One mobile suite (`UpdateChannelMembership`) fails identically on a
pristine 10.0.0 worktree — it's a local `react-native-paper` native-module invariant, not a merge defect.

### Please merge this with a merge commit, not "Squash and merge"

If this squashes, the next release→develop merge hits exactly the same stale-base trap for 11.x.
`2e0a9cade` adds that to `PUBLISHING.md`, which had nothing to say about the merge method.


holmesworcester:
Two follow-ups after running the failing checks locally on a faithful environment (fresh worktree, real `npm ci` + `lerna bootstrap`, submodules at the pinned commits, same AppImage recipe as CI):

**`linux` / `mac` e2e — `multipleClients.privateChannels.test.ts`: flake, not a defect.** Locally it passes **100/100 on the first attempt** (after `oneClient` 31/31 as the gate). On CI both jobs got 97–99/100 per attempt with a *different* handful missing each time — file-download visibility and the re-added-member stages, all 40–60 s propagation waits over real Tor on a 4-core runner. The test file is byte-identical to 10.0.0, and this workflow has 0 successes in its last 60 runs, so it's not something this PR can fix by retrying.

**`regression-tests` — `Channel.cy.tsx` viewport-660 PageUp/PageDown: a real race in 10.0.0, fixed here as `04bde8f2a`.** `ChannelComponent` re-pinned the list to the bottom on *every* container resize. The resize observer fires asynchronously, so a PageUp that landed between the resize and the callback was undone. It fails deterministically on the pure 10.0.0 tree locally and passes/fails on timing in CI. The fix re-pins only when the reader was already at the bottom, tracked by a ref that only the reader's own scrolling moves (scroll events within 300 ms of the component's own `scrollTo` are ignored for it — a first attempt that used the last *measured* position broke the 400 px cases, because the mount-time scroll event was evaluated after the viewport had already shrunk). Verified 4× consecutive 14/14 on the Cypress spec and 25/25 Channel jest suites.

This is the one commit on the branch that is not "10.0.0 + the develop delta"; happy to move it to its own PR if you'd rather keep this one pure.

Issue #3424, opened 2026-09-09
Community owner is not recognized as one


kingalg:
**Version**: 9.0.0-alpha.17
**System**: MacOS

**How to reproduce**: I don't know yet; I haven't managed to reproduce this bug.

**Issue**: A user on macOS was a member of a Tor community. The user left and created their own Tor community. When I open an invite tab on the new community - the user who created it was not recognized as the owner. Logs and screenshot below.

[Quiet9.zip](https://github.com/user-attachments/files/32017375/Quiet9.zip)

In the screenshot below, you'll see that this community has only one member. It's the user who created it, but it's impossible to invite new users.
<img width="802" height="551" alt="Image" src="https://github.com/user-attachments/assets/39755653-6ba1-49eb-8680-f46a00503324" />

PR #3552, opened 2026-09-16
Merge device linking into develop (with the 10.0.0 release)

holmesworcester:
Puts `feat/2610-device-linking` (@adrastaea, 70 commits) on top of develop now that 10.0.0 has
been merged into it.

**Stacked on #3551** — base is `merge/10.0.0-into-develop`, so the diff here is just the
device-linking delta. Land #3551 first (with a merge commit).

The original branch is untouched. Merge base with it is honest (`1ed08d8fb`), so git's own
three-way merge was usable here; all 19 textual conflicts are resolved. Where the two branches
refactored the same code incompatibly, **the released 10.0.0 behaviour is kept and the
device-linking feature is re-expressed on top of it.**

Three conflicts were semantic rather than textual — git merged the text cleanly but the two
sides disagreed about behaviour. Those, and the smaller calls, are written up in a conversation
comment below. **@adrastaea, that comment is the review surface** — particularly the onboarding
supersede-vs-ignore decision, which changes the behaviour `ade57a0f5` introduced, and the
failed-leave guard on `linkDevice`, which is an addition rather than a merge.

### Tests

`backend` 87 suites / 1113 tests, `desktop` 103 / 305, `state-manager` 63 / 252, `mobile` 197,
`common` 69 — all passing. (Backend was 31 failing before the fixes described in the comment.)

Two of those failures were defects in my own resolution rather than conflicts, both caught by
10.0.0's tests and attributed against a pristine 10.0.0 worktree; they're described in the comment too.

### Submodules

No submodule PRs needed. `3rd-party/auth` → `6f534c89` (10.0.0's pin; device-linking's `f09a50c0`
has a tree identical to develop's base pin, which `6f534c89` already contains).
`3rd-party/qss` → `0c80fc3`, the existing upstream `feat/device-linking-10.0`, which already merges
both lines and itself pins auth at `6f534c89`.


holmesworcester:
## Merge notes: `feat/2610-device-linking` → develop (with 10.0.0)

Base is `merge/10.0.0-into-develop`. 19 conflicts. Where the two branches refactored the same
code in incompatible ways I kept the **released 10.0.0 behaviour** and re-expressed the
device-linking feature on top of it. Everything below is worth a second opinion from the
device-linking author.

### Undeclared semantic conflicts (git did not mark these)

1. **Onboarding supersede vs ignore** — `communities.master.saga.ts`.
   10.0.0 cancels an in-flight onboarding task when a new request arrives; `ade57a0f5`
   ("preserve failed-admission cleanup across restart") changed that to ignore the new request.
   Keeping "ignore" broke three released join-recovery tests (the replacement invitation never
   reached the backend). Kept 10.0.0's supersede, kept device-linking's admission gate — which is
   the actual mechanism that commit needed — and rewrote its "ignores overlapping onboarding" test
   as "supersedes an in-progress onboarding request", adding a separate test that a *replayed*
   join for an already-running attempt is still dropped.

2. **Auth sessions: per-peer vs per-transport** — `libp2p.auth.ts` (12 hunks).
   10.0.0 re-keyed `authConnections`/`peerConnections` by `connection.id` so each transport owns
   one LFA session (message numbers are per-session). Device-linking kept per-peer keying and
   layered admission gating on it. Kept the per-transport keying and ported the gating onto it:
   the gate contexts live in a `WeakMap` keyed by the `Auth.Connection`, so they were mostly
   orthogonal. Two liveness checks in device-linking's code looked sessions up by peer id and
   silently found nothing — fixed to `connection.id` (this is what made `redialPeers` stop firing).
   Dropped device-linking's per-peer "replace the stale auth connection" block: 10.0.0 deliberately
   removed it, and closed transports are now retired by the `connection:close` handler.

3. **Peer addressing: user profiles vs device endpoints** — `storage.service.ts`.
   Device-linking removed `userData` from `UserProfile` outright and moved addressing to a
   per-device `NetworkEndpointsStore` (one address per user no longer works once a user has
   several devices). So the endpoint enumeration wins, and 10.0.0's profile-driven
   `updatePeerStore` is gone. 10.0.0's authenticated-identity filter is a security hardening and
   survives: an existing peer address is only carried over when its secured-session binding still
   names a current device of a current member. Added a removal pass for endpoints whose device has
   left the team, and made re-admission via an endpoint drop a stale session binding rather than
   copy it. `peerStore-recovery.spec.ts` ported from profiles to endpoints.

### Smaller calls

- **One mutex, not two** — `connections-manager.service.ts`. Both branches added a mutex
  (`communityTransitionMutex` for the #3424 leave race, `admissionMutationMutex` for device
  linking). Two locks would let a leave interleave with a create, reintroducing that race.
  Collapsed to `admissionMutationMutex` and kept 10.0.0's `requireCompletedLeave()` guard.
  Also applied that guard to `linkDevice`, which provisions community state exactly like join —
  **this one is an addition, not a merge; drop it if you disagree.**
- **CI workflows** — kept device-linking's policy (`LOCAL_TRANSPORT=true`, one run, no
  `nick-fields/retry` wrapper, as its README section documents) and 10.0.0's test coverage
  (`selectors.test.ts`, `compatibilityBaseline.test.ts`, no `major-ver` guard on back-compat).
- **libp2p.service.ts** — 10.0.0 tears down the failing *transport* rather than the peer address;
  device-linking's grace delay and admission-gate check now wrap that call. Its
  `redial && !peerAddress.startsWith('/p2p/')` suppression moved into `hangUpAuthTransport`,
  where address resolution now lives. Dropped the `preSharedKey` import (unused after 10.0.0).
- **Submodule pins.** `3rd-party/auth` → `6f534c89` (10.0.0's pin). Device-linking's pin
  `f09a50c0` has a tree identical to develop's base pin `742ea3dad`, which `6f534c89` already
  contains, so nothing is lost. `3rd-party/qss` → `0c80fc3`, the existing upstream
  `feat/device-linking-10.0`, which already merges both lines and itself pins auth at `6f534c89`;
  the only qss commits it lacks are three auth-pin bumps that `6f534c89` supersedes.
  **No submodule PRs are needed.**
- **Test-only adjustments:** `InvitationKind.Member` added to 10.0.0-era invitation expectations;
  `LocalUserContext` built explicitly where the chain context is now a wider union that includes
  an invitee *device* context with no `user`; the auth spec's `components` mock gained `events`
  and `connectionManager`, which 10.0.0 requires.

### Two defects in my own resolution (not conflicts)

Both surfaced as backend failures and were attributed by re-running the same suites on a pristine
10.0.0 worktree, where they pass.

1. **`libp2p.auth.lifecycle.spec` timed out at 80s.** Device-linking refuses auth connections unless
   `libp2pService.state` is running, but `Libp2pState` *defaults* to `Stopped`, and that spec drives
   `Libp2pAuth` directly without starting the service — so nothing ever connected. Fixed in the
   harness (`services.forEach(s => (s.state = Libp2pState.Started))`), not by weakening the guard.

2. **A create queued behind a slow leave was refused.** Device-linking's `rejectWhenCommunityExists`
   has no equivalent in 10.0.0, which erases previous community artifacts inside create. Scoped it to
   an *actually active* community — `communityState === DEFAULT && communityId === ''` means a
   completed leave, so the stale records left behind don't count. Nothing tests that guard, so it is
   worth a look.

### Reproducing the test run

`packages/state-manager` must be rebuilt (`npx tsc -p tsconfig.build.json --preserveSymlinks`) after
every source edit, or the mobile suite silently tests the stale `lib/`. That cost me a wrong diagnosis.

holmesworcester:
Two more merges onto this branch since the notes above:

- `6dd1a0db5` (taea) — restores the root `package-lock.json`; every CI job here had been dying at `npm ci` with `Missing: unicode-emoji-json@0.8.0 from lock file`.
- `04bde8f2a` via the base branch — fixes the `regression-tests` Cypress scroll failure. It's a resize/PageUp race in 10.0.0's `ChannelComponent`, not device linking: it reproduces on the pure 10.0.0 tree locally. Details on #3551.

Nothing in the device-linking resolution changed.

holmesworcester:
Folded taea's four 2026-09-16 commits from the original `feat/2610-device-linking` branch into this PR (merge commit 2fe0c2b63), so the original branch is now fully contained here and can be deleted after the squash.

Conflict resolution:
- `3rd-party/qss`: both sides pointed at qss branches that have since been folded into qss main (taea's pin was #51 itself; the old pin was the pre-#51 device-linking line). Pinned the current qss main head 6c8849b, which contains #51 and the 3.0.0 merge-back (qss#61).
- `CHANGELOG.md`: kept both the [unreleased] device-linking entry and the [10.0.0] section.

Squash-merging this PR is fine (feature branch, not a release merge-back). Please delete `feat/2610-device-linking` right after, so no one merges from it again.

PR #3555, opened 2026-09-16
fix(desktop): ignore layout-induced scroll events when tracking the reader position

holmesworcester:
Follow-up to the resize/scroll fix (04bde8f2a) that landed with #3551. The Cypress `Channel.cy.tsx` scroll tests were still flaky: 4 of 6 local runs failed on develop (`f63d3d790`) and the regression-tests job failed on #3552's last run.

**Root cause** (from instrumenting the component): a viewport resize makes the browser itself move `scrollTop` (scroll anchoring / clamping), and that scroll event can be delivered a few ms *before* the ResizeObserver callback. The old 300 ms "programmatic scroll" window classified it as the reader scrolling away from the bottom, so `onResize` declined to snap. The window was also too long the other way: a PageUp typed right after a resize was taken as ours.

**Fix**: a scroll event is ours if it lands on the last position we scrolled to; and it is ignored for reader tracking when the container's `clientHeight`/`scrollHeight` changed since we last looked (the signature of a layout-induced scroll). The geometry baseline is recorded on every programmatic scroll and resize, since under slow rendering no scroll event is delivered between mount and the first resize.

**Verification**: `Channel.cy.tsx` 10/10 local runs pass (baseline 2/6). Desktop Channel jest suites pass.

Same commit is also opened against `11.0.0`.


holmesworcester:
Superseded by #3568. This commit is byte-identical to `1c4c62801`, which landed on 11.0.0 as #3556 and reached develop through the 11.0.0 back-merge — verified with a diff of both commits restricted to `packages/desktop/src`. Closing rather than merging so develop keeps one copy of the fix.

PR #3569, opened 2026-09-17
Retry channel metadata entries that couldn't be decrypted yet

### LINKED ISSUE #3563: Private channel entry can stay unindexed forever when the role key is missing
A private channel's metadata entry is re-indexed only when something else happens. If a device is a member of the channel's role but is missing that role's key, the entry stays unindexed and there is no path back to it on its own.

## The recoverable case (works today)

`decryptChannelEntry` rejects an entry up front when we are not a member of the scope's role (`packages/backend/src/nest/storage/channels/channels.service.ts:313-320`), `validateChannelMetadataEntry` swallows `NotAMemberError` and returns `false`, and `keyValueIndexedValidated` logs `Invalid entry detected: <hash>, skipping indexing` without writing the hash to `indexedEntries` (`packages/backend/src/nest/storage/orbitDb/keyValueIndexedValidated.ts:108-123`). Because only valid entries are marked indexed, the entry is revisited on a later pass.

Becoming a member of the role is itself a sigchain update, and `handleSigchainUpdated` reindexes both metadata stores on `SigchainEvents.UPDATED` (`channels.service.ts:70-82`, wired at `:189`). So the event that fixes the precondition also triggers the retry. Observed working on 11.0.0-alpha.0: three clients sat at `Not a member of this channel, skipping channel entry decrypt` for ~2m18s and all decrypted the channel within ~1s of the chain update that granted the role.

## The case that does not recover

If `chain.roles.amIMemberOfRole(...)` passes but the role key is missing, `chain.crypto.decryptAndVerify` throws something that is not `NotAMemberError`. `validateChannelMetadataEntry` logs `Failed to validate channel entry` and returns `false` (`channels.service.ts:706-712`), the hash stays out of `indexedEntries`, and the channel never appears for that device.

The retry is entirely event-driven — `retryIndexingUnindexedEntries()` is called only from `handleSigchainUpdated` (`:76-77`) and `handleChannelMetadataUpdate` (`:247`). There is no timer or periodic sweep in either file. So recovery depends on some *unrelated* future sigchain or channel-metadata event. In a quiet community there may not be one, and restarting does not help: reopening the store replays the same log against the same missing key.

This is reachable in the current build. LFA's reducer drops key material out of a link that also conveys role membership:

```
WARN localfirst:...:decryptTeamGraph:reducer Dropping 1 lockbox(es) for ROLE:6xdLFyK8aZmwAv8oVVYCToG9rab5zs88ym8cucaUWv9jbw4wmbs63ahzScFM77gT
  from link DsBpuTBb21UHxcAdVPBQ7aBWX6nK4oD4ow6MAp6aYYfF: ADD_MEMBER_ROLE may deliver existing keys of that role only to the assigned member
```

Every client in a six-instance test logged that warning. None of them ended up short of a key — `Failed to decrypt channel entry` appears 0 times across all six logs — so this is a hole in the recovery path rather than an observed failure.

## Suggested fix

1. Treat a decrypt failure the same as `NotAMemberError` for retry bookkeeping, so the two failure modes cannot diverge.
2. Add a retry trigger that does not depend on new chain activity — e.g. call `retryIndexingUnindexedEntries()` on auth reconnect. Those connections cycle often, so it costs little and bounds the stall.

Adjacent, different layer: #3559 retries the heads exchange once membership authentication lands, i.e. gets the entries to arrive. This issue is about an entry that has already arrived and been rejected by the index.

Found on 11.0.0-alpha.0 (`@quiet/desktop@11.0.0-alpha.0`), code read on branch `11.0.0`.

holmesworcester:
Fixes #3563.

A private channel's metadata entry is deliberately left unindexed when it fails validation, so that a later pass can revisit it. The only things that start a later pass are `SigchainEvents.UPDATED` and a channel metadata update. That covers the common case — becoming a member of the role *is* a sigchain update — but not a device that is already a member of the role and is missing the role's key. There is no timer and no periodic sweep, so recovery waits on unrelated community activity that may never come, and restarting doesn't help: reopening the store replays the same log against the same missing key.

## What changed

**The two failure modes converge.** `decryptChannelEntry` now throws `MissingChannelKeyError` when we *are* a member of the scope but `decryptAndVerify` fails — that can only be key material we don't have yet. `isChannelEntryDecryptPending` treats it exactly like `NotAMemberError`, and the three places that discriminated on `NotAMemberError` (metadata validation, `getChannel`, `getChannels`) now call that one helper. Previously a missing key was logged as `Failed to validate channel entry` and a missing membership as something ignorable, with no difference in outcome — the two could drift apart with the next edit.

**Recovery no longer depends on someone else doing something.** Rejecting an entry for either reason arms a reindex one minute out (`UNDECRYPTABLE_ENTRY_RETRY_INTERVAL_MS`). The sweep clears the pending flag before running, so validation re-arms it only while something is genuinely still undecryptable; when everything indexes, the sweep stops. A community with no missing keys never schedules a timer at all, and the timer is `unref`'d and cancelled on `closeChannels`/`clean`.

Entries that are simply invalid — wrong store, forged metadata, non-admin delete — don't set the flag, so a bad entry can't keep a sweep running forever. The one case that can is an entry whose signature or ciphertext is corrupt in a way `decryptAndVerify` rejects: that looks the same as a missing key from here, and will be re-examined once a minute. The cost is one traversal of a metadata store (one entry per channel) per minute, and the access controller already rejects unauthorized writes, so this is a narrow window.

## Tests

`channels.service.spec.ts`, new describe block `Undecryptable channel metadata retries`:

- schedules a retry when a channel entry we belong to can't be decrypted
- **does not** schedule one for an entry that is simply invalid
- **reindexes channel metadata on its own, with no sigchain or metadata activity** — the sigchain listener is detached, the entry fails to decrypt, the key "arrives", and a minute later both stores are reindexed and channels rebroadcast; then five more minutes pass with no further sweep
- keeps retrying while entries still can't be decrypted
- stops retrying when the service is closed

The third one is written only in terms of behavior that exists either way. Against `origin/11.0.0` it fails with `Expected number of calls: 1 / Received number of calls: 0` — the reindex never happens. With the fix, all five pass.

## Verification

- `npm test -- channels.service.spec` → 46 passed, 1 skipped
- `npm test -- src/nest/storage` → 372 passed, 2 skipped, 0 failed

An earlier revision of this description reported 2 failures in `LFAIdentityProvider identity binding` as pre-existing. They were an artifact of a stale local build of the `3rd-party/auth` submodule; after rebuilding it, those tests pass. Nothing was failing.
- Prettier clean.


holmesworcester:
## Merging with the three `unit-tests` jobs red

The three `unit-tests` jobs of **Backend unit tests** fail on this PR, and the failure is not a test failure and not caused by this change. Filed as #3634. Recording the evidence here because the merge goes ahead with them red.

### What the job actually did

`unit-tests (ubuntu-22.04)` on https://github.com/TryQuiet/quiet/actions/runs/35658959464 logged 100 `PASS` suites and zero `✕`, then at 350 s:

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
[3965:0x44650000]   352299 ms: Mark-Compact 4010.3 (4136.8) -> 3978.1 (4135.9) MB ... allocation failure
```

`test-ci` runs Jest `--runInBand` under `--max-old-space-size=4096`; the heap is pinned at ~4.0 GB against that 4 GB cap. The crash replaces Jest's summary, which is why the job reports no failing test. `long-running-tests` and `unit-tests-with-tor` passed.

### Not this branch

`fix/windows-app-close` (based on `11.2.0`, and `git merge-base --is-ancestor` confirms it does **not** contain this branch's commit) hits the identical OOM in the same job:

- https://github.com/TryQuiet/quiet/actions/runs/35655871685 — OOM, ubuntu-22.04
- https://github.com/TryQuiet/quiet/actions/runs/35650789583 — OOM, macos-26
- https://github.com/TryQuiet/quiet/actions/runs/35656508425 — **passed**, ubuntu-22.04

The same branch both passing and OOMing on the same OS is the signature of a suite sitting right on the limit rather than a defect in any one change.

### Suite is green locally

Full `test-ci` command on this branch, same 4 GB cap:

```
Test Suites: 2 skipped, 101 passed, 103 total
Tests:       6 skipped, 1255 passed, 1261 total
Time:        361.284 s
```

Zero failures. `channels.service.spec.ts` on its own: 46 passed, 1 skipped.

### A/B against pristine develop

Both runs use identical flags (`--runInBand --ci --silent --verbose --forceExit --logHeapUsage`, `--max-old-space-size=4096`) under `/usr/bin/time -v`, same machine, back to back.

| | `develop` @ `343f2b160` | this branch @ `b608cd72a` |
|---|---|---|
| Test Suites | 101 passed, 2 skipped | 101 passed, 2 skipped |
| Tests | 1250 passed, 6 skipped, **0 failed** | 1255 passed, 6 skipped, **0 failed** |
| Wall clock | 5:21.51 | 5:25.31 |
| Peak RSS | 4,793,340 kB | **4,787,340 kB** |
| Highest per-suite heap | **3,938 MB** | **3,892 MB** |

`develop`'s heap over its final eight suites: `3885 → 3893 → 3907 → 3919 → 3911 → 3925 → 3938 → 3934 MB`, against a 4,096 MB cap.

This branch comes out 6 MB lower on peak RSS (0.1%) and 46 MB lower on the highest per-suite heap. Both differences are within run-to-run noise and both point the wrong way for a leak, so there is nothing here to attribute to this change. The 5 extra passing tests are this PR's new `Undecryptable channel metadata retries` block.

The conclusion is that this branch is indistinguishable from `develop` on memory, and `develop` itself already finishes the run with only ~150 MB of the 4 GB cap to spare. The retry timer this PR adds is one `setTimeout` per `ChannelsService`, `unref`'d, cancelled in `closeChannels()` and `clean()`, and armed only while an entry is genuinely undecryptable — which the numbers bear out.

PR #3591, opened 2026-09-19
Make interrupted device admission recovery deterministic

### LINKED ISSUE #3580: E2E: device-link admission case waits 60s for libp2p start, below the backend's own 120s onion-descriptor allowance
## Symptom

`admissionTimeout.test.ts` → "links a device after reopening the target during an interrupted admission" fails on macOS:

```
Process output for e2e_reopenedlinkeddevice_1mgbfowudwo0
  did not contain "Starting libp2p" within 60000ms

  at BuildSetup.waitForProcessOutput (src/utils.ts:287:11)
  at Object.<anonymous> (src/tests/admissionTimeout.test.ts:382:9)
```

Run: https://github.com/TryQuiet/quiet/actions/runs/35372534309/job/105689943139 (6 of 7 in the suite passed)

## What the backend was doing

This is the one case in the suite that sets `LOCAL_TRANSPORT=false`. Over real transport `launch()` awaits `registerHiddenService()`, which blocks until Tor reports `HS_DESC … UPLOADED` for the device's own onion address — so libp2p cannot start until the descriptor is published.

From the failing run:

```
18:13:14.367  Registering hidden service for community …
18:13:14.367  backend:Tor Spawning Tor hidden service
18:13:14.371  backend:TorControl Sending Tor command { ADD_ONION … }
18:13:14.425  backend:TorControl Tor command response { code: 250 }
              ← nothing for 60s
18:14:14.550  e2e:selectors Closing the app      (test gave up)
```

`ADD_ONION` was answered in 58ms. The `HS_DESC UPLOADED` event had not arrived 60s later.

## Why the budget is inconsistent

`TorControl.sendCommandAndWaitForEvent` allows that upload `TOR_EVENT_TIMEOUT_MS` = **120s** before it gives up. The test allows **60s** (`TOR_LIBP2P_START_TIMEOUT_MS`). So the test can fail while the backend is still working exactly as designed, and never reports the backend's own error.

Measured on a Linux dev box, same code path:

```
00:16:18.826  Spawning Tor hidden service
00:17:54.937  Published hidden service descriptor for onion address bmed37wu5…
```

**96 seconds** — over the test's budget, under the backend's.

## Not just a number to bump

The transport-scaled constants at the top of the file (`TOR_BOOTSTRAP_TIMEOUT_MS`, `TOR_JOIN_COMPLETION_TIMEOUT_MS`, `TOR_LIBP2P_START_TIMEOUT_MS`, `PEER_DIAL_TIMEOUT_MS`, `TOR_ADMISSION_TIMEOUT_MS`) are unexplained magic numbers that don't relate to each other or to the backend's own limits. Raising this one to 180s makes the case pass but leaves the rest arbitrary. Worth deciding what these should actually be derived from — or whether this case needs real Tor at all — rather than tuning it again.

Note the case also deliberately disables QSS (see https://github.com/TryQuiet/quiet/issues/3581), so it exercises the libp2p fallback specifically.

Found while fixing https://github.com/TryQuiet/quiet/pull/3579; deliberately left out of that PR.

### LINKED ISSUE #3581: E2E: withUnavailableQss points at the default QSS port, which is not unavailable when a QSS stack is running
## Problem

`admissionTimeout.test.ts` makes an invitation point at a QSS endpoint that is supposed to be unreachable, so the device falls back to libp2p:

```ts
function withUnavailableQss(invitationLink: string): string {
  const invitation = parseInvitationLink(new URL(invitationLink).hash.slice(1))
  return composeInvitationShareUrl({
    ...invitation,
    version: InvitationDataVersion.v5,
    qssEnabled: true,
    qssEndpoint: 'ws://127.0.0.1:3003',
  } as DeviceInvitationDataV5)
}
```

`ws://127.0.0.1:3003` is the *default QSS port*. It is only unavailable when nothing happens to be serving it.

## What goes wrong when something is

On a machine running the QSS compose stack (`quiet-qss-e2e-*` / `qss-quiet-qss-quiet-1`, which binds `0.0.0.0:3003`) the client connects instead of being refused, and the fallback under test never runs:

```
backend:qss:client  Creating and connecting client socket on ws://127.0.0.1:3003
backend:qss:client  QSS connected! T9Ccc4cX7y62GH7ZAAAZ
backend:qss:service Signing in to community F2NYQ33oZX119FUeTtGpRQ5ZYBGPWbT3cim84t6k31
backend:qss:service Pending invitation admission is coordinated externally; skipping …
backend:qss:service Pausing QSS service
backend:ConnectionsManagerService Failed to launch community … AdmissionError
```

The test then fails for a reason unrelated to what it is asserting, and the failure looks like a device-linking bug.

CI's plain `linux` and `mac` jobs don't run QSS, so this is currently green there — but it makes local runs and any machine with a QSS stack an invalid oracle for this suite, and it would misfire if these tests ever ran alongside the QSS job.

## Suggested fix

Name an endpoint that cannot be served rather than one that usually isn't — e.g. `ws://127.0.0.1:1` (port 1 can't be bound without root, so the connection is refused on every platform), with a comment saying why not the default port.

Found while fixing https://github.com/TryQuiet/quiet/pull/3579; deliberately left out of that PR.

holmesworcester:
The interrupted device-admission test waits 60 seconds for libp2p while real Tor may spend up to 120 seconds publishing this device's onion. It also uses the default QSS port as an allegedly unavailable endpoint, so a local QSS stack changes the behavior under test.

Run admission recovery over real local libp2p sockets with the inviter offline during the first attempt. Wait for dialing after provisional state is stored, restart the target, bring the inviter back, and retry the same invitation. Own an ephemeral loopback endpoint that rejects every QSS connection and assert both attempts hit it. Verify the linked identity and messages in both directions after recovery. The successful retry uses the production admission deadline.

Keep real-Tor device-link/message coverage as a separate Linux release test using the existing two-device case. Remove the recovery test's Tor-specific timeout overrides. Include the endpoint fixture regression in the Linux, macOS, and Windows admission steps.

Closes #3580. Closes #3581.

Validation:

- Packaged Linux app: all 7 admission scenarios and the endpoint fixture pass (8 tests, 454 seconds). The changed interrupted-admission case also passes when run alone.
- Endpoint fixture test proves repeated connection rejection, exclusive ownership while running, and port release after cleanup.
- Separate real-Tor device-link/message case: the initial run passed (88 seconds). A later run with the final app branch failed because the linked device stalled at 55% Tor bootstrap and hit the existing 300-second bootstrap deadline, before publication or dialing. Public-network Tor coverage remains dependent on the environment.
- E2E TypeScript build, lint, formatting, and modified workflow YAML validation pass.

The packaged app also includes the Tor-readiness fix in https://github.com/TryQuiet/quiet/pull/3592. This test change has no source dependency on that fix; local transport already bypasses onion registration. macOS and Windows execution remain for CI.

## Mirrors #3630 on 11.2.0

https://github.com/TryQuiet/quiet/pull/3630 landed the same fix on `11.2.0`, independently, as `packages/e2e-tests/src/unavailableQssEndpoint.ts` + `src/tests/unavailableQssEndpoint.test.ts`. Release branches come back to develop as a merge commit, so two differently named fixtures would both land and neither would conflict. This branch therefore uses #3630's file names, export name and `url` property, and its content byte-for-byte wherever the two are equivalent, so the back-merge either merges clean or conflicts visibly in one place.

The only deliberate divergences, both additive:

- `connectionCount`, which this PR's admission assertions need in order to prove the client actually attempted QSS on both attempts rather than never reaching the endpoint. The fixture test gains one case covering it; #3630's three cases are unchanged.
- #3630's `new App({ ..., qssEndpoint })`, adopted here: without it the app's own `QSS_ENDPOINT` still defaults to `ws://127.0.0.1:3003`, so only the invitation pointed at the dead endpoint.

Verified on a machine with a real QSS stack listening on 3003 — the condition #3581 describes — where `unavailableQssEndpoint.test.ts` passes 4/4.

## Merge with develop

Merged `origin/develop`, which had moved in all three conflicting files. develop's rewrites of the two cases preceding this PR's won whole (try/finally env restoration, `clearProcessOutput` before the retry, `assertAdmissionNotReset` after it, the timeout argument on `expectJoinCommunityError`). Their calls to `joinCompletionTimeoutMs()` and `LOCAL_JOIN_COMPLETION_TIMEOUT_MS` now read `JOIN_COMPLETION_TIMEOUT_MS` — the same 60s they already resolved to, since `beforeAll` sets `LOCAL_TRANSPORT=true` and, once the last case stops overriding it, nothing sets it to `false`. `waitForTransportReady`, `isLocalTransport`, `libp2pStartTimeoutMs` and the `TOR_*` constants have no callers left on either side and stay deleted. In `e2e-win.yml` develop's log-collection and artifact steps are kept. In the CHANGELOG the `[Unreleased]` heading is dropped and the test note sits in a `Tests` subsection at the end of develop's `[11.1.0]`.

### Pull Request Checklist

- [x] I have added the change and issue references to the root CHANGELOG.md.


holmesworcester:
Pushed `2adde19d7` (fast-forward from `d202128d2`) to fix the one red case in the `Run admission timeout recovery test` step.

The failure was inherited from `develop`, not caused here. The same single case, `Timed-out P2P admission recovery › keeps the guest on Join Community when an invalid invitation is submitted`, also fails on every other develop-based branch in the same CI wave: `fix/1701-decode-filenames` (run 35660379445), `fix/1349-settings-close-on-nav` (35660197771) and `fix/2751-today-rollover` (35658916878). The assertion is present on `develop` at `admissionTimeout.test.ts:231`.

Cause: the test expected the guest to be shown "Please check your invitation code and try again", a string that appears nowhere in the app. On a submit that fails to parse, `PerformCommunityActionComponent` sets `InviteLinkErrors.InvalidCode` from `packages/desktop/src/renderer/forms/fieldsErrors.ts`, which reads "Please check your invite link and try again". The xpath lookup therefore never resolved and the step died with `NoSuchElementError`. `JoinCommunity.test.tsx` already pins the app side of this copy against the enum, so only the e2e literal was wrong; `e2e-tests` cannot import the desktop renderer, so the literal now carries a comment naming the source of truth.

The two other red jobs in `E2E cross platform` are unrelated to this PR. `windows / windows` failed at `Run user profile test` and skipped the admission step entirely. `linux-back-compat / linux` timed out waiting for the loading panel to clear during owner community creation in the same-build 11.0.0 baseline; that same flow passed in this run's `linux / linux` job (one-client, multiple-clients, invitation-link and device-linking all green), and the job is green on the four sibling branches above.

holmesworcester:
Merged as `cbcc7afcd` while the re-run of `E2E cross platform` (35665883685) was still at `Run one client test`, per a policy call that CI is advisory here and local verification is the gate.

What that gate was, for the record: the only red case on the previous head was the invalid-invitation copy mismatch above, and the chain behind the fix was checked against real code rather than re-run. `getInvitationCodes('invalid-invite')` throws `Invitation link does not match either v4 or v5 format` when called on the built `@quiet/state-manager`, so `PerformCommunityActionComponent` takes the `setError('name', { message: InviteLinkErrors.InvalidCode })` branch, and that constant is the literal the test now expects. `JoinCommunity.test.tsx` covers the render side against the same enum. The other 10 cases in the step passed on the previous head.

holmesworcester:
Attribution correction to my first comment above. I called the copy mismatch inherited develop debt, which reads as if the assertion had always been wrong. It had not: it was correct when written, and #3509 broke it the same day.

`f6d5c69df` ("Say invite link, not invitation code, in join errors (#2441) (#3509)", merged 2026-09-21 20:46 UTC) changed `InviteLinkErrors.InvalidCode` in `packages/desktop/src/renderer/forms/fieldsErrors.ts` from `'Please check your invitation code and try again'` to `'Please check your invite link and try again'`, and made the matching change to `packages/mobile/src/components/JoinCommunity/JoinCommunity.component.tsx`. It updated the mobile test alongside it but did not touch `packages/e2e-tests`, so `admissionTimeout.test.ts` was left asserting the old string.

That is why the `Run admission timeout recovery test` step went red on every develop-based branch at once rather than only here: the first e2e run on this PR started 21:59 UTC, after #3509 landed. The fix in `2adde19d7` is the e2e half of #3509 that was missed, not a correction to this PR's own work.

PR #3639, opened 2026-09-22
Make admission recovery run against the onboarding redesign, and cover the branches that had no tests

holmesworcester:
Now targets `develop`, which carries the onboarding redesign as of #3512.

`develop` also carries #3591's rewrite of `admissionTimeout.test.ts`, and the two had never
met. #3591 reads the invite field by its old `Invite link` placeholder, which the redesign
renamed to `Link`, and waits for a modal heading the redesign no longer renders on the screen
where a join error is reported. Against the redesigned UI that suite cannot pass.

## What was broken

Measured on the redesign branch before any change: `admissionTimeout.test.ts` was **1 passed,
6 failed**. Two causes.

**`JoinCommunityModal.isReady` could not see the error state.** It waited for an `h3` reading
"Join community" or "Let's get started...". The redesign splits the flow into four screens, and
`JoinCommunity.tsx` deliberately reopens it on the *paste* step when a join fails, because the
three-way choice has no field to carry the message. Five cases failed on "Join community modal
couldn't be found within timeout" after waiting the full 30s. `isReady` now matches any screen
of the flow by test id, which also takes it off the per-step copy.

**The link field was read by placeholder.** `JoinCommunityModal` gains `inviteLinkInput`, which
reads it by test id, and `expectJoinCommunityError` uses that. The message is now read through
`driver.wait` rather than a single `findElement` racing the form's asynchronous validation.

With those two fixed the suite is **6 passed, 1 failed** on that branch, the remaining failure
being a real-Tor case that `develop`'s version of the file does not have.

## Also here

**Two onboarding tests** for redesign entry points that had none: the QR-code branch of the
three-way join, carried to the general channel so the branch is proven to reach admission, and
a one-app walk back out of every branch to Get started, which is the only coverage of
`linkDevicesModalBack`, `joinCommunityModalBack` and the Display QR code step. That step cannot
show a QR from onboarding, because `connection.selectors.deviceLinkUrl` returns `''` without a
current community, so the test asserts the settled copy "Device link unavailable". Negative
control: with the expected string changed, the failure reports
`Received: "Device link unavailable"`.

**The invitation deep-link opener is bounded.** `execSync` on `xdg-open` had no timeout, so a
stale `x-scheme-handler/quiet` association hangs the suite and then reports a guest that never
reached the username modal, which reads as a join-flow regression. The command composition and
the bounded call move to `src/deepLink.ts` with unit tests.

**`onboarding.test.ts` ran in no workflow.** Every E2E step names its files and there is no
catch-all, so both new cases would only ever have run by hand. It gets its own Linux step, and
`deepLink.test.ts` joins the existing unit-test step.

## Merge resolution

Four files conflicted; everything else merged clean.

- **`admissionTimeout.test.ts`** takes `develop`'s version as the base. Only the edits the
  redesign requires are re-applied on top: `inviteLinkInput` in place of the placeholder
  lookup, `driver.wait` on the error message, and `until` in the import. `develop` already
  baselines the QSS dial count, so the only other addition is asserting that the baseline is
  non-zero, which keeps that premise from rotting back into "greater than zero". Taking
  `develop`'s base also drops this branch's real-transport variant of the device-link case,
  which needed a Tor bootstrap and was the one load-dependent failure in local runs.
- **`unavailableQssEndpoint.test.ts`** (add/add) keeps this branch's version: `develop`'s file
  plus an error listener on each bind, a `listening` guard so an unheld port reports itself
  rather than hanging, and a 30s jest timeout.
- **`selectors.ts`** and **`onboarding.test.ts`** keep this branch's versions. `develop`'s
  copies are these same files without the additions made here, with nothing else new in them.
- **`.github/workflows/e2e-linux.yml`** merged cleanly. `develop` had already added the QSS
  fixture's test to the admission step, so the unit-test step only gains `deepLink.test.ts`
  and this branch's separate onboarding step is kept.

## Verified on the merged tree

Against an AppImage packaged from this merge, backend and renderer bundles hash-matched inside
the asar.

| suite | result |
| --- | --- |
| `admissionTimeout.test.ts` | **7 passed, 7 total** (529s) |
| `onboarding.test.ts` | 7 passed, 2 skipped (218s) |
| harness unit tests (`selectors`, `deepLink`, `unavailableQssEndpoint`) | 32 passed (6s) |

The device-link case passes here too. On the pre-merge branch it was the one load-dependent
failure, because that branch ran it over real Tor; `develop`'s version keeps the suite on local
transport.

Earlier full-suite runs on the pre-merge branch, for reference: `deviceLinking` 3 passed,
`invitationLink` 15 passed, `deviceLinking.qss` 12 passed, `multipleClients` 91 passed,
`oneClient` + `oneClient.appImage` 31 passed.

## Environment notes, not this branch

Two local causes made `invitationLink.test.ts` fail on the machine used here before any change:
a stale, header-less `~/.local/share/applications/quiet.desktop`, and
`kernel.apparmor_restrict_unprivileged_userns=1`, which kills the Quiet the desktop entry
launches on the SUID sandbox check before it can forward the URL. CI's ubuntu-22.04 runner
needs neither.


holmesworcester:
Merged `develop` and retargeted this PR to it. Body rewritten for the new base.

`develop` now carries the onboarding redesign (#3512) and, separately, #3591's rewrite of
`admissionTimeout.test.ts`. Those two had never met: #3591 reads the invite field by its old
`Invite link` placeholder, which the redesign renamed to `Link`, and waits for a modal heading
the redesign no longer renders where a join error is reported. Four files conflicted.

- **`admissionTimeout.test.ts`** takes `develop`'s version as the base, with only the edits the
  redesign requires re-applied: `JoinCommunityModal.inviteLinkInput` in place of the placeholder
  lookup, `driver.wait` on the error message instead of a single `findElement` racing the form's
  asynchronous validation, and `until` in the import. `develop` already baselines the QSS dial
  count, so the only other addition is asserting that the baseline is non-zero. Taking
  `develop`'s base also drops this branch's real-transport variant of the device-link case,
  which needed a Tor bootstrap and was the one load-dependent failure in local runs.
- **`unavailableQssEndpoint.test.ts`** (add/add) keeps this branch's version: `develop`'s file
  plus an error listener on each bind, a `listening` guard so an unheld port reports itself
  rather than hanging, and a 30s jest timeout.
- **`selectors.ts`** and **`onboarding.test.ts`** keep this branch's versions. `develop`'s
  copies are the same files without the additions made here, with nothing else new in them.
- **`.github/workflows/e2e-linux.yml`** merged cleanly: `develop` had already added the QSS
  fixture's test to the admission step, so the unit-test step only gains `deepLink.test.ts`,
  and this branch's separate onboarding step is kept.

Verified against an AppImage packaged from the merge, with the backend and renderer bundles
hash-matched inside the asar:

| suite | result |
| --- | --- |
| `admissionTimeout.test.ts` | 7 passed, 7 total (529s) |
| `onboarding.test.ts` | 7 passed, 2 skipped (218s) |
| harness unit tests | 32 passed (6s) |

Note on the commit count: #3512 was squash-merged, so `develop` does not contain the
redesign branch's individual commits and this PR lists 68 of them by SHA. The content delta
against `develop` is the 8 files above, 268 insertions.

Not merging this; leaving it for the train's owner.

PR #3524, opened 2026-09-13
onboarding: Agree & join to the design (mobile + desktop); Want a server? close glyph; Starting Quiet without bar

holmesworcester:
Three follow-ups from the code scan of the merged demo tree, on top of design/h1-no-bar (#3521). Merges design/storybook-grid @ a7cba7e9f first (the Join-from-invite-link stages incl. the joiner's Agree & join 2811:2724 and Starting Quiet 2811:2770).

| item | frame | before | after |
|---|---|---|---|
| Starting Quiet (desktop `LoadingPanel/StartingPanelComponent`) | dSEZJr9crJjcV3ILogea9C 2811:2770 | `Modal` with no title: an empty 60px bar + hairline over the logo | `withoutHeader`, like Get started and JoiningPanel — see note 1 |
| Want a server? (desktop `ServerOffer/ServerOfferComponent`) | 2922:10009 | `withoutHeader`: no bar, no close glyph | `withoutTitle` + close glyph at the left; glyph / backdrop / Escape = "No thanks" (`handleClose(false)`; before, a backdrop close passed the MUI event object as the selection) |
| Agree & join (desktop `TermsOfService/TermsOfServiceComponent`) | library modal/small 6209:16742 (800×636) / 6209:17520 (1064×636); copy from 2811:2724 / 3054:4090 | full-window modal "Accept Terms of Service", centred text, "Agree and Join" + "Leave Community" | modal/small card: 800 wide, 1064 from a 1280 window; titled bar "Agree & join" + back arrow + hairline; 468 column centred (166 in at 800 / 298 at 1064, measured); left-aligned body 14/20 naming the host; "Privacy Policy and Terms of Use" one underlined link; one pill "Agree & Join" (50, r16); back arrow = not agreeing; Leave Community gone |
| Agree & join (mobile `TermsOfService/TermsOfService.component`) | 3054:4090 | Appbar "Terms of Use", centred 14 text ("…end-to-end encrypted messaging…"), "Agree & Continue" + "Leave Community" | Appbar "Agree & join" + back arrow (designed titled); body 14/20 left-aligned with the host, one underlined link; pill "Agree & Join" (newDesign: 50, r16) left-aligned; padding 16, gap 16 (measured); `onLeave` removed |

[...]

PR #3519, opened 2026-09-13
onboarding: Paste link row on Link devices, device-link validation

holmesworcester:
Stacks on #3516 (`design/onboarding-qr-scan`): impl → qr-scan → this.

User requests (2026-09-13), verbatim: *add a "Paste Link" item to the linked devices screen (not a small blue link but a normal menu item with a "link" anchor logo to match the others) and add logic and an error message to ensure it is a device invite link* — then *also make the desktop linked devices screen match the mobile one and the designs (make people display the QR code i think)*, with the decision that the QR sheet carries **Copy link** as a primary button in the Add-members "Share code" slot on both platforms and never shows the raw link.

## 1 · Paste link row + device-link validation

**Desktop** (`packages/desktop`)
- `Onboarding/LinkDevicesComponent.tsx`: a third `ActionRow`, **Paste link**, under *Scan QR code* — the same library Button row as the two above it with the library's link glyph (`onboardingIcons.inviteLink`, the one *Join with invite link* uses on 2811:2562). Not a text link.
- `Onboarding/LinkDevices.tsx`: new step `pasteLink` — the existing paste step (`PasteLinkComponent`, heading *Paste a link to Join*) under the **Link devices** title bar; back returns to the rows. The scanner's paste fallback is unchanged.
- `Onboarding/PasteLinkComponent.tsx`: `linkKind?: 'any' | 'device'`. With `'device'` the pasted value is parsed by the shared parser (`getInvitationCodes` → `@quiet/common` `parseInvitationLink`) and checked with `isDeviceInvitationData`; a member link, or any other invitation, sets an inline error under the input and calls nothing. Both paste routes in the Link devices flow pass it. Text that is not an invitation keeps *Please check your invitation code and try again*.
- `forms/fieldsErrors.ts`: `InviteLinkErrors.NotDeviceLink`.

**Mobile** (`packages/mobile`)
- `LinkDevices/LinkDevices.component.tsx`: the same third `ActionRow` with `InviteLinkIcon`; the screen navigates to `PasteInviteLinkScreen` with the new variant `pasteDeviceLink` (title *Link devices*, heading *Paste a link to Join*).
- `JoinCommunity/JoinCommunity.component.tsx`: in the Link devices variants (`pasteDeviceLink`, and `deviceLink` — Scan QR code's stand-in on mobile) a non-device link shows `NOT_A_DEVICE_LINK_ERROR` under the input and dispatches nothing.

## 2 · Desktop Link devices / Linked devices match mobile and the designs

[...]

PR #3517, opened 2026-09-13
design: consistency audit — radii, spacing, type, states across the desktop app

holmesworcester:
Draft, opened at the end of a session: the audit and its one-place fixes are committed; the agent that produced them hit a usage limit before its final screenshot pass and verification report, so **treat verification as pending** — tsc/jest and a Storybook look at every touched story must be re-run before review.

## What this is
A design-consistency sweep of the desktop renderer against the design library (tokens, `ONBOARDING.md` decisions, and the Figma files indexed for this project), excluding what `design/onboarding-impl` is already fixing. The report is `packages/desktop/src/renderer/design-system/CONSISTENCY_AUDIT.md`: **99 findings** — 73 fixed here, 18 not fixed (reason in each row), 4 already consistent, and the in-flight items listed separately.

## Shape of the fixes
One-place theme overrides in `design-system/theme/components.ts` cited to library nodes (Input3.0, Tooltip-content, Overlay menu, modal/small, the action bar's secondary button, Search result, body text), spread into both themes; palette additions (`ink`, `error10`, corrected `statusGreen` and caption colour). Then component edits removing copied numbers and snapping to the 4px grid and the type scale: sidebar (window-controls bar, section gaps, weights, hover from the desktop sidebar design), channel view (message list + header on the type scale), composer (radius 16, themed placeholder/dropdown), modals/menus/settings (shell titles, outlined secondary buttons, grid spacing), add-members (outlined secondary, glyph size, online green), plus Components/Consistency stories for the controls moved onto the theme.

## Verification status
Pending — see above. Each finding in the report cites a node id, token or spec line.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01F9YfCnS6xqgfsN6r1njc57

PR #3643, opened 2026-09-22
Want a server? follows the design, on both platforms

holmesworcester:
Rebuilds the "use Quiet's server?" step from the design. The rule is that this step follows the Figma frames rather than what the app happens to draw, so the UI is replaced and every bit of develop's behaviour is kept.

## What the frame shows

**Want a server?** — `2922:10009` ([Figma](https://www.figma.com/design/f6Nr5b5wtvk6Xoh1HJZ8Dd?node-id=2922-10009), Storybook `onboarding-flow--want-a-server`): the 60 bar zone with the close glyph at (14, 16) and no title, a 48×51 server glyph in a 64 box, the heading *Want a server?* (28/34 500 #222222), an *It's free!* pill (24 tall, #F9EFFF inside a #ECDCF5 hairline, radius 4, label 14/20 500 #461863), the body *Messages are still end-to-end encrypted, joining will be faster, and Quiet will work much better on iPhones.* (14/20 #222222, centred in a 327 column), the pill button *Use Quiet's server* (50 tall, radius 16, #521C74, label 16), the *Not now* text link (16/16 #7F7F7F), a #F0F0F0 rule running the full width of the frame, and *Don't show this again* with the library's checkbox (a 16 box at radius 3, #999999 hairline, #7F7F7F filled with a white tick). 24 between every block; the text is padded 24 either side and the rule runs past that padding.

Per the no-bar rule (#3521): the frame draws a hairline under its glyph, and this screen has a large heading, so the bar keeps only the glyph — no title, no line. The frame's own title-bar text ("Add members") is hidden in the file and stays out.

One deviation, deliberate: the frame's pill reads `It's free!` with a straight apostrophe while its other two strings use typographic ones. Both platforms use `It’s free!` so the screen is internally consistent; the words are the frame's.

## Desktop

`ServerOffer/ServerOfferComponent.tsx` is rebuilt around the frame's geometry: the library modal shell with `withoutTitle alignCloseLeft`, and the prototype's 375 column (`CONTENT_COLUMN_WIDTH`) centred in it the way the other onboarding stages host it. It does not use `OnboardingBody`, which models a heading/intro pair in a padded column — this frame puts a pill between the two and needs its rule full-bleed. The old arrangement had a gray `body2`, a *No thanks* button and a MUI-default checkbox in a `FormControlLabel`. `theme.palette.colors` gains `borderLightPurple` (#ECDCF5) for the pill's hairline, in both the light and dark palettes.

## Mobile

The frame is a full screen, so `ServerOffer/ServerOffer.component.tsx` is now that screen — `Appbar withoutTitle crossBackIcon`, content top-anchored 24 under the bar zone — and the 2/3 bottom drawer it used to arrive in is gone, the same way Agree & join's drawer went in #3524. The duplicate `ServerOffer/CreatingOffer/ServerOffer.component.tsx` goes with it; `CreateCommunityScreen` renders the offer directly while it is open. The palette gains `background.lightPurple03` (#F9EFFF, the value desktop calls `colors.lightPurple`); the heavier `#F3E8FF` keeps its own name and its one other user.

## Behaviour kept verbatim

- Shown during community creation, not from the design's Home → Add members entry (recorded as the stage's divergence in `flow/implementation.ts`). Desktop gates on `QSS_ALLOWED` plus a parseable `QSS_ENDPOINT`; mobile on `Config.QSS_ALLOWED`.
- *Use Quiet's server* → `createCommunity({ name, useServer: true })`, mobile queueing the terms step through `setPendingNavigation`; *Not now* → the same payload with `useServer: false`. Then the username step, as before. The resulting `qssEndpoint` path is untouched.
- Closing is *Not now*: the desktop modal's close glyph and the mobile bar's close glyph both decline, which is what dismissing the drawer used to mean.
- `showDontShowAgain` still gates the rule and the checkbox, and mobile still reports the checkbox value alongside the decision. Nothing persists it, as before.
- Desktop testids (`ServerOffer-UseQuietServer`, `ServerOffer-NotNow`, the `ServerOffer` modal prefix) are unchanged, so the rtl and Selenium suites keep working. Two selector updates follow the UI: the desktop e2e checkbox xpath uses the renamed row class, and the mobile detox waits move from the drawer's id to `server-offer-component` with the buttons matched by testid.

## Verified

- Desktop jest: 117 suites / 426 passed (5 skipped, 1 todo) and 142 snapshots, with only this component's inline snapshot refreshed; the 118th suite, `src/main/main.test.ts`, cannot load without `packages/backend-bundle/bundle.cjs`, a build artifact a fresh worktree has not produced — with that file in place its 16 tests pass as well. Re-ran `ServerOffer` and `CreateCommunity` with `--ci` to confirm the inline snapshot is stable.
- Mobile jest (`npm test`): 89 suites / 340 tests green, 46 snapshots unchanged.
- `tsc --noEmit` clean on both packages; eslint/prettier clean on everything touched. A full `eslint ./src` on desktop reports 4 prettier errors in `TextMessage.blankLines.{test,cy}.tsx`, which are untouched here and fail the same way on the base branch.
- Mobile has no emulator run, so the screen's geometry was checked by dumping the rendered tree's resolved styles: bar 60 with the 28 glyph box and a 16 icon, content top 24 and 24 between blocks, the glyph 48 in a 64 box, heading 28/36 500 #222222, pill r4 / 1px #ECDCF5 / #F9EFFF / 8-2 padding / 14-20 500 #461863, body 14/20 #222222, button 12-20 padding #521C74 r16 min-height 50 with a 16 label, link 16/16 #7F7F7F, rule 1px #F0F0F0 full width, checkbox 16 at r3 with a #999999 hairline and an 8 gap to a 14/20 #222222 label.
- Storybook: static build served and shot under xvfb — `Components/ServerOffer` *Want a server?*, *With Don't show this again* and *Don't show this again, checked*, each compared against the frame export, plus the `Components/Onboarding` group to confirm nothing else moved. The bar shows the glyph alone, with no title and no hairline.

Not verified: no device or desktop app run, so the mobile screen has only been exercised through react-native-testing-library, and the detox and Selenium selector changes have not been run against a build.


holmesworcester:
Review follow-up. Each finding and what it changed:

1. **The glyph goes back, on both platforms.** It was an answer to the offer, which created the community without a server and left no way back to the name. The frame's own link is *Glyph → back*, so the × now returns to the create step and decides nothing; *Not now* remains the explicit decline. Desktop: `handleBack` drives the modal's glyph, the backdrop and Escape, and the container clears the pending name without dispatching. Mobile: `onBack` does the same. The doc row and the *Implemented by* line in `ONBOARDING.md` say this, and both platforms have a test for it. Because the icon and the action now differ, the glyph also says what it does to assistive tech: `Modal` takes `closeAriaLabel` and `Appbar` takes `backAccessibilityLabel`, both defaulting to today's wording, and the offer passes "Go back".

2. **"Don't show this again" reaches the caller.** Desktop's `handleClose` now has mobile's `(useServer, dontShowAgain)` shape, mobile's screen no longer drops the value, and both containers act on it — which today means logging it, because nothing stores it. Opened https://github.com/TryQuiet/quiet/issues/3644 for the persistence decision (per device or per account, what suppressing the offer should then do, and the store migration), and the code points at it. The desktop e2e selector is reconciled by deletion: no test used `dontShowAgainCheckbox`/`setDontShowAgain`, and neither container passes `showDontShowAgain`, so the product never renders that checkbox; the class the xpath was chasing is gone from the suite with a comment saying why.

3. **The checked story is args-driven.** The component takes `defaultDontShowAgain`, so the story sets the initial state instead of clicking `document.body`'s first checkbox — which on the Docs page was the neighbouring story's.

4. **Mobile story ids are ids.** `.add('default')`, `.add('with-checkbox')`, `.add('checkbox-checked')` — single-quoted, no spaces or `?`, and not a copy of the heading. Verified against the regex `packages/mobile/e2e/storybook.test.js` actually uses, which returns all three (the old double-quoted one matched nothing).

5. **`colors.gray90` exists and is used** for the heading, the body, the glyph and the checkbox label, so desktop draws the frame's #222222 instead of MUI's `rgba(0,0,0,.87)` and the doc's claim is true of both platforms. The dark palette gets #FFFFFF for the same token: dark mode is live (`useTheme` follows the OS) and the frames only specify the light theme, so a literal #222222 there would have been invisible.

6. **The mobile column scrolls.** The content is in a `ScrollView` with `flexGrow: 1` and the frame's 32 at the foot, so at 1.3–2.0 font scale the rule, the checkbox and the link stay reachable.

7. **The create form is no longer unmounted.** The offer renders over it (`StyleSheet.absoluteFill`), with the form hidden from assistive tech while it is up, so the typed name stays in the form's own state. The test presses Continue again after going back, without retyping, and asserts the same name reaches `createCommunity`; against the early-return version that test fails, which is how I checked it.

8. **`JoiningOptIn` no longer shares testIDs** — it is `joining-opt-in-component` / `joining-opt-in-dont-show-again` now. Kept rather than deleted: it has no production caller, but `flow/implementation.ts` still names it as the v1 joiner agree screen and it has its own story, so retiring it belongs with that stage's row rather than with this PR.

9. **The dead `visible` prop is gone** from the mobile component, along with the test for its unreachable branch; the parent gates the render.

10. **`ModalBottomDrawer` stays, and here is the "which".** `ONBOARDING.md`'s library → code table maps the design's three sheets (Join with QR code 2811:2460, Link devices' QR code 2811:2601, Scan QR code 2811:2587) to it on mobile, and the audit rows list all three as not yet built as sheets — so it is the planned implementation, not an orphan, and the `AndroidCompatibility` story documents what it does when the activity is resized, which those sheets will need. It also still has its own unit test. I added a header comment recording that it has no consumer until those sheets are built. Say the word and I will delete it with that table row in a follow-up.

11. **Cleanups done.** The `onChoose`/`onDismiss` `useCallback` aliases are gone, the rule and the checkbox are under one guard, and `data-testid='server-offer'` is removed (nothing used it; the convention is `ServerOffer-*`).

Re-verified on this head: desktop jest green with the ServerOffer inline snapshot refreshed; mobile `npm test` 89 suites / 342 tests green, 46 snapshots unchanged; `tsc --noEmit` clean on both; the three stories rebuilt, re-shot under xvfb and compared against the frame export, including the ticked state now coming from args.

On CI: `linux`, `linux-back-compat` and `linux-qss` fail identically on #3524 and #3521, which do not contain this work, so they are inherited from the stack rather than caused here.

holmesworcester:

This branch was cut from `design/agree-and-join`, so `develop` has moved nine PRs past its base.
Nineteen files conflicted; twelve of them are that staleness alone, in files this branch never
touches, and they take `develop` verbatim: both JoinCommunity components and their tests, both
JoinCommunityOptions components, desktop `LinkDevices.tsx` and its test, mobile `App.tsx`, mobile
`LinkDevices.test.tsx`, `flow/gen.cjs` and the more-horiz glyph.

The seven this branch actually owns:

| file | resolution |
|---|---|
| `ServerOfferComponent.tsx` + test | this branch. It rebuilds that surface to the frame, which is the point of the PR, and supersedes #3524's close-glyph pass over the old arrangement. |
| `ui/Modal.d.ts`, `ui/Modal.tsx` | union. This branch's `closeAriaLabel` is kept, so Want a server? can tell assistive tech that its close glyph goes back, and `develop`'s 56 title margin (#3517) and `cornerRadius` (#3524) are kept. |
| `mobile Appbar.component.tsx` | this branch, which adds `backAccessibilityLabel` on top of `develop`'s bar-zone implementation. |
| `e2e-tests/src/selectors.ts` | `develop`, whose version already contains this branch's additions plus the scanner helpers from #3516. |
| `ONBOARDING.md` | `develop`'s file with this branch's Want a server? row, now reading **done**. |

## One snapshot regenerated, and one thing swept up

`ServerOfferComponent.test.tsx`'s inline snapshot drifted by two lines once #3517's theme work
reached `develop`: both are `class` attributes and they are identical once the `css-…` token is
stripped, so no element, attribute or value moved.

This also removes a duplicated "Device linking — desktop designs" section from `ONBOARDING.md`.
That duplicate is mine: my #3519 union concatenated both sides of that file when both already
carried the section. The richer of the two copies is kept.

|---|---|---|
| desktop | clean | 127 of 128 suites, 553 passed |
| mobile | clean | 97 suites, 413 passed |
| e2e-tests | clean | not run: they need a packaged AppImage |

PR #3646, opened 2026-09-22
chore(desktop): clear dist/ before every packaging script

holmesworcester:
electron-builder's `linux.files` glob is `dist/**/*` and electron-builder also writes its output into `dist/`, so a repeat local build in the same worktree packs the previous AppImage into the new `app.asar` — observed today as a 321 MB AppImage (exactly double) with `/dist/Quiet-11.1.0.AppImage` inside it.

`distMac:local` already ran `rmDist` first; this makes `distMac`, `distUbuntu`, `distUbuntu:qss` and `distwin` do the same. CI never hit it (fresh runners). Verified: the repackaged build came out at the expected 161 MB.


holmesworcester:
## Merged with `develop` and gated

The merge was clean; no file conflicted. `develop` had moved on since this was cut, by #3507,
#3493, #3494, #3414, #3386, #3648 and #3649. The diff against current `develop` is still exactly
the four script lines and the changelog entry.

**Gate:** `npm run rmDist` in `packages/desktop` resolves and exits 0.

## One thing worth a follow-up

`rimraf` is **not declared** in `packages/desktop/package.json`, in either `dependencies` or
`devDependencies`. It resolves only because it is hoisted: `node_modules/rimraf -> 2.6.3` is a
top-level entry in `packages/desktop/package-lock.json`, pulled in transitively by `cacache`,
`copy-concurrently`, `fs-extra` and others.

That is safe today — the lockfile pins it, so `npm ci` places it deterministically — and the
undeclared dependency predates this PR, since `distMac:local` already called `rmDist`. What
changes is the blast radius: `rmDist` now runs in the three CI packaging scripts as well, so if a
future lockfile regeneration stops hoisting `rimraf` to the top level, every packaging job breaks
rather than one local script. Declaring it explicitly would close that off. Not done here, because
it needs a lockfile regeneration this worktree cannot do faithfully offline.

PR #3648, opened 2026-09-22
Recover design work dropped in the onboarding-impl develop merge

holmesworcester:
A commit-level audit of the `design/onboarding-impl` ← `develop` root merge and of the three sibling branches the merge train landed by conflict resolution, plus the recovery of the two things it found lost.

## Method

For every non-merge commit, each added line was normalized for whitespace and quote style and looked for in current `origin/develop`: in the same file, elsewhere in the same package, or anywhere in `packages/`. Lines a *later commit on the same branch* replaced were excluded, so each commit is scored on the contribution that actually reached its branch head. Emotion class hashes in snapshots were counted separately, since develop legitimately regenerates them. Every commit that did not score 100% was then read by hand against develop.

| range | commits | branch-added significant lines | absent from develop |
|---|---|---|---|
| `f95d8cdb8..1c24b6b0b` (root) | 53 | 12,498 | 666 (5.3%) |
| `design/action-progress-states` | 24 | 7,335 | 16 (0.2%) |
| `design/mobile-home` | 70 | 7,054 | 18 (0.3%) |
| `design/desktop-sidebar` | 65 | 5,450 | 18 (0.3%) |

The three siblings' own commits landed essentially whole. Their entire residue is five files they inherit from the root branch, and every line in it is prose in a doc comment, the `qrCode` copy variant that #3520 deliberately deleted when the real scanner landed, or `Modal.tsx`'s close-glyph margin, which #3517 changed from 36 to 56 with the reason in the code ("56 keeps the centred title clear of the 56-wide glyph zone"). Nothing of theirs is lost.

## Every on-the-fly refinement survived

Each of the late design commits was checked line by line. All present:

| commit | refinement | where it is now |
|---|---|---|
| `1c24b6b0b` | clip the exported circular logo to its frame | present |
| `8d62aec75` | Get started shows the circular Quiet mark at 120px | present; reflowed by prettier, and the Appbar above it was dropped later by a recorded 2026-09-13 decision |
| `eb6f6b912` | the circular Quiet mark on desktop and mobile | present |
| `d3ade16eb` | 16px corner radius on every MuiButton | present as `BUTTON_RADIUS = 16`; 19 of its 21 unmatched lines are emotion hashes |
| `c992dceff` | the library's interaction-state values, Continue disabled until valid | present; all 3 unmatched lines are emotion hashes |
[...]

PR #3386, opened 2026-07-27
Fix: write to clipboard unconditionally, only skip the in-app confirmation toast on Android 13+.

JohnieWalkerCZ:
Fixes #3369

### Pull Request Checklist

- [x] I have linked this PR to a related GitHub issue.
- [X] I have added a description of the change (and Github issue number, if any) to the root CHANGELOG.md.

### (Optional) Mobile checklist

- [ ] I have run e2e tests for mobile
- [ ] I have updated base screenshots for visual regression tests

PR #3649, opened 2026-09-22
storybook: restore the walkthrough camera toggle and the Link devices stories

holmesworcester:
The merge train landed the 11.3 onboarding stack on `develop`, but two design-library entries could not come with it: `develop` and each sibling branch had restructured `OnboardingScreens.stories.tsx` around different `Screen` shapes, so #3516's **walkthrough camera** and #3519's **six Link devices stories** were dropped. Both resolution comments recorded them as follow-ups. The components and their jest coverage all landed; only the stories were missing. This restores them, re-expressed against `develop`'s `Screen` shape and components.

## Screens/Onboarding, before and after

| | before (develop) | after |
| --- | --- | --- |
| stories in the group | 20 | 26 |

**Added (six, #3519's):** `LinkDevicesEmpty`, `PasteLinkOnLinkDevices`, `PasteLinkNotADeviceLink`, `DisplayQrCodeGenerating`, `DisplayQrCodeUnavailable`, `SettingsLinkedDevices`. `ONBOARDING.md` already named all six under "Stories:"; it was describing entries that did not exist.

**Changed:** `LinkDevices` is now the **share** direction a community gives (Display QR code, Copy link), and `LinkDevicesEmpty` the **receive** direction without one (Scan QR code, Paste link) — the pair #3519 intended. `DisplayQrCode` draws the real sample device link rather than a placeholder string. `Walkthrough` gets the camera.

**Unchanged:** `GetStarted`, `JoinCommunity`, `RecoverAccount`, `OpenInviteLink`, `PasteALink`, the six `JoinWithQrCode*`, `CreateCommunity`, `ChooseUsername`, `InviteLinkPath`. The nine standalone scanner stories were already present.

## The walkthrough's camera

Join with QR code and Scan QR code rendered `PasteLinkComponent` stand-ins. They now render the real `QrScannerComponent`, with the four camera buttons (sample link QR code / blank / denied / none) driving `installStoryCamera`, so the story runs the real video → canvas → jsQR pipeline. Three things follow from re-expressing it here rather than porting it:

- **The scanner is keyed on the camera mode.** It asks for the camera once, on mount, so without a key the buttons did nothing while you were looking at the scanner. Switching now re-opens it.
- **Link devices takes a direction**, as `LinkDevices.tsx` does from `currentCommunity`. On `develop` the walkthrough drew the share direction while wiring `onScanQrCode`, which that direction does not draw — the `scanQrCode` step was unreachable. A button on the Link devices step toggles the community condition; Display QR code and the Paste link row both stay reachable.
- **A device link records the consent step** (`device-link consent, then linkDevice({ inviteData, deviceLinkConsent: true })`), which is what both containers do since #3516. The standalone scanner stories already said this; the walkthrough still recorded the old bare `linkDevice`.

The scanner's *Paste a link* falls back to its own step, so the back arrow retraces it, matching `pasteFromQrCode` and `paste` in the two containers.

## Two bar titles corrected

Checking the no-bar rule turned up two stories that state the opposite of what the app does. `LinkDevices.tsx` has `TITLED_STEPS = { display: 'QR code', scan: 'Scan QR code' }`, and its own test asserts both; neither screen draws a heading of its own, so nothing would repeat. The stories showed both with the title dropped. They now keep it, and Display QR code shows the close glyph its sheet has. Join with QR code keeps its dropped title — `JoinCommunity.tsx` has an empty `TITLED_STEPS` and drops every one.

The shell's bar hairline is also gone. The app's `Modal` draws one only for a caller passing `addBorder`, which defaults to `false` and which neither onboarding modal sets. No story used a kept bar title before this change, so nothing else moves.

## Verified

- `tsc --noEmit` on `packages/desktop`: exit 0. ESLint on both touched files: clean.
- Desktop jest: **128 of 129 suites, 563 passed, 142 snapshots unchanged.** The one failure is `src/main/main.test.ts`, which cannot resolve `backend-bundle` unless the webpack bundle is built in this worktree; it fails identically on `develop` and this change touches nothing in `src/main`.
- **New suite**, `OnboardingScreens.stories.test.tsx`, 13 cases: every restored story rendered and asserted on what distinguishes it, the walkthrough driven from Get started through both scanners, and the bar title of each checked against the container. It drives the real decoder over a jsdom canvas stub rather than asserting the stories merely mount. Negative controls: putting the paste stand-in back fails three cases, putting `LinkDevices` back on the receive direction fails a fourth.
- **Storybook**: static `build-storybook`, served over http, all 26 stories in the group shot with Electron under xvfb and looked at one by one. No story errors (the only console warning is the pre-existing `Addon controls` colour warning). The walkthrough was additionally driven through twelve states — both scanners scanning, decoded, denied and with no camera, the paste fallback, both Link devices directions, the QR sheet and the Paste link row — and each looked at.

## Not covered

- `ONBOARDING.md` says the Link devices and Display QR code stories show the Figma frames alongside. `develop` dropped the per-story export columns when it restructured the file, and this change does not bring them back, so that clause is still stale.
- The six `JoinWithQrCode*` captions read "dropped — the screen has its own heading". The bar treatment is right, but the scanner has no heading; the real reason is that `JoinCommunity.tsx` drops every bar title. That wording is `develop`'s shared `barCaption` helper and is left alone.


holmesworcester:
## Review findings applied

Four findings from the review of this branch, in one follow-up commit.

**The walkthrough recorded every decode twice (medium).** The story draws each step in both columns, so two `QrScannerComponent`s mounted and both decoded the same code; the log read `dispatched (2)` with the line repeated, and `go`/`finish` ran twice. The app has one camera, and a design surface that names what the app dispatches has to name it once. `onInvitation` is now gated on a `handled` ref: the first scanner through wins, and `go`, `back` and `finish` clear it so the next screen starts fresh. The flag is set **after** the navigation call, so the clear that navigation performs for the screen being opened does not re-open this decode to the second column.

The standalone scanner stories still record both columns on purpose — their log is labelled "dispatched, both columns", and showing two independent cameras is the point there.

**The step-dependent camera text was untested (low).** `installJsdomCanvas` hands the scanner the frame the test injected, so which link the story chose to paint could not be observed through it. The test now spies on `drawQr`, which `storyCamera.ts` paints with, and asserts the text directly: the member link on Join with QR code, the device link on Scan QR code, and nothing at all while the camera is denied or absent. The harness comment now says plainly which half of the pipeline it stands in for and which it does not.

**The stub track never ended (low).** `storyCamera`'s repaint interval clears itself when every video track reports `ended`, so a stub whose `stop()` left `readyState` at `'live'` kept the story painting into a released camera for the rest of the suite. `stop()` now ends the track, and two cases assert the release.

**`ScannerScreen` could take no bar treatment at all (low).** Widening `droppedBar` to optional made `bar` and `droppedBar` both optional, so a scanner with neither would have silently drawn a bar zone with no title. They are now a `{ bar } | { droppedBar }` union.

## Re-verified

`tsc --noEmit` exit 0, ESLint clean, desktop jest **128 of 129 suites, 563 passed** (`src/main/main.test.ts` still needs the `backend-bundle` artifact and fails the same way on `develop`). The 13-case suite is green, and each fix has a control that fails without it: removing the dispatch gate fails three cases, leaving the track live fails two, painting the wrong link fails one, and dropping a scanner's bar treatment is now a type error. Storybook rebuilt and the walkthrough re-driven through its camera states; the log now reads `dispatched (1)`.

holmesworcester:
## Merged with `develop` and gated

The merge was clean; no file conflicted. `develop` had moved on since this branch was cut, by
#3643, #3507, #3493, #3494, #3414, #3386 and #3648.

This closes the one gap I left open in the train. My resolution comments on
[#3516](https://github.com/TryQuiet/quiet/pull/3516#issuecomment-5771764754) and
[#3519](https://github.com/TryQuiet/quiet/pull/3519#issuecomment-5771902581) recorded the
walkthrough camera and the six Link devices stories as not carried over, because the two versions
of `OnboardingScreens.stories.tsx` had been restructured around different `Screen` shapes and
grafting one onto the other was not mechanical. Re-expressing them against `develop`'s shape,
which is what this does, was the right way round.

Verified present on the merged tree: 26 story exports, including `LinkDevicesEmpty`,
`PasteLinkOnLinkDevices`, `PasteLinkNotADeviceLink`, `DisplayQrCodeGenerating`,
`DisplayQrCodeUnavailable` and `SettingsLinkedDevices`, and the walkthrough's camera helper and
mode list are back.

## Gate

| package | typecheck | tests | lint |
|---|---|---|---|
| desktop | clean | **130 of 130 suites, 611 passed**, snapshots unchanged | unchanged: the 4 pre-existing errors in `TextMessage.blankLines.{test,cy}.tsx` |

The whole desktop suite passes here, `src/main/main.test.ts` included. That suite only ever failed
in this train because it cannot resolve `backend-bundle` in a tree that has not built the backend;
it mocks the module outright, so a stub at the gitignored `packages/backend-bundle/bundle.cjs`
is enough. The stub is local to the worktree.

PR #3651, opened 2026-09-22
Linked devices: show the device list on the share direction (fixes #3636)

### LINKED ISSUE #3636: Backend: linked-device list endpoint for the Linked devices screen
The onboarding redesign (the #3512 stack) draws a **list of linked devices** on the Link devices screen and inside the Linked devices settings tab. That list currently has no data source: the redesign branch carried its own `GET_LINKED_DEVICES` socket action, `LinkedDevice` type, `getLinkedDevices` saga and backend handler, all of which arrived as part of the #3400 device-linking patch applied to the branch. When #3400 merged to develop in its reviewed form it did **not** include that endpoint, and the root merge of develop into `design/onboarding-impl` takes develop's backend verbatim, so the whole path was dropped.

Dropped in merge commit `bbb6e4ab2e625b0933fc761b55c116f9733840e5` (branch `design/onboarding-impl`, PR https://github.com/TryQuiet/quiet/pull/3512):

- `SocketActions.GET_LINKED_DEVICES` and `RequestLinkedDevicesPayload` / `LinkedDevice` in `packages/types`
- `packages/state-manager/src/sagas/appConnection/linkedDevices/getLinkedDevices.saga.ts` (+ its test), the `linkedDevices` slice field, selector and `setLinkedDevices` reducer
- the backend `getLinkedDevices` handler bridged through `SocketService`
- the list UI in `packages/desktop/.../Settings/Tabs/LinkedDevices/LinkedDevices.component.tsx` and `packages/desktop/.../Onboarding/LinkDevicesComponent.tsx`, and `packages/mobile/src/components/LinkDevices/LinkDevices.component.tsx`
- the e2e helpers `LinkDevicesModal.linkedDeviceNames()` / `hasNoLinkedDevices()`, and the `onboarding.test.ts` case `"A's device list shows B"` is now `it.skip`

### What the UI needs

One device per entry on the current user's LFA member record:

| field | use |
| --- | --- |
| `deviceId` | React key; stable across renders |
| `deviceName` | the row label |
| `isCurrent` | this device is excluded from the list |
| `removedAt?` | removed devices are filtered out (and are what a future unlink would set) |
| `created?` | not rendered today; wanted for "linked on …" |

Two things beyond a one-shot read:

1. **A request/response** for the current list, so the screen can populate when it opens.
2. **An event when the set changes**, so the list updates while the screen is open — linking a second device is exactly the moment a user is looking at it. The branch approximated this by re-requesting on `users.setUsers`, since the device set lives on the team graph and any member update may carry a new device; a dedicated event would be better.

### Storybook stories removed

Both under `Screens/Onboarding` in `packages/desktop/src/renderer/design-system/screens/OnboardingScreens.stories.tsx`:

- `LinkDevices` lost its populated `linkedDevices` prop (one current device + one other)
- `LinkDevicesEmpty` — the "no linked devices" empty state — was removed outright

They should come back with the endpoint, since the empty state is a real design state.

holmesworcester:
Fixes #3636 — https://github.com/TryQuiet/quiet/issues/3636

Develop's Link devices surfaces say, in five places, that the frames' **Linked devices** list is not built because nothing on this line can enumerate a user's devices, and that a card permanently reading *No linked devices* would be false the moment a device was linked. https://github.com/TryQuiet/quiet/pull/3648 landed the backend that enumerates them. This answers both halves.

### Where the list goes

The share direction only, which is the direction with a community and so a team graph to read the devices from. Receiving has neither.

- **Desktop:** one `LinkedDevicesList` under the rows of `LinkDevicesComponent`, so Settings → Linked devices and the Link devices modal's entry get it from the same place.
- **Mobile:** the same list on the Link devices screen.

Both ask for the list when the surface opens sharing. The master saga already re-runs the read on `users/setUsers`, so it refreshes when the graph changes. This device is never a row, and a device removed from the account drops out even though the graph still carries it.

### Unread is not empty

That was the second half of the objection, and it needed a state change. `connection.linkedDevices` was `LinkedDevice[]` defaulting to `[]`, and the selector returned `linkedDevices ?? []`, so "not asked yet" and "nothing linked" were one value. Every surface would have painted *No linked devices* from its first frame.

It is now `LinkedDevice[] | undefined`, and a surface draws no list at all until a read comes back. The rows belong to one community's team graph, so they are dropped on `setCurrentCommunity` and `deleteCommunity`; the persist transform hands back `undefined` for the same reason.

### Geometry

The frames' hidden nodes, now that the bordered group exists: overline header in `#7F7F7F` over the library's card, one row per device with a hairline between, and *No linked devices* in the frames' `#767676` (the `gray60` token, already carrying that comment in both themes). Rows are keyed and identified by `deviceId`, which is unique; the name is only what is shown. The mobile card scrolls past about eight rows.

Not drawn: the frames' per-row *Active* subtitle. The backend reports no such status, and it would read the same on every row this can draw. Recorded as a gap in `ONBOARDING.md`.

### Tests

| gate | result |
| --- | --- |
| desktop `tsc --noEmit` | clean |
| mobile `tsc --noEmit` | clean |
| e2e-tests `tsc --noEmit` | clean |
| desktop jest | 131 suites, 623 passed |
| mobile jest | 98 suites, 429 passed |
| state-manager jest (appConnection) | 8 suites, 42 passed |
| desktop eslint (changed dirs) | clean |

New coverage: `LinkedDevicesList` (populated, this-device excluded, removed excluded, empty, unread, duplicate names keyed apart); the shared component drawing it on share and never on receive; both desktop containers and the mobile screen asking for it only in a community; the state-manager telling unread from empty and forgetting the list on a community change.

E2E: `A's device list shows B, and B's shows A` was `it.skip` for want of a backend listing and now runs. The empty-list assertion reads the *No linked devices* line off A's tab before B links, which means something now that the line only appears after a real read. The retry for a populated list **reopens the Settings tab between attempts**, because the surface only asks the backend when it mounts. The assertion names the devices rather than only proving the lists differ: each surface lists every device on the account except its own, so the row A sees cannot appear in B's list unless it were some third device.

### Storybook

Static build, looked at all five states: Link devices in a community (two rows, this device excluded), in a community with nothing linked yet, device list not read yet (no card at all), no community (receive, no list), and Settings · Linked devices. Mobile gains the same three share variants as React Native stories. The no-bar rule still holds.

### Base

Merged `origin/develop` after #3648 landed squashed. Develop's share/receive split and restored stories had rewritten both platforms' Link devices surfaces underneath this branch, so the merge takes develop wholesale and the list is re-expressed on the new components in the commit after it.


holmesworcester:
Rebased onto develop now that #3648 has landed, and reworked for the review findings.

**Merged develop (`92b91f78c`).** #3648 was squashed, so the four recovery commits this branch was built on conflicted with identical content already in develop. On top of that the share/receive split and the restored stories had rewritten both platforms' Link devices surfaces underneath this branch: the desktop Settings tab now renders `Onboarding/LinkDevicesComponent` with a `direction`, its old `.component`/`.types` files are gone, and the QR moved into a modal step. Every conflict is resolved to develop; the list is re-expressed on the new components in `5092b692c`. Base stayed `develop` throughout, so no retarget was needed.

**Review findings, all applied.**

- **A (selector).** `connection.linkedDevices` is now `LinkedDevice[] | undefined` and the selector no longer coerces to `[]`, so "not asked yet" is distinguishable from "nothing linked". Surfaces draw no list until a read returns. The cached rows are dropped on `setCurrentCommunity` and `deleteCommunity`, which also closes the stale-rows-after-switching nit; the persist transform returns `undefined` for the same reason. This is what makes the list safe to ship at all: develop's own components said a card that always read *No linked devices* would be false the moment a device was linked.
- **B (vacuous e2e).** `hasNoLinkedDevices()` waits for the empty line instead of probing once, and the line only exists after a real read, so the assertion can now fail.
- **C (the wait re-triggers a fetch).** The retry **reopens the Settings tab between attempts**, remounting the surface, because the surface only asks the backend on mount. I chose remount over widening the saga's refetch so the fix stays in the test.
- **D (mobile screen test).** Covers both gating directions plus the unread state.
- **F.** The mobile card scrolls past about eight rows.
- **H.** Row testids key on `deviceId`; the name stays as the row text. There is a test with two devices sharing a name to hold that.
- **Overline vs heading.** The label carries `linked-devices-list-label` and tests query it by testid. I kept the frame's copy rather than inventing "Your devices" — the frames really do draw the same words twice.
- **"A shows B".** Now names the devices: each surface lists every device on the account except its own, so the row A sees cannot appear in B's list unless it were some third device, and asserting its absence there identifies it as B's.

**Not done, deliberately.** The frames draw an *Active* subtitle under each device name. The backend reports no such status, and every row the list can draw is active by construction, so it would be decoration. Recorded as a gap in the `ONBOARDING.md` screen table rather than invented.

The `otherLinkedDevices` helper is still duplicated between the desktop and mobile components (nit E). Sharing it means a new export in `@quiet/common` for a two-line filter; happy to do it if you want it.

Gates on the merged branch: desktop 131 suites / 623 passed, mobile 98 / 429, state-manager appConnection 8 / 42, three typechecks and desktop eslint clean. I looked at all five Storybook states, including the new "device list not read yet".

Issue #3537, opened 2026-09-15
Reuse validated sigchain state instead of rescanning all prior lockboxes


holmesworcester:
## Problem and priority

**Second milestone after message-history growth, but still required for this work.** Loading a larger valid sigchain repeatedly scans and structurally validates previously accepted lockboxes while rebuilding existing key-commitment maps. Each ordinary addition should reuse established validation facts instead of treating the entire known chain as new.

The concern is distinct from message-history rereads and from iOS primitive crypto speed. Context-independent signature/structure facts can be retained, while membership and authorization decisions must still be reevaluated when their graph/context changes.

## Measured reproduction

Quiet 10.0.0-alpha.0 (`edd51028f3c66c30d0692e8833fa7d0a7266b242`), auth `6f534c89bceb875e8c71997943e5e76e48ccbd88`; physical iPhone 16e, iOS 18.5. Offline fixtures contain real protocol-4 invitations, possession proofs, admissions, device joins and role grants.

| Users | Graph links | Load time | `establishedCommitments` sampled inclusive time |
| ---: | ---: | ---: | ---: |
| 10 | 31 | 4.047 s | 0.077 s |
| 25 | 76 | 11.157 s | 0.423 s |
| 50 | 151 | 24.077 s | 1.649 s |
| 100 | 301 | 55.207 s | 6.598 s |

Doubling 25→50→100 users approximately quadruples time in commitment-map reconstruction. At 100 users the whole load also performs 301 asymmetric decryptions and 499 signature verifications. Libsodium accounts for 41.18 of 55.25 sampled seconds; `authorizedLockboxes` takes 13.52 seconds inclusive, containing the 6.60-second commitment scan. Do not add overlapping inclusive figures together.

This does not imply all load cost is quadratic: primitive crypto is still the largest component. Per-message time stays nearly flat across these user counts. The initial matrix uses one process, increasing fixture sizes, stable roles and one message author; common-prefix caches and thermal state are not controlled.

## Code and implementation work

Auth `packages/auth/src/team/lockboxAuthorization.ts`: `authorizedLockboxes()` calls `establishedCommitments(previousState.lockboxes)`, which scans and validates the accumulated lockboxes. Related filters validate the same manifests again. LFA graph-concurrency code has nested traversal too, but it was not the leading quadratic cost in this profile.

- [ ] Maintain a persistent/incremental established-commitment index as validated lockboxes append; avoid reconstructing it for every ordinary link.
[...]

Issue #3536, opened 2026-09-15
Avoid quadratic message-history processing and repeated key validation


holmesworcester:
## Problem and priority

**First priority: message-count growth.** Receiving one message currently consumes it and then decrypts/verifies the entire channel history again to refresh message IDs. Catch-up from an empty channel therefore performs quadratic work. The same path also repeatedly rebuilds the LFA key map and revalidates unchanged key material for each message.

This was observed during real QSS delivery in Quiet 10.0.0-alpha.0 and reproduced locally on a physical iPhone 16e / iOS 18.5. It is client processing, not Tor latency or a QSS deployment problem. The implementation is shared with desktop and Android; iOS makes its cost especially visible.

## Measured reproduction

Published alpha: Quiet `edd51028f3c66c30d0692e8833fa7d0a7266b242`, auth `6f534c89bceb875e8c71997943e5e76e48ccbd88`.

| Workload | Observed result |
| --- | ---: |
| Refresh message IDs for a 10-message channel | 5.627 seconds |
| Refresh message IDs for a 100-message channel | 57.058 seconds |
| Actual update handler, 100 sequential arrivals | 5,150 consumer calls |
| Actual update handler, 1,000 sequential arrivals | 501,500 consumer calls |
| One decrypt/verify pass through 1,000 messages, two-member team | 541.801 seconds |

The handler's measured count is `N + N(N+1)/2`. The 501,500 figure is a work-count test of the actual handler with a counted consumer, not a completed 501,500-decryption phone run. Each normal ordinary-member message also reconstructs 13 Ed25519 keypairs, derives 13 Curve25519 public keys and opens three unchanged lockboxes. Key lookup accounts for 90.6% of the 1,000-message processing batch.

## Code and cause

- `packages/backend/src/nest/storage/channels/channel.store.ts`: `subscribe()` update handler calls `refreshMessageIds()`; that calls `getEntries()` and consumes every entry.
- `PublicChannelMessagesService.onConsume` → `CryptoService.decryptAndVerify` → LFA `team.keys` / `keyMap` / `visibleKeys` / lockbox validation.
- The full-history reread already existed in 8.0.0. Auth's August 24 keyset-manifest-binding change removed lockbox-open memoization and added expensive validation; that code is in both 9.0.2 and 10 alpha. The profiling work did not perform a fresh full-release historical bisect.

## Implementation work

- [ ] Maintain an incremental index of validated/readable messages. New arrivals should not reconsume all existing history.
[...]

PR #3654, opened 2026-09-22
Avoid repeated message-history processing (forward-port of #3539)

holmesworcester:
Forward-port of #3539 "Avoid repeated message-history processing" onto `develop` for 11.3. This supersedes #3539, which targets `10.0.0`.

## What #3539 does

Quiet re-scanned and re-consumed a channel's whole message history on every OrbitDB head update. `ChannelStore` now keeps an index of the entries it has successfully consumed (hash → message ID, ID → hashes, fully checked ancestry). An update walks only the ancestry the joined head brought in, using the log's own `has`/`get`, so one head that imports a whole backlog is processed once. `MESSAGE_IDS_STORED` announces the delta for ordinary arrivals and a snapshot after a rebuild. A sigchain update invalidates the index (epoch bump) and rebuilds it from `log.heads()`; heads a peer announced before the local log joined them wait for local acceptance. A pre-existing race is fixed where a delayed mobile notification could fire after an authorization change or after the store closed. The state-manager `missingChannelMessages` selector uses the entity map instead of scanning the ID array per announced ID. Message signatures and current authorization are still checked on every consume.

## What changed in the port

- `git merge origin/fix/ios-qss-processing`: `10.0.0` is fully contained in `develop`, so the merge brings only the PR's nine commits. The backend and state-manager diff is line-for-line the same as #3539's.
- `ChannelStore` on `develop` also serves DM channels (`DirectMessagesService`, `deliveredDmIds`). The one conflict was `_handleMessageOnUpdate`: it keeps develop's DM delivery de-dup and takes the epoch parameter.
- #3539's e2e changes (`UsersList.waitForVisibleUser` and the seven roster assertions) are dropped: `develop` removed the sidebar user list and its `-user-link` test id; those roster checks now live in Community membership and are partly skipped on `develop`. The e2e package is unchanged from `develop`.
- Spec adaptations only: develop's mobile-notification and DM fan-out specs fake the store with an iterator only, so they now fake a `log` keyed by entry hash and a redelivered DM gets its own entry (as a second OrbitDB entry would); they expect the delta announcement. #3539's deferred-notification spec supplies `getActiveChain`, which `develop` reads for the local identity. Two constructor calls pass the `DirectMessagesService` stub.
- CHANGELOG line under 11.1.0 Fixes.

## Auth submodule pin

`develop` stays pinned to auth `17e0b5bc`. #3539 pinned auth to the head of TryQuiet/auth#39 (`f8fb337a`, five commits on top of `10.0.0`'s pin). auth#39 is still open and none of its commits are in `develop`'s pin, so carrying it forward would mean pinning `develop` to a PR branch. Nothing in the quiet-side change uses a new auth API; it builds and tests against develop's pin. When auth#39 merges to auth `main`, a separate pin bump brings its cache in.

## Interaction with develop-side work

- #3559 (heads exchange retry after membership authentication): runs on `AUTH_CONNECTED`, re-sends heads via `store.applyOperation`; the OrbitDB fork emits `update` only for newly joined entries, so retried heads are either no-ops or one normal indexed walk. If a sigchain update lands mid-walk, the walk is discarded and the rebuild from `log.heads()` recovers the IDs. No double processing.
- #3569 (retry undecryptable channel metadata): sweeps only the three `KeyValueIndexedValidated` metadata stores; never touches `ChannelStore.getEntries`/`close`, and `subscribe` stays guarded to once per repo.
- #3625 (join announcements): state-manager saga and types only; no backend contact.
- Secure DMs: DM message channels use `ChannelStore` itself, so they get the index with no extra code; the DM de-dup set and the ID → hashes map cover a message re-appended under the same ID. The DM *metadata* store is a different mechanism (a persisted per-entry indexed flag with an incremental walk on its own `onUpdate`); its retry path walks the whole metadata log but skips validation for already-indexed entries, and that is bounded by the number of channels, not messages. Not extended here.

## Tests (local, on this branch)

- Backend `test-ci` (with #3638's worker recycling): 103 suites passed, 2 skipped; 1287 tests passed, 6 skipped, 0 failed. `channel.store.spec.ts`, `channel.store.replication.spec.ts` (real OrbitDB backlog, impersonating ancestor, 1,000-ancestor scaling case) and `author-impersonation.spec.ts` all run under `test-ci`.
- State-manager: 70 suites, 306 tests passed, 13 snapshots.
- e2e-tests typecheck: clean. Backend lint on changed files: clean.
- Desktop QSS e2e against an AppImage packaged from this branch (`ENVFILE=.env.e2e.qss`, QSS at develop's pin): `multipleClients.qss.test.ts` 63 passed, 23 skipped (develop's skips), 0 failed. `multipleClients.privateChannels.qss.test.ts` 60 passed, 15 failed: every message-history assertion passed; the failures begin at "First user's sidebar is missing private channel" (sidebar channel-list selector timeout) and "Owner creates a second private channel" (channel link button not found), i.e. sidebar/channel-creation UI steps, while another session's e2e run shared the box. Not attributed to this change, but not cleared either: no `develop` baseline run of that suite was available to compare.

## Not run

Backend Tor and long-running suites, the other QSS e2e suites, mobile, and a `develop`-baseline private-channels e2e run.

Refs #3539, #3536, #3537, TryQuiet/auth#39.


holmesworcester:
## Private-channels QSS e2e: develop baseline vs this branch

The PR body left `multipleClients.privateChannels.qss.test.ts` unresolved (60 passed / 15 failed on a busy box, no develop baseline). Ran it twice more on the same box, back to back, same recipe (`ENVFILE=.env.e2e.qss` AppImage, `--no-sandbox`, Xvfb + fluxbox, the same QSS stack at `ws://127.0.0.1:3003`):

| AppImage built from | passed | failed | runtime |
|---|---|---|---|
| `develop` @ d2007b039 (baseline) | 68 | 7 | 371 s |
| this branch @ 27fd84098 | 68 | 7 | 377 s |

The seven failures are the identical set on both sides:

- Creating Private Channel Before User Joins › First User Doesn't See Private Channel › First user's sidebar is missing private channel
- Creating Private Channel Before User Joins › User Sees Private Channel When Owner Offline › First user sees private channel in sidebar
- Creating Private Channel After User Joins › Owner Creates Another Private Channel › Owner creates a second private channel
- Creating Private Channel After User Joins › Owner Adds User To Second Private Channel › First user's sidebar is missing private channel
- Creating Private Channel After User Joins › Owner Adds User To Second Private Channel › First user sees second private channel in sidebar
- Delete private channel › Owner deletes private channel
- Second User Joins Using QSS Only › Second User Joins › Second user joins through QSS and sees expected state

They are sidebar channel-list / channel-link selector timeouts and one join-state check. Every message-history assertion passes on both sides.

**Verdict: pre-existing on `develop`, not caused by this port.** The earlier 15-failure run was the same suite on a more heavily shared box (the extra eight were the same UI steps timing out earlier).

Also re-ran backend `test-ci` at 27fd84098, i.e. after the final `develop` merge (the PR body's run predated it): 103 suites passed, 2 skipped; 1290 tests passed, 6 skipped, 0 failed.

holmesworcester:
## Code review findings and fixes

A high-effort review of the backend diff (`channel.store.ts` and its specs) found six blocking defects in the incremental message index. All six are fixed in commit 6fdedc797, each with a test that reproduces the failure mode first.

| # | Finding | Fix | Test |
|---|---|---|---|
| 1 | `indexJoinedAncestry` threw (`not joined` / `Missing channel log entry`) inside the un-awaited `'update'` listener. That is an `unhandledRejection`, which `backendManager` answers with `initiateShutdown`; the same throw rejected `subscribe()`, so the channel was never marked subscribed. | The walk never throws for unreadable ancestry. The head stays in a pending set and a backoff timer (1 s doubling to 30 s, `unref`) re-walks only pending heads. `refreshMessageIds` / the rebuild log and retry instead of rejecting `subscribe()`. Other read failures in the listener are caught and retried the same way. | `does not treat an unavailable or unjoined ancestor as completed ancestry` (now asserts the listener resolves), `retries an unreadable ancestor in the background…`, `subscribes when the initial rebuild cannot read an ancestor…`, `never rejects the shared update listener when a log read fails…` |
| 2 | The rebuild `continue`d over a head present in `log.heads()` but absent from the log index (`append` persists heads before the index entry; a kill in between leaves exactly that) and marked the index ready, so the message stayed invisible. | Heads the log itself reports (`trusted`) are read from their entry bytes when `has()` denies them, matching what the iterator would show. Heads announced over the shared bus keep the strict `has()` gate (no bytes fetched for unjoined entries). | `indexes a head the log reports from its bytes when the log index lags behind heads()`; `defers a snapshot head…` adjusted for the one extra read |
| 3 | `MESSAGES_STORED` and the mobile push were dropped when a head's walk was superseded by an auth rebuild, or when an earlier queued walk had already indexed it. | Heads are registered as pending (`notify: true`) synchronously in the listener, before the walk is queued. Whichever walk indexes the entry (its own, an earlier head's, or the rebuild) announces it, exactly once; pending entries survive a full invalidation. | `announces an arrival exactly once when an auth update interrupts its walk`, `announces each concurrent head once even when an earlier walk indexed it`, existing `emits a deferred mobile notification…` cases |
| 4 | Every `SigchainEvents.UPDATED` discarded the whole index and re-consumed every message of every channel, reintroducing the O(history) scan the PR exists to remove. | The store fingerprints each member's keys and roles (`team.members()`) when it checks the index. An update that leaves the local user's own facts unchanged bumps the epoch (discarding in-flight walks, which re-run from the pending set) and rechecks only entries authored by members whose facts changed plus entries rejected so far; a change to the local user's own facts, or an unavailable context, still rebuilds everything. | `keeps checked consumes when a sigchain update only admits a new member` (50 messages, 0 re-consumes), `rechecks only the messages of a member whose keys or roles changed`, `retries rejected entries on a sigchain update without rewalking accepted history`, `rebuilds the whole index when the local user's own authorization changes` |
| 5 | The direct-fetch fast path fired only for exactly one ID; the frontend's batch `getMessages({ ids })` after a multi-entry import fell through to a full log scan. | When the index is ready and every requested ID maps to exactly one log entry, each is fetched by hash and re-consumed under the current epoch. Unknown IDs are skipped; any ID with several entries keeps the ordered iterator. | `fetches a batch of announced IDs directly without rescanning history` (3 hash reads, 0 iterations) |
| 6 | The `'update'` listener on the shared static `OrbitDbService.events` bus was never removed in `close()` / `clean()`: a leak plus dead-store callbacks across leave/rejoin. | The listener is kept and detached in `close()` and `clean()` (like the auth listener); `subscribe()` detaches a previous one first; the retry timer is cleared too. | `detaches its update listener from the shared events bus on close` / `on clean` |

Relation to TryQuiet/auth#39: that companion PR caches checked lockbox keys per Team and reuses graph-crypto facts, which makes each `onConsume` cheaper; it does not change how many entries are re-consumed, since message signatures and current authorization are still checked per message. Finding 4 is the scope half of the same problem: quiet keeps nothing about keys or lockboxes, only the per-entry consume outcome and a fingerprint of each member's public keys and roles, so the two compose without overlap.

Also done (the review's optional items): notifications and the username lookup now run after a walk returns, outside the index queue, so they no longer serialize indexing (7); the `hash → id` and `hash → author` maps are one `indexedEntries` map (8).

Semantics that changed on purpose:

- Ancestors are consumed before their descendants (unchanged), so an arrival whose ancestor is unreadable announces nothing until the retry completes; the entries it did consume are indexed and announced immediately.
- On a sigchain update that does not touch the local user, removed IDs are communicated by the reconciliation snapshot as before; the frontend still treats IDs as candidates, not an authoritative list.

Test results on the fixed branch are in the next comment.

holmesworcester:
## Test results at 6fdedc797 (the review fixes)

- Backend `test-ci`: 103 suites passed, 2 skipped; 1303 tests passed, 6 skipped, 0 failed (13 new tests in `channel.store.spec.ts`, now 48). `channel.store.replication.spec.ts`, `direct-messages.security.spec.ts`, `author-impersonation.spec.ts`, `channels.service.spec.ts` included. ESLint and Prettier clean on the changed files.
- Desktop QSS e2e against an AppImage packaged from 6fdedc797 (`ENVFILE=.env.e2e.qss`, same QSS stack and recipe as the baseline comment above):
  - `multipleClients.qss.test.ts`: 63 passed, 23 skipped, 0 failed.
  - `multipleClients.privateChannels.qss.test.ts`: 68 passed, 7 failed. The 7 are the identical pre-existing set that `develop` fails on this box (see the baseline comment); every message-history assertion passes.

Not run: backend Tor and long-running suites, the other QSS e2e suites, mobile.

holmesworcester:
## Security audit iterations 2–6 (Daybreak Blue) → ACCEPT at a85fe1765

Adversarial re-audits of the incremental message index, one per fix commit; each iteration's probes were converted into regression tests that fail against the previous commit. Head a85fe1765 also bumps `3rd-party/auth` to `922f9490a` (TryQuiet/auth#39 + TryQuiet/auth#42) and merges `develop` at 5c9844930.

| Iter | Verdict | Blocking findings | Fix |
|---|---|---|---|
| 2 | REJECT | A walk paused in `onConsume` under an author's old keys could publish after the author's keys rotated (narrow updates changed no epoch); a second change to the same author during their running recheck was erased by the unversioned dirty set; `recheckAuthorization`/rebuild ignored a failed publication; a DM was marked delivered before emission; partial fan-out duplicated | 23c6f17e6: `authRevision` versions every consume and its announcement; `dirtyAuthors` is `author → revision`; publish failures schedule a retry on every path; DM marked delivered only after all listeners; delivery per listener with retry |
| 3 | REJECT | The superseded walk still emitted its ID through the epoch-only `MESSAGE_IDS_STORED` delta; dispatcher broke `once`/duplicate registrations; throw-after-effect duplication | 875cdd750: ID delta gated on epoch + revision, snapshots exclude authors awaiting a recheck; delivery per registration via `rawListeners`; contract documented |
| 4 | REJECT | `refreshMessageIds` selected IDs and then awaited the community lookup (TOCTOU); `ChannelsService` re-emitted with `emit`, which stops at the first throwing consumer and acknowledged the loss | 114aa9c29: IDs decided after the last await; the store delivers to the service's consumers registration by registration (`forwardAnnouncementsTo`), ID failures re-send a snapshot |
| 5 | REJECT | A consumer changing authorization from inside its callback still let later registrations receive the old-facts payload; four service tests faked the store without the new method | a85fe1765: currentness re-checked before every registration; doubles fixed |
| 6 | **ACCEPT** | none | — |

Non-blocking follow-up kept by the auditor across iterations: every accepted sigchain update rechecks the whole rejected set (O(rejected history)); narrowing it needs a consumer-level rejection taxonomy. Pre-existing and out of scope: the iterator fallback in `getEntries` is not atomic across auth changes; `deliveredDmIds` is process-lifetime.

### Results at a85fe1765
- Backend `test-ci`: 103 suites passed, 2 skipped; 1336 tests passed, 6 skipped, 0 failed. `channel.store.spec.ts` 79 (32 audit regressions), `channels.service.spec.ts` 47. Typecheck, ESLint, Prettier clean.
- State-manager 313 passed; e2e-tests typecheck clean.
- Desktop QSS e2e from an AppImage packaged at this head: `multipleClients.qss` 63 passed / 0 failed; `multipleClients.privateChannels.qss` 68 passed / 7 failed, the identical `develop` baseline set. Two runs during heavy box load failed the Tor presence step first (owner OFFLINE after a failed onion dial) and cascaded; the rerun returned to the baseline set.

